PR #1, opened 2026-06-11
feat(gsm8k): add dataset and 8-shot base-model task

jack-scitix-ai:
## Type

## Summary

- Add the GSM8K dataset wrapper backed by `hf:openai/gsm8k`.
- Add `gsm8k_8shot_base_gen`, an 8-shot generative task for base models.
- Mirror lm-evaluation-harness GSM8K prompt, stop sequences, and answer extraction behavior, with `num_fewshot` set to 8.
- Register the GSM8K dataset/task exports and meta index entries for discovery.
- Reference paper: https://arxiv.org/abs/2110.14168
- Reference repo: https://github.com/EleutherAI/lm-evaluation-harness/blob/1dd931087362abba74e0375c8c631295559f48b2/lm_eval/tasks/gsm8k/gsm8k.yaml

### Automated

### Manual

- [x] Dataset loading tested (`pdm run sieval dataset download gsm8k --data-dir /tmp/sieval-gsm8k-download-test`) succeeds
- [x] Ran `gsm8k_8shot_base_gen` on the full GSM8K test split with Qwen2.5-72B Base
- [x] Run completed 1319/1319 samples with 0 failures
- [x] Actual score: 86.35 exact match; reference score: 88.30; diff: -1.95

| Model | Expected| Actual | Diff |
| --- | ---: | ---: | ---: |
| Qwen2.5-72B Base | 88.30 | 86.35 | -1.95(-2.21%) |

### Required (all PRs)

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -47,7 +47,6 @@ tasks:
         model: math_model
         args:
             k: 8
-            n: 1
         # infer_args:                  # per-task inference parameter overrides
         #     max_tokens: 512
 ```
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -41,7 +41,7 @@ datasets:
         path: "openai/gsm8k"
 
 tasks:
-    gsm8k_8shot:
+    gsm8k_kshot_base_gen:
         class: GSM8KFewShotBaseGenTask
         dataset: gsm8k
         model: math_model
--- a/sieval/cli/dataset/render.py
+++ b/sieval/cli/dataset/render.py
@@ -10,6 +10,7 @@ from sieval.cli._readiness import evaluate_dataset_readiness, readiness_to_wire
 from sieval.cli.output import CommandResult
 from sieval.core.datasets.meta import DatasetMeta
 from sieval.core.tasks.meta import TaskMeta
+from sieval.datasets.downloaders.hf import parse_hf_source
 
 
 def _suggested_yaml_path(meta: DatasetMeta) -> str | None:
@@ -25,7 +26,7 @@ def _suggested_yaml_path(meta: DatasetMeta) -> str | None:
     if url_sources and not hf_sources:
         return "${SIEVAL_DATA_DIR}/" + meta.name
     if len(hf_sources) == 1 and not url_sources:
-        return hf_sources[0][len("hf:") :]
+        return parse_hf_source(hf_sources[0]).repo_id
     return None
 
 
--- a/sieval/datasets/__init__.pyi
+++ b/sieval/datasets/__init__.pyi
@@ -17,6 +17,10 @@ from .gpqa_diamond import (
     GPQADiamondDataset,
     GPQADiamondDatasetSample,
 )
+from .gsm8k import (
+    GSM8KDataset,
+    GSM8KDatasetSample,
+)
 from .human_eval import (
     HumanEvalDataset,
     HumanEvalDatasetSample,
@@ -55,6 +59,8 @@ __all__ = [
     "DROPDatasetSample",
     "GPQADiamondDataset",
     "GPQADiamondDatasetSample",
+    "GSM8KDataset",
+    "GSM8KDatasetSample",
     "HumanEvalDataset",
     "HumanEvalDatasetSample",
     "IFEvalDataset",
--- a/sieval/datasets/downloaders/hf.py
+++ b/sieval/datasets/downloaders/hf.py
@@ -14,9 +14,29 @@ AI-Generated Code - Claude Opus 4.7 (1M context) (Anthropic)
 """
 
 from pathlib import Path
+from typing import NamedTuple
 
 # Payload suffixes found across pilot datasets; extend as other formats surface.
 _DATA_SUFFIXES = (".arrow", ".parquet", ".jsonl", ".json", ".csv")
+
+
+class HFSource(NamedTuple):
+    repo_id: str
+    revision: str | None
+
+
+def parse_hf_source(source: str) -> HFSource:
+    if not source.startswith("hf:"):
+        raise ValueError(f"Expected hf: scheme, got {source!r}")
+    repo_ref = source[len("hf:") :]
+    if not repo_ref:
+        raise ValueError(f"Invalid hf source: {source!r}")
+    repo_id, separator, revision = repo_ref.rpartition("@")
+    if not separator:
+        return HFSource(repo_id=repo_ref, revision=None)
+    if not repo_id or not revision:
+        raise ValueError(f"Invalid hf source revision pin: {source!r}")
+    return HFSource(repo_id=repo_id, revision=revision)
 
 
 class HFHandler:
@@ -31,12 +51,13 @@ class HFHandler:
     ) -> None:
         from huggingface_hub import snapshot_download
 
-        repo_id = self._strip_scheme(source)
-        local_dir = dest_root / repo_id
+        parsed = parse_hf_source(source)
+        local_dir = dest_root / parsed.repo_id
         local_dir.mkdir(parents=True, exist_ok=True)
         snapshot_download(
-            repo_id=repo_id,
+            repo_id=parsed.repo_id,
             repo_type="dataset",
+            revision=parsed.revision,
             local_dir=str(local_dir),
             force_download=force,
         )
@@ -47,8 +68,8 @@ class HFHandler:
         dest_root: Path,
         dataset_name: str,
     ) -> bool:
-        repo_id = self._strip_scheme(source)
-        local_dir = dest_root / repo_id
+        parsed = parse_hf_source(source)
+        local_dir = dest_root / parsed.repo_id
         if not local_dir.exists():
             return False
         # `huggingface_hub` writes `.incomplete` sidecars under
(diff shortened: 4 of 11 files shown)


ethan-scitix (changes requested):
Structure is clean and preflight passes. I re-ran the metric on the validated run (`outputs/gsm8k_qwen2.5-72b`, 1319 samples) to ground these. Four things before merge:

**1. Which metric are we aligning to?** Recomputing both lm-eval filters on the run:

| metric | EM |
| --- | ---: |
| strict-match (`#### N`) | 85.29 |
| flexible-extract (last number) | 86.35 |
| committed `report()` (strict→flexible fused) | 86.35 |

The fused number equals flexible-extract here. The target 88.3 is from DeepSeek-V3 Table 3 (GSM8K **EM, 8-shot**, Qwen2.5-72B-Base — [arXiv:2412.19437](https://arxiv.org/abs/2412.19437)), which is why `num_fewshot=8` (not lm-eval's 5). But dsv3 doesn't publish its extraction/normalization convention, so we can't tell whether 88.3 is strict or flexible. Please (a) state the real target = dsv3 8-shot EM in the docstring, and (b) pick one reported metric (or report both strict+flexible) rather than the fused one.

**2. `max_tokens=256` is truncating answers.** 24/1319 outputs hit the cap and **22 of them scored wrong (~1.67 pts)** — likely most of the −1.95 gap to 88.3. Raise to ≥512 and re-run.

**3. The run is non-deterministic** (`deterministic: false`; temp=0 is not bitwise-stable under SGLang batching). For an alignment number, either enable deterministic inference or run a few times to confirm 86.35 isn't noise.

**4. Code:**

- `load()`: `if name_or_path.startswith("hf://")` is dead code — the scheme is `hf:` not `hf://`, and `maybe_resolve_hf_path` already rewrites it to a local path before `load()`. Sibling `HumanEval` (also `hf:`) doesn't have it. Delete.
- Rename `gsm8k_8shot_base_gen` → `gsm8k_kshot_base_gen` (+ file), display "GSM8K (few-shot, base generative)". `k` is configurable, so `8shot` in the CLI name is misleading; mirror `drop_kshot_gen` and note "default k=8 (DeepSeek-V3)" in `reference_impl.notes`. (Not referenced by any config/card in the repo.)
- `_sample_fewshot_examples`: reuse `Dataset.retrieve_samples(k, split="train", mode="random", seed=...)` (as `drop_kshot_gen` does). It returns `list[TSample]`, dropping both the `cast` (only there for `ty`) and the full-train materialization. Note: it uses `.shuffle().select` vs the current `random.sample`, so examples — and the score — shift; re-validate.
- `report()` counts failed samples in the denominator, while `drop_kshot_gen` divides by `len(finals)` only — make them consistent.
- Minor: stop sequences use `"\n\n"` where the cited lm-eval yaml uses `"</s>"`; and `load_dataset` has no `revision=` pin (matters here since this is a leaderboard anchor).

jack-scitix-ai:
I’ve addressed the GSM8K alignment points in the latest push.

- Metric: the task now reports strict-match EM only. `score` and `exact_match` are documented as lm-eval-harness `strict-match`, aligned with the original GSM8K `####` answer delimiter. The docstring/metadata also state that the comparison target is DeepSeek-V3 Table 3, Qwen2.5-72B-Base GSM8K 8-shot EM = 88.3, and that DeepSeek-V3 does not specify strict vs flexible extraction.
- Generation cap: raised the default `max_tokens` to 2048.
- Determinism: re-ran with `deterministic: true`; the persisted run metadata records `deterministic=true`.
- Code cleanup: removed the dead `hf://` branch, renamed the task to `gsm8k_kshot_base_gen`, updated display name/docs/meta/stubs, switched few-shot selection to `Dataset.retrieve_samples(...)`, aligned the report denominator with DROP, changed the stop sequence to `</s>`, and pinned the GSM8K dataset revision.

New validation result on `outputs/gsm8k_qwen2.5-72b`:
- 1319 / 1319 completed
- 0 failed
- strict EM / exact_match: 90.82638362395754
- score: 90.82638362395754（diff: +2.87%）

There is still one `truncated_output` anomaly, sample 1056. I inspected it: the model entered a repetition loop and hit the length cap without emitting `#### N`, so it scored wrong under strict-match.

jack-scitix-ai:
I added one more small follow-up: HF dataset sources now support optional `@revision` pins so `sieval dataset download` can fetch a specific dataset revision. GSM8K is pinned to the validated HF revision, and `dataset show` still suggests the revision-free YAML path (`openai/gsm8k`) for eval configs.

ethan-scitix (changes requested):
Previous round looks well addressed. A few things left:

**Config — question:** the run was done at `max_tokens=1024`, but the committed default is `2048`. Which is intended? The reported score came from 1024, so the validated config and the default should agree (impact is tiny — 1 truncation).

**Code:**

- **C1 (should fix):** `parse_hf_source` / `@revision` → `snapshot_download(revision=...)` (the point of `3926cdab`) has no test — `test_hf.py` covers neither the `@`-split nor the two `ValueError` branches. Add a few cases.
- **C2 (should fix):** `n` is forwarded to `agenerate`, but `postprocess` only reads `texts[0]`, so `n>1` silently generates and discards the extra samples. Aggregate (maj@k) or drop the `n` knob.
- **C3 (cleanup):** `_normalize_exact_match`'s `re.sub(r"(?s).*#### ", "", text)` is dead — both callers pass an already-extracted group with no `####`. Remove.
- **C4 (low):** `revision=` in `load()` is a no-op on the staged local dir (the real pin is the downloader's `snapshot_download`); redundant/misleading — drop or comment.

**Why was flexible-extract removed?** lm-eval reports strict **and** flexible separately; this drops flexible and reports strict-only, which couples scoring to the `####`-format demos with no fallback — a CoT-style prompt or a model that doesn't emit `####` collapses the score toward 0. What's the rationale vs keeping it as a secondary metric? If strict-only is intentional, add a docstring line that the demo format and the strict extractor must stay in lockstep.

jack-scitix-ai:
Hi, I pushed the follow-up in `5175510` (`fix(gsm8k): report strict EM with flexible secondary metric`).

What changed:

- Kept `score` / `exact_match` as the primary GSM8K metric, but clarified in the task docstring and metadata that this is **strict GSM8K EM**: the prediction must emit the original GSM8K `#### N` final-answer delimiter. This is intentional because the original GSM8K evaluation format is delimiter-based; `exact_match` is therefore the official/strict metric in this task.
- Added `flexible_exact_match` as a secondary diagnostic metric, using lm-eval-harness `flexible-extract` last-number extraction. This is reported separately and is not treated as the original GSM8K official metric.
- Removed the fused strict→flexible behavior, so the primary score no longer silently falls back to flexible extraction.
- Dropped the unsupported `n` task knob. This task reports single-generation EM; accepting `n > 1` while only reading `texts[0]` was misleading. If we want maj@k later, it should be added explicitly as a separate behavior.
- Removed the dead `####` cleanup branch from `_normalize_exact_match`; both strict callers already pass the extracted answer group.
- Removed runtime `revision=` from `GSM8KDataset.load()`. The revision pin remains on the dataset source for `snapshot_download`, while the staged local dataset load stays revision-free.
- Added HF revision tests for `parse_hf_source`, `snapshot_download(revision=...)`, invalid pins, and revision-stripped local readiness.
- Added GSM8K task tests covering strict vs flexible extraction, `n` not being forwarded, and reporting `flexible_exact_match`.

Validation was rerun with the updated task code and default `max_tokens=2048`:

| metric | value |
| --- | ---: |
| completed | 1319 / 1319 |
| failed | 0 |
| score / exact_match (strict GSM8K EM) | 90.82638362395754 |
| flexible_exact_match (lm-eval flexible-extract, secondary) | 91.20545868081881 |

The persisted run metadata has `deterministic=true`. The request params also show `max_tokens=2048` and no `n` parameter. There is still one `truncated_output` anomaly, sample 1056; this is the same repetition-loop style case and remains wrong under strict matching.

Local checks:

- `pdm run ruff check sieval/tasks/gsm8k_kshot_base_gen.py sieval/datasets/gsm8k.py tests/unit/tasks/test_gsm8k_kshot_base_gen.py tests/unit/datasets/downloaders/test_hf.py`
- `pdm run ruff format --check sieval/tasks/gsm8k_kshot_base_gen.py sieval/datasets/gsm8k.py tests/unit/tasks/test_gsm8k_kshot_base_gen.py tests/unit/datasets/downloaders/test_hf.py`
- `pdm run pytest tests/unit/tasks/test_gsm8k_kshot_base_gen.py tests/unit/datasets/downloaders/test_hf.py -q`

ethan-scitix (changes requested):
Latest push looks good — flexible secondary metric, `n` removed, dead branch gone, `load()` revision dropped, tests added, 2048 re-run. Two small things:

1. **`meta/index.json` has 6 stray entries** — `mmmlu` / `mmmlu_zh_cn` / `theoremqa` + their `_kshot_base_gen` tasks, with no source files in this PR (and not in the stubs). Looks like the index sync ran against a dirty tree. Preflight misses it, but `dataset/task list` reads the index and would show non-loadable benchmarks. Please regenerate in a clean tree (drops those 6, −128 lines).

2. **`HFSource` → `@dataclass(frozen=True)`** — value objects here are almost all frozen dataclasses; NamedTuple is the odd one out. Tests keep passing as-is.

jack-scitix-ai:
Thanks, fixed in the latest push.
- Regenerated `sieval/meta/index.json` from a clean GSM8K tree; the stray MMMLU/TheoremQA dataset/task entries are gone.
- Changed `HFSource` from `NamedTuple` to `@dataclass(frozen=True)`.

ethan-scitix (approved):

PR #3, opened 2026-06-17
feat(theoremqa): add dataset and k-shot base-model task

jack-scitix-ai:
## Type

## Summary

- Add TheoremQA dataset support with a pinned HuggingFace source revision.
- Add `theoremqa_kshot_base_gen`, a base-model generative task aligned with the official TheoremQA vLLM runner.
- The Qwen2.5 technical report reports TheoremQA 5-shot = 42.4 for Qwen2.5-72B, but does not publish the exact evaluation script, prompt template, decoding parameters, or answer extraction details. Therefore this implementation aligns with the original TheoremQA repository's vLLM runner.
- Reference implementation: https://github.com/TIGER-AI-Lab/TheoremQA/blob/acfc9686aa9b49f3c8f189364a9a9ee9c53da039/run.py
- The task follows the upstream short-form prompt, built-in 5-shot examples, `temperature=0`, `top_p=1`, `max_tokens=2048`, stop tokens, and official `answer_clean` / numeric matching behavior.
- The task uses the existing `math` optional dependency group through `math-verify`, which brings `latex2sympy2-extended`; `latex2sympy` is lazy-imported so registration and metadata checks remain import-safe.
- Known alignment artifact: the upstream `answer_clean` heuristic can treat repeated `The answer is` triggers as ICL leakage. In the validated run, this caused 4 outputs to be cleaned to an empty postprocess result, even though the model produced non-empty text. This behavior is preserved for official implementation alignment.

### Automated

### Manual

- [x] Ran TheoremQA 5-shot evaluation with Qwen2.5-72B.
- [x] Compared against the Qwen2.5 technical report target.

| --- | ---: | ---: | ---: |
| Qwen2.5-72B | 42.4 | 44.63 | +2.23 (+2.23%) |

Validated report:

```json
{'score': 44.625, 'accuracy': 44.625, 'fails': 0, 'empty': 2}
```
The 2 empty postprocess cases are expected compatibility artifacts from the official TheoremQA cleaner. The model outputs were non-empty, but repeated `The answer is` triggers activated the upstream ICL-leakage heuristic and resulted in empty extracted predictions.

### Required (all PRs)

- [x] PR title follows conventional format: `feat(theoremqa): add dataset and k-shot base-model task`

- [x] Score comparison table included
- [x] Dataset loading tested

### If: New Dependency

- [x] Justified in Summary


ethan-scitix (changes requested):
Faithful reproduction overall — verified the prompt/examples/stop-tokens/sampling against the pinned upstream, and the reference run scores 43.25 vs the report's 42.8. A few changes before merge.

### `eval_theoremqa_5shot_qwen72b.yaml` — delete this file
No other benchmark checks in a top-level `eval_*.yaml`; it's the only one in the repo and hardcodes a personal mount path. Users compose their own run config — remove it.

### Decoding params belong in the model layer — fix both tasks
`agenerate` merges `{**self._kwargs, **kwargs}` (`sieval/core/models/gen_model.py:16`), so decoding params a task passes at call time **override** the model's configured `args` — the task silently wins over user config. They're model-owned and should come from the `models:` block.
- **`sieval/tasks/theoremqa_kshot_base_gen.py`** (`infer()`, L430–436): drop hardcoded `temperature=0, top_p=1, max_tokens=2048` (document the repro values in the docstring); keep `stop=_STOP_TOKENS` (prompt-coupled).
- **`sieval/tasks/gsm8k_kshot_base_gen.py`** (L141–142, `infer()` L172–178): same issue — drop the `max_tokens`/`temperature` args and their injection; keep `k`/`fewshot_*`/`stop`. Folding it in since this PR surfaces it.

### Annotate every deviation from official (`theoremqa_kshot_base_gen.py`)
Code can stay; just call these out in a comment/docstring so they're auditable:
- `_EVAL_GLOBALS` (L50): sandboxed vs upstream's bare `eval(num)` (drops builtins like `abs`/`round`/`pow`).
- `_load_latex2sympy` (L32): prefers official `latex2sympy2`, but on Python ≥3.12 only the `latex2sympy2_extended` fallback loads — so the parser differs from upstream in practice.

### Nits (`theoremqa_kshot_base_gen.py`)
- `preprocess()` (L424): rebuilds the k-shot prefix per sample though it only depends on `self._k`; compute once in `setup()` and cache.
- L405–407: reference URL split across two string literals; keep it on one line like the other tasks.

jack-scitix-ai:
Thanks for the review. I addressed the comments in the latest follow-up commit:

- Removed the top-level `eval_theoremqa_5shot_qwen72b.yaml`.
- Moved hardcoded decoding parameters out of the task layer; TheoremQA/GSM8K now only forward prompt-coupled stop sequences.
- Cached the fixed TheoremQA k-shot prompt prefix instead of rebuilding it per sample.
- Documented the TheoremQA deviations from upstream: sandboxed `eval` and the `latex2sympy2_extended` fallback.
- Kept the TheoremQA reference URL as a single-line constant while preserving the `run.py` reference.

ethan-scitix (commented):
Core scoring path verified faithful: matcher **0/1800** verdict diffs vs the reference over 800 real groundtruths; `Answer_type` groundtruth derivation **0/800** diffs vs the official runtime-type dispatch; `math` deps, HF revision pin, `index.json` sync, license (MIT), and all 17 tests OK. A few things to confirm:

**`sieval/tasks/theoremqa_kshot_base_gen.py`**

- `Question: :17` — where does the 42.8 target come from? In the Qwen2.5 report (arXiv:2412.15115, Table 2), 42.8 is **Qwen2-72B**; **Qwen2.5-72B is 42.4**. Is this from another source (a leaderboard / different shot count / the Qwen2 report)? If it's that table, suggest 42.4, or note 42.8 is actually Qwen2-72B.

- `Question:` few-shot examples — can you confirm the prompt actually fed to the model matches the official runtime prompt byte-for-byte? Two spots differ from upstream's evaluated output:
  - `:99` (ex0): upstream has `≈` (U+2248), the port has `is approximately`.
  - `:145` (ex2): upstream `examples.py` uses single-backslash `\frac`/`\right` in a non-raw string, so `from examples import examples` actually yields 3 control chars (FF/FF/CR) on that line (reproduce: `python3 -c "from examples import examples; print(repr(examples[2][1]))"`); the port emits clean LaTeX. The whole point of this task is reproduction, so it's worth confirming the assembled prompt lines up with what the official runner sends.

- `Question: :13` — the docstring says deviations are "limited to" the eval sandbox + latex2sympy fallback, but the prompt text above and the `Answer_type` mechanism (verified equivalent) are also deviations. Suggest completing the list once the prompt question is settled.

Not blocking on correctness — the open items are all prompt-fidelity / docstring and live in one file, so they can be handled in this PR.

ethan-scitix (changes requested):
Port verified faithful (examples/prompt/stop-tokens/matcher match upstream `acfc9686`; matcher re-validated against a real Qwen2.5-72B run, 0/800 mismatch). Code adjustments before merge:

### `sieval/datasets/theoremqa.py`

**Architecture:** `theoremqa.py:43-49` — `load()` strips `Picture` at the dataset layer. Filtering is the **task's** job; the dataset should mirror its `source` as-is. Dropping it here welds `theoremqa` to text-only and blocks any future task from consuming the column. No forcing function to drop it: `raw_sample` is never persisted (`context.py:207-229` serializes only stage-result fields), so the PIL `Image` cells (53/800) never hit JSON — keeping them is transparent to this task, which already reads only `Question`/`Answer`/`Answer_type`.

```python
@override
def load(self, name_or_path: str, **kwargs) -> HFDatasetDict:
    return ensure_dataset_dict(load_dataset(name_or_path, **kwargs))
```

Update `test_theoremqa.py`: keep the revision-not-forwarded assertions; flip the `Picture` one to assert it's retained.

### `sieval/tasks/theoremqa_kshot_base_gen.py`

**Nit:** `:645` / `:267-274` — real runtime dep is `latex2sympy2_extended`, provided only transitively via `math-verify`. Declare it directly in the `math` group so the contract can't silently break.

**Nit:** `:660` — make `k` keyword-only (`*,`) to match the gsm8k sibling.

jack-scitix-ai:
Thanks for the review — all four points are addressed in the last commit:
- `load()` now mirrors the source as-is and no longer strips `Picture`
  (filtering is the task's job); added `Picture` as `NotRequired` to the
  sample `TypedDict`.
- Declared `latex2sympy2-extended` directly in the `math` group instead of
  relying on it transitively via `math-verify`.
- Made the Task `k` argument keyword-only to match the gsm8k sibling.
- Flipped the test to assert `Picture` is retained.

ethan-scitix (changes requested):
Reviewed as a reproduction port against the official runner at the pinned SHA. The high-risk parts are faithful: the 5 few-shot examples match upstream `examples.py` byte-for-byte (including the non-raw `\f`/`\r` escape artifacts), `answer_clean` / `extract_theoremqa_answer` / `number_it` / `compare_*` / the short-format prompt all match `utils.py` / `number_utils.py` / `data_loader.py`, stop tokens and `n_shot=5` match the official defaults, and the documented deviations (sandboxed `eval`, latex2sympy2_extended fallback, typed-groundtruth-from-`Answer_type`) are annotated and behaviorally equivalent — verified on the full `Answer_type` distribution and 100 real rows. Both pins resolve and the run.py URL is commit-pinned. One blocking issue below.

## `sieval/datasets/theoremqa.py`

**`Bug:` `theoremqa.py:43` — implicit image decode pulls in an undeclared Pillow dependency.**

`Picture` is an HF `Image` feature (defaults to `decode=True`), and 53/800 test samples carry images, so materializing rows (`test_set[sid]`, `runner.py:850`) decodes them into `PIL.Image` — requiring Pillow. Nothing provides it: not the `math` group, not the default deps (`datasets` only pulls Pillow under its `vision` extra), so a clean `sieval[math]` install crashes mid-run.

Two things worth addressing: the decode is riding on an implicit HF default rather than a stated choice, and the resulting dependency shouldn't be folded into `math` — it's a loader-side image dep that conceptually wants its own group, set on the dataset's `deps_group` (readiness checks it independently of the task's group). Up to you how to handle the decode itself, but the dependency needs to be declared somewhere the readiness check can see it.

## `sieval/tasks/theoremqa_kshot_base_gen.py`

**`Nit:` `theoremqa_kshot_base_gen.py:484,493` — `report()` returns `fails`/`empty` as `float(...)`, while the gsm8k sibling returns `fails` as a bare `int`.** Purely cosmetic (both serialize to JSON numbers); only worth aligning if you want a uniform base-gen report schema. Non-blocking.

jack-scitix-ai:
Thanks — both points are addressed:

**Blocking — implicit Pillow dependency (`theoremqa.py`):** You're right, and this was a latent consequence of the round-3 "mirror source as-is" change. `Picture` is an HF `Image` feature defaulting to `decode=True`, so materializing the 53/800 picture rows decodes to PIL and pulls in Pillow — which `datasets` only ships under its `vision` extra, outside `sieval[math]`. Confirmed it crashes a clean install mid-run.

Rather than fold an image dep into `math`, I made the decode an explicit choice:
`load()` now casts `Picture` to `Image(decode=False)`. The task is text-only and never reads the image (and `raw_sample` isn't persisted), so the column is still mirrored as raw bytes and **no new dependency / `deps_group` is needed**. This felt cleaner than declaring Pillow for data the eval never decodes — happy to switch to a dataset-level group + decode-on if you'd prefer to keep the images materialized.

**Nit — `report()` field types:** aligned `fails`/`empty` to bare `int` to match the gsm8k sibling's schema.

I also replaced the dataset unit test's string-column `Picture` mock (which never exercised the decode path) with a real, Pillow-free `Image` feature fixture that asserts `decode is False`, so this can't regress silently.

ethan-scitix (approved):

jack-scitix-ai:
Re-ran on the post-review code (aligned prompt + model-layer decoding) and updated the PR body.

**TheoremQA 5-shot, Qwen2.5-72B (greedy, full 800):**
- **score / accuracy: 44.625** (357/800), `fails: 0`, `empty: 2`

The two empties are the documented ICL-leakage artifact — the model emits the answer then repeats `The answer is …`, which upstream `answer_clean` treats as leakage and returns empty (both happened to be correct, but we keep upstream's behavior). The 27 `truncated_output` anomalies are long chain-of-thought finishing with `reason="length"`; they're still scored, as upstream does. 

vs. the report's **42.4** for Qwen2.5-72B, this lands at +2.2 — cross-harness "nearby" rather than exact, which is expected given prompt-formatting/decoding/harness differences (the report doesn't publish exact eval scripts).

Also fixed the target-number mix-up you flagged earlier: the docstring now cites 42.4 for Qwen2.5-72B (42.8 is Qwen2-72B), corrected manually.

PR #6, opened 2026-06-18
feat(humaneval): add HumanEval 0-shot base model task

jack-scitix-ai:
## Type

## Summary
- Adds `human_eval_0shot_base_gen` for base completion models, aligned with [lm-evaluation-harness HumanEval YAML](https://github.com/EleutherAI/lm-evaluation-harness/blob/1dd931087362abba74e0375c8c631295559f48b2/lm_eval/tasks/humaneval/humaneval.yaml) for prompt, stop sequences, max generation tokens, zero-shot setting, repeats, and raw completions.
- Pins `human_eval` dataset source to `openai/openai_humaneval@7dce6050a7d6d172f3cc5c32aa97f52fa1a2e544`.
- Registers the new task in task metadata/stubs and adds focused unit tests.

### Automated

### Manual

- [x] Ran HumanEval 0-shot evaluation with Qwen2.5-72B.
- [x] Compared against the Qwen2.5 technical report target.

| --- | ---: | ---: | ---: |
| Qwen2.5 72B Base | 53.0 | 56.1 | +3.1(+5.8%) |

### Required (all PRs)

### DIFF
--- a/sieval/tasks/human_eval_0shot_base_gen.py
+++ b/sieval/tasks/human_eval_0shot_base_gen.py
@@ -102,7 +102,6 @@ class HumanEvalZeroShotBaseGenTask(
         super().__init__(dataset=dataset, model=model, name=name)
         self._k = k
         self._n = n
-        self._max_concurrency = max_concurrency
         self._timeout = timeout
         self._stop = stop
         self._code_eval_api = os.getenv(
@@ -155,6 +154,7 @@ class HumanEvalZeroShotBaseGenTask(
                         "uuid": f"{idx}-{time.perf_counter_ns()}",
                         "source": "human-eval",
                         "code": check_program,
+                        "timeout": self._timeout,
                     },
                     timeout=self._timeout + 2,
                 )
--- a/sieval/tasks/human_eval_0shot_base_gen.py
+++ b/sieval/tasks/human_eval_0shot_base_gen.py
@@ -1,5 +1,21 @@
 """HumanEval zero-shot base-model generative task.
 
+Reproduces the lm-evaluation-harness ``humaneval.yaml`` for base completion
+models, scored with pass@k via the SiEval code-eval API.
+
+Decoding follows the harness defaults and is configured on the model, not
+injected by this task: greedy sampling (``temperature=0``, ``top_p=1``) and
+``max_gen_toks=1024``. Set these through the model ``args`` or per-task
+``infer_args`` in the run config. The task owns only the prompt-coupled
+``stop`` sequences and ``n`` (the pass@k sampling count).
+
+HumanEval has no single canonical Qwen2.5-72B-Base score: references span
+roughly six points — DeepSeek-V3 Table 3 reports 53.0, while the Qwen2.5
+technical report self-reports 59.1 (with no published eval config). Because
+this task reproduces lm-eval-harness rather than Qwen's own setup, its score
+is expected to land between those references rather than match the Qwen
+self-report; the full 164-item run scores 56.1 (92/164, fails=0).
+
 AI-Generated Code - GPT-5.5-Codex (OpenAI)
 """
 
@@ -53,7 +69,12 @@ STOP_SEQUENCES = ("\nclass", "\ndef", "\n#", "\nif", "\nprint")
             "Aligned with lm-evaluation-harness humaneval.yaml prompt, stop "
             "sequences, max_gen_toks, zero-shot setting, repeats=1, and raw "
             "completion filtering; code execution is handled by the SiEval "
-            "code-eval API."
+            "code-eval API. No single canonical Qwen2.5-72B-Base target "
+            "exists: DeepSeek-V3 Table 3 reports 53.0 and the Qwen2.5 "
+            "technical report self-reports 59.1 (no published eval config). "
+            "This task reproduces lm-eval-harness, so its score is expected "
+            "to land between those references rather than match Qwen's "
+            "self-report."
         ),
     ),
 )
@@ -74,7 +95,6 @@ class HumanEvalZeroShotBaseGenTask(
         name: str | None = None,
         k: int = 1,
         n: int = 1,
-        max_tokens: int = 1024,
         max_concurrency: int = 4,
         timeout: float = 5.0,
         stop: tuple[str, ...] = STOP_SEQUENCES,
@@ -82,7 +102,6 @@ class HumanEvalZeroShotBaseGenTask(
         super().__init__(dataset=dataset, model=model, name=name)
         self._k = k
         self._n = n
-        self._max_tokens = max_tokens
         self._max_concurrency = max_concurrency
         self._timeout = timeout
         self._stop = stop
@@ -99,10 +118,13 @@ class HumanEvalZeroShotBaseGenTask(
 
     @override
     async def infer(self, pre, ctx):
+        # Decoding params (temperature, top_p, max_tokens) come from the
+        # model's configured args / per-task infer_args, not this task. Only
+        # the prompt-coupled stop sequences and the pass@k sample count live
+        # here.
         return await self.model.agenerate(
             pre,
             n=self._n,
-            max_tokens=self._max_tokens,
             stop=list(self._stop),
         )
 
--- a/sieval/tasks/human_eval_0shot_base_gen.py
+++ b/sieval/tasks/human_eval_0shot_base_gen.py
@@ -1,0 +1,198 @@
+"""HumanEval zero-shot base-model generative task.
+
+AI-Generated Code - GPT-5.5-Codex (OpenAI)
+"""
+
+import os
+import time
+from typing import TypedDict, override
+
+import httpx
+from loguru import logger
+
+from sieval.core.models import ModelOutput
+from sieval.core.tasks import (
+    EvalMode,
+    ReferenceImpl,
+    Task,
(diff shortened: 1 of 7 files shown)


ethan-scitix (changes requested):
**Bug: decoding params injected in the task layer** — `sieval/tasks/human_eval_0shot_base_gen.py:101-107` (+ `:77`).
`infer()` forwards `max_tokens` into `agenerate`, which merges `{**self._kwargs, **kwargs}` and silently overrides the model's configured `args`. Same anti-pattern flagged in PR #3 and already fixed for `theoremqa`/`gsm8k`. Fix the same way: drop `max_tokens` from the `agenerate` call and `__init__`; document the repro values (greedy `temperature=0`/`top_p=1`, `max_gen_toks=1024`) in the docstring; keep `stop` (prompt-coupled) and `n` (pass@k semantics).

**Question: no comparison target / harness caveat in `reference_impl.notes`.**
The `gsm8k` sibling cites its target (DSV3 88.3); this task cites none, despite a ~6-pt spread across references (DSV3 53.0 ↔ Qwen self-report 59.1, and Qwen publishes no eval config). The full-164 run scores 56.1 (92/164, fails=0), which sits sensibly between them. Add a line: reproduces lm-eval-harness, expected to land between those rather than match Qwen's self-report.

jack-scitix-ai:
Thanks, both problems are addressed

1. Decode params injected in the task layer. Dropped max_tokens from both infer() and __init__. Decoding now comes from the model config / per-task infer_args (greedy temperature=0, top_p=1, max_gen_toks=1024, documented in the module docstring); the task keeps only stop (prompt-coupled) and n (pass@k). The smoke YAML moves max_tokens: 1024 into infer_args accordingly. </br></br>No re-run was needed — I confirmed the on-the-wire request is identical to the original run by intercepting completions.create on both paths:

> * OLD (task-injected): {temperature: 0.0, n: 1, max_tokens: 1024, stop: [...]}
> * NEW (with_args(max_tokens=1024) → infer()): same fields, identical values.
> So the 56.1 (92/164, fails=0) number stands.


2. Comparison target / harness caveat. Added to reference_impl.notes (and the docstring): there's no single canonical Qwen2.5-72B-Base target — DSV3 Table 3 reports 53.0 vs. the Qwen2.5 self-report 59.1 (no published eval config) — so this task reproduces lm-eval-harness and is expected to land between those references rather than match Qwen's self-report; the full run scores 56.1.

sieval/tasks/human_eval_0shot_base_gen.py
```diff
@@ -0,0 +1,220 @@
...
+    peak_memory_mb: float
+
+
+class Feedback(TypedDict):
+    correct: bool
+    msg: str
+    metrics: ResourceMetrics | None
+
+
+STOP_SEQUENCES = ("\nclass", "\ndef", "\n#", "\nif", "\nprint")
```
ethan-scitix, line 52: **Nit:** `STOP_SEQUENCES` must stay byte-for-byte aligned with the harness, but nothing guards the default — the existing test only checks an explicitly-passed `stop`. Add a pin:

```python
def test_default_stop_matches_lm_eval_harness():
    assert STOP_SEQUENCES == ("\\nclass", "\\ndef", "\\n#", "\\nif", "\\nprint")
```

sieval/tasks/human_eval_0shot_base_gen.py
```diff
@@ -0,0 +1,220 @@
...
+``max_gen_toks=1024``. Set these through the model ``args`` or per-task
+``infer_args`` in the run config. The task owns only the prompt-coupled
+``stop`` sequences and ``n`` (the pass@k sampling count).
+
+HumanEval has no single canonical Qwen2.5-72B-Base score: references span
+roughly six points — DeepSeek-V3 Table 3 reports 53.0, while the Qwen2.5
+technical report self-reports 59.1 (with no published eval config). Because
+this task reproduces lm-eval-harness rather than Qwen's own setup, its score
+is expected to land between those references rather than match the Qwen
+self-report; the full 164-item run scores 56.1 (92/164, fails=0).
```
ethan-scitix: **Nit:** Drop the hardcoded `56.1 (92/164, fails=0)` from the docstring — it's a run-specific, unverifiable number that goes stale on any model/config change. Keep only the essential alignment notes; move reproduction results/config to a follow-up doc.

ethan-scitix (commented):
Reviewed against `humaneval.yaml` @ `1dd9310` — faithful port: prompt (`{{prompt}}`), stop sequences, raw-completion filtering, `check_program` structure, `repeats=1`, and pass@k all match. The 3 deltas vs the chat sibling (raw prompt / `stop` / no markdown extraction) are exactly what base generation requires. `index.json` consistent (`sync_meta_index.py --check` passes), reference URL commit-pinned, `deps_group: null` correct, tests pass. Two non-blocking nits inline — clean to merge after.

ethan-scitix (changes requested):
Faithful port overall — prompt, stop sequences, `repeats=1`/`n=1`, raw-completion filtering, the pinned HF revision, and the reference URL all check out, and the 4 new tests pass. Two issues before merge:

### `sieval/tasks/human_eval_0shot_base_gen.py`

**Bug:** `:154-159` — `timeout` (default `5.0`) is never sent to the code-eval service. The payload is only `{uuid, source, code}`, but the service accepts a `timeout` field and otherwise defaults to a hardcoded **3.0s** for human-eval python (`submodules/code-evaluator/app/server.py:70,83,88`). So `self._timeout` only sizes the httpx deadline (`+2`); the sandbox always runs at 3.0s, and configuring it does nothing. Fix: add `"timeout": self._timeout` to the payload. Pre-existing in `human_eval_0shot_gen.py:121` and `livecodebench_code_generation_0shot_gen.py:153` too — worth fixing all three in one pass.

**Nit:** `:105` — `self._max_concurrency` is stored but never read (the cap uses the `max_concurrency` param directly at `:111`). Drop it.

jack-scitix-ai:
Thanks, both fixed across all three code-exec tasks in one pass.

- **Sandbox timeout**: added `"timeout": self._timeout` to the payload in `human_eval_0shot_base_gen.py`, `human_eval_0shot_gen.py`, and `livecodebench_code_generation_0shot_gen.py`. Verified against the pinned submodule (`e480226`): `Sample.timeout: float | None = None` and the python executor defaults to `3.0` when it's absent, so the sandbox now honors the configured value instead of the hardcoded default.
- **Dead attribute**: dropped `self._max_concurrency` in all three — the `httpx.Limits` cap already uses the `max_concurrency` param directly.

ethan-scitix (changes requested):
**Request changes.** Faithful reproduction overall (verified the port against lm-evaluation-harness `humaneval.yaml` at the pinned SHA — prompt, the five `until` stops, raw-completion filtering, `check_program`, pass@k all match; 4/4 unit tests pass). One blocking issue in the bundled timeout change, plus a faithfulness nit. The HumanEval base task itself is good.

---

**`Bug:` LiveCodeBench — flat timeout breaks per-test-case scaling (`sieval/tasks/livecodebench_code_generation_0shot_gen.py:154`)**

Sending a flat `"timeout": self._timeout` (6.0) overrides the evaluator's scaled default `6 + 2·N` (`code-evaluator app/server.py:134`), where all N test cases run sequentially under one budget. For any problem with N≥1 tests the budget strictly shrinks (e.g. N=20: 46s → 6s), so slow-but-correct solutions now hit `subprocess timeout` → LiveCodeBench scores can only drop. Note the HTTP wall-clock at `:158` *still* scales (`self._timeout + len(inputs)*2 + 2`), so the two are now contradictory. Official LCB is per-case `timeout=6` + global `(6+1)·N+5` join. Fix: send the **scaled** value `self._timeout + len(inputs) * 2.0` (base `self._timeout=6.0`), and add a one-line comment that this approximates official per-case-6s within a single sequential-execution budget.

**`Nit:` HumanEval — forwarded timeout deviates from the official 3.0 (`sieval/tasks/human_eval_0shot_base_gen.py:99,157` + `sieval/tasks/human_eval_0shot_gen.py:66,125`)**

Default `timeout=5.0` is now actively sent, but the official value is 3.0 across both cited references (simple-evals `humaneval_eval.py:24`, openai/human-eval `evaluation.py:43`, and HF `code_eval` used by lm-eval-harness). Previously the field wasn't sent, so the evaluator applied its own 3.0 (`app/server.py:83`) — i.e. this PR moves *away* from official. Score impact is ~nil (solutions finish <1s), but it's an unannotated deviation in a reproduction task. Set the default to `3.0` (single execution → flat is correct) so the contract is explicit and faithful.

**Context (why explicit > dropping it):** forwarding timeout from the task is the right call — `stop`-style execution-contract params belong in the task layer, and relying on the evaluator's defaults leaves the reproduction contract implicit in a separate repo. The issue is only the **values**: HumanEval should be flat `3.0`, LiveCodeBench must stay **scaled**. The `_max_concurrency` dead-attr removal is good — keep it.

---

Non-blocking: `_pass_at_k` + the feedback/report bodies are now duplicated across the two HumanEval tasks (and LCB). Project convention allows task duplication, so leave it for now; worth a family-wide refactor once the base-model task batch lands. The gsm8k decode-param drift I'd have flagged is already handled in #3.

ethan-scitix (approved):
Can be merged after rebasing

PR #10, opened 2026-06-22
feat(cmmlu): add CMMLU few-shot base-model task and dataset

jack-scitix-ai:
## Type

## Summary

- Add **CMMLU** (Chinese multi-domain MCQ, 67 subjects): dataset + few-shot base-model task.
- `CMMLUDataset` loads the pinned official GitHub archive (zip or CSV root); fails fast on malformed rows / empty loads instead of emitting silent blank samples.
- `CMMLUFewShotBaseGenTask` (`model_type=gen`): same-subject dev shots, one-call next-token A/B/C/D scoring via `GenModel.alogprobs` `top_logprobs`, subject-level macro report.
- Reference: official CMMLU `qwen2.py` **`eval`** (base) path — https://github.com/haonan-li/CMMLU (commit `d6e7b716d8ac694f38969a6c0407437d1fded799`)

###  core/ changes (please review)

Modifies `sieval/core/models/` — required because the task needs the per-token candidate distribution, which existing `ModelOutput` fields can't carry. Two files, one data path (`GenModel` writes → `ModelOutput` holds → CMMLU task reads):

- **`core/models/model.py`** — add one optional field `ModelOutput.top_logprobs: list[dict[str, float]] | None = None`. Generic / benchmark-agnostic, defaults to `None` (no behavior change to existing callers).
- **`core/models/gen_model.py`** — extract `top_logprobs` from the completions response (`_completion_top_logprobs`) and populate the new field. Also switches the two `create` calls from `cast(Any, ...)` to the inline `# ty: ignore[no-matching-overload]` style (reviewer-preferred; matches the chat backend). The chat backend itself is **intentionally untouched** — CMMLU is `model_type=gen`.

No new upper-layer dependencies added to `core/`.

### Scoring method & alignment
**What we reproduce.** This task mirrors the official CMMLU `qwen2.py` **`eval`** (base-model, letter-argmax) path — non-CoT prompt, same-subject dev shots, and next-token A/B/C/D scoring. (The official script routes base models to `eval`, instruct models to `eval_instruct`; we follow the base path.) Mechanic detail: official `eval` argmaxes raw last-token **logits** over the A/B/C/D token IDs; we argmax over `top_logprobs` for the same tokens. These are identical whenever all four option tokens are within the requested top-k (softmax is monotonic); the only divergence is a token outside top-k (scored `-inf`). The validated run (`logprobs=100`) had **0 missing** A/B/C/D entries across all 11,582 samples.

**1. No official base-model number to validate against.** We reproduce the official method, but the official CMMLU leaderboard publishes only **Qwen2.5-72B-Instruct** (85.67, 5-shot) — there is **no official base-model score** for Qwen2.5-72B. So the official leaderboard can confirm our *method* but not our *base-model number*.

**2. DeepSeek supplies a base number; its method is vague but inferable.** The DeepSeek-V3 report's base-model table lists Qwen2.5-72B **Base** at **89.5** (5-shot), which we adopt as the reference. DeepSeek describes its mechanic only loosely (*"calculating the perplexity of each option and selecting the lowest"*, *"length normalization"*; letter-vs-text unstated). But its V2 report's appendix template lists `OPTIONS: A/B/C/D` (single letters) — which strongly suggests its scoring is **letter-based**, i.e. essentially the same as the official `qwen2.py eval` and this task.

**3. The close score corroborates that inference.** Our base run scored **89.92**, within **+0.42** of DeepSeek's 89.5. This agreement is consistent with — and provides empirical support for — the inference in (2) that DeepSeek's CMMLU scoring is effectively the same letter-argmax method we reproduce.

### Automated

- [x] Type check clean (`ty check`) — **red on a pre-existing issue, see note**

### Manual

- [x] Qwen2.5-72B **Base**, 5-shot, sglang, `deterministic: true`, `logprobs: 100`, full CMMLU test set (11,582 samples, 0 failures).

| --- | --- | --- | --- |
| Qwen2.5-72B Base (5-shot) | 89.5 (DeepSeek-V3 base table) | 89.92 | +0.42 (+0.42%  |

Per-category: STEM 85.16 · Humanities 93.37 · Social Science 89.92 · Other 92.34 · China-specific 89.78.

Note: all 11,582 samples flag `truncated_output` — expected, not a defect. The task requests `max_tokens=1` for next-token choice scoring, so every output hits the cap by design; scoring unaffected (`fails: 0`).

### Required (all PRs)

- [x] Dataset loading tested (`sieval dataset download cmmlu` succeeds)

### DIFF
--- a/sieval/tasks/cmmlu_kshot_base_gen.py
+++ b/sieval/tasks/cmmlu_kshot_base_gen.py
@@ -32,6 +32,13 @@ method this task reproduces. The reproduced method is the official CMMLU
 ``qwen2.py`` ``eval`` path above. Note the official CMMLU leaderboard's
 "Qwen2.5-72B" (85.67) is the *Instruct* model, not this base-model target.
 
+Few-shot selection takes the first ``k`` same-subject dev examples in order and,
+unlike the official ``gen_prompt``, does not drop the longest shots to fit a
+token budget. For CMMLU this is a no-op divergence: the official cap is
+``max_length=2048``, the dev pool is 5 examples/subject, and the longest full
+5-shot prompt across all 11,582 test rows (Qwen2.5 tokenizer) is 802 tokens — so
+the official truncation never fires and the assembled prompts are identical.
+
 AI-Generated Code - GPT-5.5 (OpenAI)
 """
 
@@ -426,32 +433,28 @@ class CMMLUFewShotBaseGenTask(
             if ctx.feedback_result and ctx.feedback_result["correct"]:
                 subject_corrects[subject] += 1
 
-        if not subject_totals:
-            return {"score": 0.0, "fails": float(len(fails)), "pass@1": 0.0}
-
         subject_acc = {
             subject: subject_corrects.get(subject, 0) * 100 / total
             for subject, total in subject_totals.items()
             if total
         }
-        overall = sum(subject_acc.values()) / len(subject_acc)
+        overall = sum(subject_acc.values()) / len(subject_acc) if subject_acc else 0.0
 
         metrics: dict[str, float] = {
             "score": overall,
             "fails": float(len(fails)),
-            "pass@1": overall,
             "overall": overall,
         }
 
+        # Only report categories that actually have evaluated subjects, so a
+        # subject-subset run omits absent categories instead of reporting a
+        # misleading 0.0 (indistinguishable from "all wrong").
         for category, subjects in CMMLU_CATEGORY_SUBJECTS.items():
             available_scores = [
                 subject_acc[subject] for subject in subjects if subject in subject_acc
             ]
-            key = category.lower().replace(" ", "_")
-            metrics[key] = (
-                sum(available_scores) / len(available_scores)
-                if available_scores
-                else 0.0
-            )
+            if available_scores:
+                key = category.lower().replace(" ", "_")
+                metrics[key] = sum(available_scores) / len(available_scores)
 
         return metrics
--- a/sieval/tasks/cmmlu_kshot_base_gen.py
+++ b/sieval/tasks/cmmlu_kshot_base_gen.py
@@ -6,9 +6,20 @@ Mirrors the official CMMLU ``qwen2.py`` *base* path (``eval``, not
 A/B/C/D choice scoring. Here scoring reads one next token from ``top_logprobs``
 and argmaxes over A/B/C/D; the official ``eval`` argmaxes raw last-token logits
 over the same token IDs. The two agree whenever all four option tokens are in
-the requested top-k (softmax is monotonic) — the only divergence is an option
-token outside top-k, scored ``-inf`` here. In the validated Qwen2.5-72B 5-shot
-run with ``logprobs=100``, all 11,582 samples had finite A/B/C/D scores.
+the requested top-k (softmax is monotonic). To keep partial coverage loud
+rather than silent, scoring *requires all four* option tokens to be present and
+raises otherwise — so a too-small top-k surfaces as a sample failure instead of
+a best-of-present guess. In the validated Qwen2.5-72B 5-shot run with
+``logprobs=100``, all 11,582 samples had the full A/B/C/D set.
+
+Infra requirement for faithful reproduction: the serving backend must return a
+top-k large enough to always include A/B/C/D. SGLang serves ``logprobs=100`` out
+of the box; on vLLM start the server with ``--max-logprobs 100`` (its default is
+20, which can drop option tokens). ``DEFAULT_LOGPROBS`` is set to 100 to match.
+
+Option matching is by the *stripped token string* (e.g. ``" A"`` → ``"A"``),
+since the OpenAI-compatible API exposes token text rather than the fixed token
+IDs the official ``eval`` indexes; this only differs in tokenizer edge cases.
 
 Target: 89.5 — Qwen2.5-72B *Base*, 5-shot, from the DeepSeek-V3 report's
 base-model table; benchmarked against this task's ``overall`` (subject-level
@@ -38,7 +49,9 @@ from sieval.datasets import CMMLUDatasetSample
 
 CHOICES = ("A", "B", "C", "D")
 DEFAULT_N_SHOT = 5
-DEFAULT_LOGPROBS = 20
+# Must be large enough that all of A/B/C/D land in the returned top-k. The
+# validated run used 100; vLLM needs `--max-logprobs 100` (default 20).
+DEFAULT_LOGPROBS = 100
 
 CMMLU_SUBJECT_DISPLAY_NAMES = {
     "agronomy": "农学",
@@ -225,17 +238,23 @@ class Feedback(TypedDict):
 def _choice_scores_from_top_logprobs(
     top_logprobs: list[dict[str, float]] | None,
 ) -> tuple[dict[str, float], bool]:
+    """Map the first token's top-k onto A/B/C/D logprobs.
+
+    Returns ``(scores, all_present)`` where *all_present* is ``True`` only when
+    every one of A/B/C/D appears in the top-k. Partial coverage is reported as
+    ``False`` so the caller can fail loudly rather than argmax a subset.
+    """
     scores = {label: float("-inf") for label in CHOICES}
     if not top_logprobs:
         return scores, False
 
-    found = False
+    seen: set[str] = set()
     for token, logprob in top_logprobs[0].items():
         label = token.strip()
         if label in scores:
             scores[label] = max(scores[label], logprob)
-            found = True
-    return scores, found
+            seen.add(label)
+    return scores, len(seen) == len(CHOICES)
 
 
 @sieval_task(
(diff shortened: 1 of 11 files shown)


ethan-scitix (changes requested):
Verified the port against the official `qwen2.py`/`mp_utils.py`/`categories.py` at the pinned SHA, and against the validated run (`outputs/cmmlu_5shot_qwen2.5-72b`: 89.92 vs 89.5 target, fails=0, logprobs=100/k=5/temp=0, SGLang, deterministic). Prompt template, few-shot selection, subject-macro aggregation, categories, and data (byte-identical to upstream) all faithful. Findings below are minor — none is a confirmed correctness bug.

### `sieval/datasets/cmmlu.py`

- **Nit:** `@sieval_dataset` is missing `license` — `cmmlu` is the only registered dataset with `license=None`, and upstream is **CC-BY-NC-SA-4.0** (non-commercial, worth recording). Add `license="CC-BY-NC-SA-4.0"`. `cmmlu.py:34`
- **Question:** The `source` only ever stages the GitHub `{sha}.zip`, but the loader also supports the HF zip name (`cmmlu_v1_0_1.zip`), the no-`data/`-prefix layout, and a pre-extracted csv-root dir — none reachable from `sieval dataset download` → `load`. Is multi-source substitution intended? If not, trimming to the GitHub layout removes `_find_zip`/`_find_csv_root`/the raw-row fallback and simplifies the loader. `cmmlu.py:222-261`

### `sieval/tasks/cmmlu_kshot_base_gen.py`

- **Question:** The validated regime is `logprobs=100`, but the default is `DEFAULT_LOGPROBS=20` and `_choice_scores_from_top_logprobs` returns `found=True` on ≥1 option — so when fewer than 4 options land in top-k (likely at 20, esp. on vLLM where `--max-logprobs` defaults to 20), it silently scores best-of-present instead of flagging. Two asks: (a) document the infra requirement (`--max-logprobs ≥ 100` on vLLM; SGLang serves 100 out of the box) for faithful reproduction; (b) consider treating "<4 options present" as a failure so partial coverage is loud, not silent. `cmmlu.py:41,225-238,371-377`

### `sieval/core/models/gen_model.py`

- **Nit:** `_completion_top_logprobs` guards against `None` / non-`Mapping` entries and filters token/logprob types — inputs neither vLLM nor SGLang produce — and no test exercises those branches (core/ coverage rule). Either simplify to the shape backends actually return, or add a test feeding `[{"A":-0.1}, None, "x", {"y":-0.5,"1":"no"}]`. `gen_model.py:9-27`

_Optional:_ the docstring documents the top-k-vs-full-vocab divergence; could also note that scoring matches by stripped token string rather than the official's fixed token ID (API-forced, edge-case only).

jack-scitix-ai:
Thanks for the thorough pass — all points addressed. I additionally add a Pinning tests in `test_cmmlu_kshot_base_gen.py`. Summary of what changed and where:

**`sieval/datasets/cmmlu.py`**
- **License (nit):** added `license="CC-BY-NC-SA-4.0"` to `@sieval_dataset`; regenerated `sieval/meta/index.json` to match.
- **Multi-source layouts (question):** not intended — trimmed the loader to the GitHub `{sha}.zip` layout only. Removed the HF-zip name, the no-`data`-prefix path, the pre-extracted csv-root dir, and the raw-row fallback (`_find_zip`/`_find_csv_root` gone). `load` now accepts only the staged dir or a direct `.zip`. The malformed-row guard is kept (it now raises loudly on an unexpected CSV header instead of silently producing empty samples) and was inlined into `_process_sample` while collapsing the helper chain.

**`sieval/tasks/cmmlu_kshot_base_gen.py`**
- **Silent partial coverage (question b):** scoring now *requires* all of A/B/C/D in the top-k. `_choice_scores_from_top_logprobs` returns `(scores, all_present)` and `postprocess` raises a sample failure when any option is missing — so a too-small top-k is loud, not a best-of-present guess.
- **Infra requirement (question a):** `DEFAULT_LOGPROBS` bumped 20 → 100; the module docstring and `reference_impl.notes` now spell out the requirement (SGLang serves 100 OOTB; on vLLM start with `--max-logprobs 100`, default 20).
- **Token-ID divergence (optional):** docstring now notes scoring matches by the stripped token string rather than the official fixed token ID (API-forced; tokenizer edge-case only).

**`sieval/core/models/gen_model.py` → tests**
- **Defensive branches (nit):** kept the guards (cheap robustness) and added the coverage you suggested — `TestCompletionTopLogprobs` feeds `[{"A": -0.1}, None, "x", {"y": -0.5, "1": "no", 2: -0.3}]`, exercising the None→`{}`, non-Mapping skip, and non-str-key / non-numeric-value filter branches.

**Pinning tests (`tests/unit/tasks/test_cmmlu_kshot_base_gen.py`)**
- Added regression locks so a faithful-reproduction drift fails loudly: the exact prompt template (`_format_example`), the few-shot header + `anatomy → 解剖学` display-name lookup, the `CMMLU_CATEGORIES` partition, 67-subject coverage across `SUBCATEGORIES`/`DISPLAY_NAMES`, and a guard that every used subcategory maps to a category bucket (a typo there would silently drop subjects from category metrics).

sieval/tasks/cmmlu_kshot_base_gen.py
```diff
@@ -0,0 +1,457 @@
...
+            )
+        for sample in split:
+            subject = str(sample.get("subject") or "miscellaneous")
+            self._few_shot_by_subject.setdefault(subject, []).append(sample)
+
+    def _select_examples(self, subject: str) -> list[CMMLUDatasetSample]:
+        if self._k == 0:
+            return []
+        self._ensure_few_shot_pool()
+        return list(self._few_shot_by_subject.get(subject, []))[: self._k]
```
ethan-scitix, line 352: `Question:` Official `gen_prompt` drops the longest shots when a k-shot prompt exceeds `max_length` (2048); `_select_examples` always returns `[: self._k]`. Minor impact (only long-passage subjects like `college_law`/`ancient_chinese`), but this task annotates every other deviation — add a one-line docstring note, or confirm it's intentional.

sieval/tasks/cmmlu_kshot_base_gen.py
```diff
@@ -0,0 +1,457 @@
...
+            subject: subject_corrects.get(subject, 0) * 100 / total
+            for subject, total in subject_totals.items()
+            if total
+        }
+        overall = sum(subject_acc.values()) / len(subject_acc)
+
+        metrics: dict[str, float] = {
+            "score": overall,
+            "fails": float(len(fails)),
+            "pass@1": overall,
```
ethan-scitix: `Nit:` `pass@1` duplicates `score`/`overall` and isn't a CMMLU metric, isn't read by the leaderboard (only `score`), and isn't emitted by the sibling `gsm8k_kshot_base_gen`. Suggest dropping it (also the empty-result branch at L430).

sieval/tasks/cmmlu_kshot_base_gen.py
```diff
@@ -0,0 +1,457 @@
...
+
+        for category, subjects in CMMLU_CATEGORY_SUBJECTS.items():
+            available_scores = [
+                subject_acc[subject] for subject in subjects if subject in subject_acc
+            ]
+            key = category.lower().replace(" ", "_")
+            metrics[key] = (
+                sum(available_scores) / len(available_scores)
+                if available_scores
+                else 0.0
```
ethan-scitix: `Nit:` On subject-subset runs, categories with no evaluated subjects report `0.0` (indistinguishable from "all wrong"). `score`/`overall` stay correct — consider omitting absent categories instead of emitting `0.0`. Full runs unaffected.

ethan-scitix (commented):
Faithful reproduction confirmed against official CMMLU at pinned `d6e7b71` — prompt template, subject header, and the full subject/category taxonomy + Chinese display names all verified verbatim against `mp_utils.py`/`categories.py`; dev-set `[:k]` few-shot matches `dev_df[:num_few_shot]`; scoring helper spot-run; 63 unit tests pass. No blocking bugs.

Three minor notes inline. Only #1 (docstring) is worth doing before merge; the other two are small cleanups.

jack-scitix-ai:
Thanks — all three addressed.

1. Few-shot truncation (Question): Confirmed intentional, and I verified it's a no-op divergence for CMMLU rather than a behavioral difference. The official eval does pass tokenizer/max_length=2048 to gen_prompt, which drops the longest shots when the encoded prompt exceeds the cap. But measuring against the official archive with the Qwen2.5 tokenizer: the dev pool is 5 examples/subject and the longest full 5-shot prompt across all 11,582 test rows is 802 tokens (≪ 2048), so the official's shot-dropping never actually fires. Our prompts are therefore identical to the official at k=5. Documented this (with the measured numbers) in the docstring rather than re-implementing tokenizer-aware truncation that would never trigger.

2. Redundant pass@1 (Nit): Dropped, along with the redundant empty-result branch; report() now has a single return path (score/overall/fails + present categories), with overall guarded against an empty subject set.

3. Categories 0.0 on subset runs (Nit): Fixed — categories with no evaluated subjects are now omitted instead of reporting 0.0. A real 0.0 (an evaluated subject that scored 0) is still reported; covered by the updated test.

ethan-scitix (approved):
Can be merged after rebasing.

PR #11, opened 2026-06-22
feat(ruler): add RULER long-context benchmark (datasets + tasks)

Mea1Ma:
> **Note:** This PR is stacked on #9 (add local: source handler and fix url: gzip truncation check
). Those commits appear in this diff too because GitHub can't target a fork branch as the base. Please review the downloader changes in #9; once it merges I'll rebase this onto main to drop them. The RULER work here depends on the `local:` handler — the Paul Graham haystack is a bring-your-own corpus, generated out-of-band by `scripts/gen_paul_graham_essays.py` and never bundled or committed (see the `local:` handler docstring), staged via the `local:` scheme.

## Type

## Summary

Adds the **RULER** long-context benchmark (13 configs across 4 categories) as
synthetic datasets + thin generative tasks, adapted from NVIDIA RULER

- **datasets/ruler/** — a single `RulerDataset` (13 subtasks selected via
  `subtask=`), backed by 5 internal per-family loader modules (niah/qa/vt/cwe/fwe)
  that build the full prompt at load time. The haystack corpus (Paul Graham
  Essays) is a bring-your-own file, generated out-of-band by
  `scripts/gen_paul_graham_essays.py` and never bundled or committed, staged
  via the `local:` scheme.
- **tasks/** — a single `RulerZeroShotGenTask` (`n_shot=0`) that sends the
  prompt and scores by RULER's `string_match_all`/`string_match_part`. CWE/VT
  bake one in-context demonstration directly into the generated prompt text
  (upstream's own convention), not via sieval's `n_shot` mechanism.
- **community/ruler/** — vendored RULER constants/tokenizer/scoring, kept close
  to upstream with attribution.
- Decoding params stay in user config (tasks never inject them); HotpotQA source
  is revision-pinned for reproducibility.

Reference: https://github.com/NVIDIA/RULER 

## Test Plan

> **CI note:** `ty` is pinned to 0.0.20 in the lock for parity with `main`
> (an upstream merge had re-resolved the loose `ty>=0.0.5` spec up to 0.0.59,
> whose stricter checks flag ~150 pre-existing diagnostics repo-wide). RULER's
> deps (`tiktoken`/`wonderwords`/`nltk`/`scipy`/`transformers`) are absent from
> the light CI env, so the ruler tests are gated behind `@_needs_ruler_deps`
> and skip there.

### Automated

- [x] Type check clean (`ty check`; ty pinned to 0.0.20)
- [x] Unit tests pass — CI light env (no `ruler` extra):
  `pdm run pytest tests/unit -m "not stress" --cov` → **2433 passed, 39 skipped,
  98.44% coverage** (coverage scope is `sieval/core`). RULER tests requiring the
  `ruler` extra skip here; `pdm install -G ruler && pdm run pytest` runs the full
  suite (ruler tests green locally).

### Manual

Reproduces RULER aligned with NVIDIA RULER. Compared against the
RULER score in the Qwen3 technical report (arXiv:2505.09388) — note Qwen does
not publish the exact prompt/decoding/answer-extraction config, so the target
is "land close to the report" rather than byte-exact reproduction.

- [x] Model tested: Qwen3-8B, both non-thinking and thinking modes (matches the report's settings)
- [x] Coverage: full 4k–128k length sweep (4k/8k/16k/32k/64k/128k), 13-task avg, for both modes
- [x] Reference (Qwen3 tech report, RULER avg over 4k–128k): 89.1 (non-thinking) / 84.4 (thinking)
- [x] Actual (this implementation): 88.12 (non-thinking) / 86.86 (thinking)
- [x] Diff: -0.98 (-1.10%) non-thinking, +2.46 (+2.91%) thinking — both within target < 3%
- [ ] Caveat: the thinking-mode diff holds only with the known CWE truncation
      issue factored in (see `ruler_0shot_gen.py` note). CWE's answer budget
      (upstream's own 120-token cap, +4 Qwen3 tag overhead) was never
      validated by upstream against thinking-mode's more verbose output, so
      4k-64k CWE recall is truncation-limited, not capability-limited.
      Correcting those lengths to their completed-only recall moves the
      thinking mean to ~88.3 — a deviation of ~+4.6% from the 84.4 reference,
      outside the <3% target. Whether to raise the answer budget is a
      separate, unresolved decision (see follow-ups).
- [x] Runs are a dev build (`meta.json`: `version 0.0.0`, `deterministic: false`)

| Model | Mode | Expected | Actual | Diff |
| --- | ---:  | ---: | ---: | ---: |
| Qwen3-8B | Non-thinking | 89.1 | 88.12 | -0.98 (-1.10%) |
| Qwen3-8B | Thinking | 84.4 | 86.86 | +2.46 (+2.91%) |

CWE truncation detail (thinking mode, n=500 per length):

| length | 4k | 8k | 16k | 32k | 64k | 128k |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| truncated/500 | 469 | 482 | 484 | 468 | 437 | 363 |
| recall, completed | 100.0 | 88.9 | 95.6 | 82.8 | 55.9 | 13.0 |
| recall, truncated | 83.1 | 68.3 | 53.8 | 56.9 | 42.9 | 14.4 |

`niah_single_1` and `vt` truncate 0/500 on the same runs, so this is
CWE-specific verbosity, not a general thinking-mode issue. At 4k-64k the
completed/truncated gap is real budget starvation; at 128k the gap closes
(13.0 vs 14.4) — that length is genuinely capability-limited.

### Required (all PRs)

- [x] AI-generated code has `AI-Generated Code - <model> (<provider>)` marker

- [x] Dataset loading tested (`sieval dataset download` path; local/url/hf sources resolve)

### If: community/ Changes

- [x] Upstream diff documented (vendored from NVIDIA RULER, attribution in headers)
- [x] License attribution preserved (Apache-2.0 headers restored)

### If: New Dependency

- [x] Added to correct PDM dependency group (`ruler`: tiktoken, wonderwords, numpy, scipy, nltk, transformers)
- [x] Justified in Summary (tiktoken/wonderwords for prompt synthesis, scipy/numpy for sampling, nltk for sentence tokenization, transformers for the HF/Qwen3 tokenizer path)


ethan-scitix (changes requested):
**Review: RULER benchmark onboarding**

The port is faithful — I verified the generation logic against NVIDIA RULER at the pinned SHA, and the 6-length Qwen3-8B run reproduces published recall numbers closely (`cwe@16k` 93.5 vs 93.2, `vt`/`niah` ~100). Requesting changes on packaging and structure before merge.

### Blocking

- **Bundled corpus doesn't belong in the repo.** `sieval/datasets/_data/paul_graham_essays/PaulGrahamEssays.json.gz` (1.1MB) should not ship in-tree. RULER's corpus isn't an official artifact — NVIDIA ships only a URL list + scraper, so this `.gz` is just one local scrape frozen. Host one copy (HF `@sha`, or any immutable/public OSS object via `url:`) and keep `scripts/gen_paul_graham_essays.py` as the BYO/regeneration path. The loader is host-agnostic once the file lands in the data dir.
- **Overlaps with the still-open #9.** This branch re-includes #9's `downloaders/{base,local,url}.py` + tests verbatim. Land the `url:` gzip fix via #9; rebase this PR on top and drop those files. `local:` itself should become true BYO (no in-wheel bundling; `download` = no-op + instructions) rather than copy-from-package.
- **`nltk` missing from the `ruler` extra.** `pyproject.toml:39` lists `tiktoken/wonderwords/numpy/scipy` but not `nltk`, which the default niah/vt essay path imports (`_common.py:71`, `ruler_niah.py:285`, `ruler_vt.py:315`). `check_dep_coverage` passes only because `nltk` is in `ifeval`; a clean `pip install sieval[ruler]` breaks at load. Add `nltk` to the `ruler` group.

### Structure (the bigger one)

- **Collapse to 1 dataset + 1 task, mirroring MMLU.** The 5 datasets / 5 empty task classes + the new `leaderboard ruler-avg` command are all symptoms of the same gap. MMLU already models 57 subtasks as one dataset with a `Subject` column and aggregates per-group in `report()` (`mmlu_0shot_gen.py:92,102`). Do the same: one `RulerDataset` emitting a `subtask` column, one `RulerTask` whose `report()` groups by subtask (right matcher per group — `string_match_all` vs `string_match_part`) and emits per-subtask scores + the headline mean. The context-length dimension stays as separate runs (64k/128k need YaRN-scaled serving). This removes `_ruler_avg.py`, the `ruler-avg` command, and the `output.py` renderer entirely — no benchmark-specific logic in the CLI, no parsing length out of task-name strings.

### Other

- **Bug: CWE `english_words.json` pinned to `main`** (`ruler_cwe.py:44`), unlike the deliberately-pinned HotpotQA revision. It's the long-context vocab fallback used exactly at 64k/128k. Pin to a SHA.
- **Nit: Category.** `Logic/TextualReasoning` is a poor fit for retrieval/aggregation tasks. Move to `Language/SemanticUnderstanding`; the `long-context` tag already carries the capability axis.
- **Reproducibility:** ship at least one example eval config (now small — one ruler task per length) and document the non-obvious prerequisites: `continue_final_message: true` (without it the answer-prefix continuation silently breaks), YaRN for >32k, and `tokenizer_path` must match the target model (RULER sizes prompts to the model's tokenizer). The config that produced the run was dropped in `4c1d1749` and `examples/` has none.

(Dockerfile / deps-in-image handled separately.)

ethan-scitix (commented):
**Verdict: Comment — port is faithful and reproduction checks out; a few consistency/onboarding items before merge.**

Verified empirically: `string_match_all`/`string_match_part`, templates, and token budgets are byte-exact to NVIDIA/RULER at `ab17b78`; the vendored matcher reproduces the stored subtask scores exactly (cwe=98.54, qa_squad=76.8, fwe=97.53), and the per-length means average to ≈88, consistent with the reported 88.2.

**Should fix before merge**

- `Bug` (consistency): the per-sample retry loops differ across siblings. Upstream RULER's niah/qa/vt/cwe *all* lack the floor `else` (all can spin at `used <= incremental`). This port added `else: break` to `_niah.py:159`, `_vt.py:155`, `_cwe.py:84` but **not** `_qa.py:81` — so three deviate from upstream and one matches it. Practically unreachable at 4k+, but pick one behavior for all four: either match upstream exactly (drop the `else`), or keep the guard everywhere and annotate the deviation in code. Author's call, but it must be consistent.
- `Nit` (architecture): `_ChatGenBase` (`ruler_0shot_gen.py:51`) has a single subclass. Inline `preprocess`/`infer`/`postprocess` into `RulerZeroShotGenTask` and delete the base.
- `Nit` (encapsulation): `ruler_0shot_gen.py:73,85` reads `self.model._kwargs`/`_model`; use the public `self.model.meta()` instead.
- `Nit` (robustness): `sieval dataset download ruler` swallows `LocalSourceUnavailable` and exits 0 (`cli/dataset/commands.py:190`). Track missing required `local:` sources and print a clear final summary + exit non-zero, so a missing corpus isn't mistaken for success.
- `Nit` (onboarding): the BYO corpus step is only in the script docstring / error message, not in the config setup or docs, and the generator's deps (`html2text`, `beautifulsoup4`, `certifi`) are undeclared — so following the example config fails at load time, and running the script fails on import. Please: (1) document the full flow (install extra → generate corpus → `download` the rest → run), (2) declare the generator deps (e.g. a `ruler-gen` optional group), and (3) have `download` print the exact generate command rather than auto-scraping.

**Nits**

- Revert the `.pre-commit-config.yaml` `--maxkb=2000` bump — the corpus is BYO/not committed, so it's vestigial and its comment (references bundling under `sieval/datasets/_data/`) is misleading.
- ty overrides (`pyproject.toml`): narrow `sieval/datasets/ruler/**` to the specific loader files (or install the `ruler` extra in the type-check env); drop the `scripts/` exemption and declare its deps instead — exempting it hides that the script isn't runnable out of the box.
- `wonderwords.random_word._get_words_from_text_file` (`_niah.py:224`, `_cwe.py:144`): private API, faithful to upstream — keep, but add a comment noting the risk and cap the version (`<3`).
- Trim examples to `nonthinking` + `thinking` (flat, no subdir); fold the 64k/128k yarn variants into a comment.

**PR body**

- "Actual: 87.7" contradicts the table's 88.2 (89.1 − 0.9 = 88.2, −1.01%) — 88.2 is the real number.
- Summary says answer_prefix is appended to the user message "without a prefilled assistant turn," but the shipped nonthinking config and the reproduction runs use the assistant-prefill path (`continue_final_message: true`). Align the description with what was validated.

ethan-scitix (commented):
### Must fix before merge

- **`Bug:` `transformers` missing from the `ruler` extra.** `pyproject.toml:52` — both examples use `tokenizer_type: hf` (and faithful repro *requires* the model tokenizer, not `cl100k_base`), which triggers `from transformers import ...` in `sieval/community/ruler/scripts/tokenizer.py:34`. Not pulled by any `ruler` dep → clean `pip install sieval[ruler]` hits ImportError. Masked here only because `t-eval` pulls transformers transitively; `check_dep_coverage` misses it (import is in `community/` behind a method). Add `transformers` to the `ruler` group.

### Cheap cleanups (this PR)

- **`Nit:` Restore the `scripts/**/*.py` T201 exemption.** `pyproject.toml` — removing it forced per-line `# noqa: T201` on unrelated `scripts/check_layer_imports.py:280` and `scripts/check_preflight.py:1265,1267`. Keeping the blanket rule also lets `gen_paul_graham_essays.py` drop its noqas.
- **`Nit:` Drop internal branch names from a shipped docstring.** `sieval/datasets/ruler/_shared.py:102-103` — "feat/ruler" / "feat/ruler_exp" mean nothing to OSS readers; describe the two message patterns behaviorally.

### Confirm intent (+comment), not necessarily code changes

- **`Question:` Generation appears uncapped.** `sieval/tasks/ruler_0shot_gen.py:118` — `agenerate(pre)` sets no `max_tokens`; upstream caps per task (128/30/120/50/32). Can only inflate `string_match_*`. Repro is within target so impact is small, but it's an unannotated divergence in a `stable` port — cap via `infer_args` or note it.
- **`Question:` Sibling `except` mismatch.** `_qa.py:80` catches `except AssertionError` while `_niah.py:156`/`_cwe.py:81`/`_vt.py:152` catch `except Exception`. In QA a `num_docs < len(curr_docs)` fit raises `ValueError` (`random.sample(curr_more, negative)`) that QA's handler won't catch → propagates instead of shrinking. Edge case; align or comment.
- **`Question:` Thinking-example budget decoupling.** `examples/ruler-qwen3-8b-thinking.yaml` — dataset `think_budget` defaults 0 while serving sets `thinking_budget: 8192`; context reservation ignores the 8192 and only fits because `context_length` (32k) is 2× `max_seq_length` (16k). `max_seq_length: 32768` would overflow. Comment the coupling.
- **`Question:` `status: "stable"` with a checksum-free BYO corpus.** `sieval/meta/index.json` — PG essays fetched live, no checksum, drift undetectable. Defensible (needles/QA don't depend on filler text; repro within target) — flagging per the non-strict-repro rule. Also PR body says "Actual: 89.1" but the table says 88.2 (typo).

ethan-scitix (commented):
### Must address (both are unresolved prior-review comments)

**1. `Bug` — `_vt.py:155-156` still has `else: break`** *(carryover from Review 2, 07-03)*
Commit `4b3f8db6` set out to drop the guard from all four loops (message: *"remove `else: break` … niah/qa/vt/cwe allow unbounded retry"*) but only touched `_niah.py`/`_cwe.py`. `_vt.py` is now the lone deviant, unannotated — `_niah.py:156`, `_cwe.py:81`, `_qa.py:89` all match upstream, `_vt.py:155` doesn't. Reviewer's "must be consistent" is still unmet. Fix: delete the two lines.

**2. `Question` — `status="stable"`** *(carryover from Review 3, 07-07)*
`@sieval_task` at `ruler_0shot_gen.py:51` defaults to `stable`. Per the non-strict-repro rule, either mark `experimental` or enumerate all divergences in `ReferenceImpl.notes` (`:65-66` currently covers only scoring).

### New correctness/faithfulness (this round)

**3. `Bug` — CWE cap defeats `english_words.json`.** `_cwe.py:119-125` caps `num_words` at the wonderwords pool (8169 words), so `randle_words` (`_cwe.py:166`, the only consumer of `english_words.json`) is dead code and CWE underfills at 32k/64k/128k (needs ~16k/32k/65k words). This nullifies the very fallback Review 1 pinned "for 64k/128k." Fix: drop the cap. (Only 16k is empirically checked — 93.5 vs 93.2; 64k/128k unverified.)

**4. `Question` — HotpotQA doc ordering diverges.** `_qa.py:183` uses insertion order; upstream `qa.py:102` uses `sorted(set(...))` (alphabetical) → different context bytes per seed. Unannotated. Fix: `sorted(total_docs_set)`, or annotate. (SQuAD is faithful.)

### Nits

**5. `Nit` — `report()` aggregation untested.** The headline mean-of-means (`ruler_0shot_gen.py:139-172`) has no discriminating test; only FWE exercises a real loader.

**6. `Nit` — unrelated infra churn.** `_resolve_run_models` (`leaderboard/commands.py:28-37`) is a single-caller extraction with no RULER connection (leftover from the removed `ruler-avg` command). Drop or split.

**7. `Nit` — `_vt.py:58` hardcodes `remove_newline_tab=False`** for ICL vs upstream applying the flag; only matters under non-default `remove_newline_tab=True`.

### Priority
Items 1 and 2 are one-liners closing standing review comments — do regardless. Item 3 is the one real new correctness call (fix-and-re-validate, or accept-and-mark-`experimental`, same fork as item 2). Items 4–7 don't block merge.

ethan-scitix (commented):
**Re: shipping `status="stable"`**

Agreed on stable — bar is: aligned with upstream, divergences explained by a repro result, no bugs. Two bugs + the diff-documentation are what's left; the three open comments from the last round map onto these.

**No bugs** — both found by running the loader end-to-end (staged corpus + `tiktoken`/`wonderwords`, all 13 subtasks):

- `Bug:` single-subtask configs crash. `ruler.py:233` only appends `/ruler` in the `all`/`list` branches, so a direct `subtask: niah_single_2` (any essay/squad subtask) → `FileNotFoundError`. `subtask: all` is fine, so the examples work, but a single-subtask config silently can't find its data. Fix by centralizing the path map — also removes the ~50-line `all`/`list` duplication.
- `Bug:` `qa` raises `ValueError` at `max_seq_length < 4096`. `_qa.py:229` — HotpotQA's gold-doc count is a fixed 10, so once 10 docs don't fit, `used_docs` underflows negative and the narrow `except AssertionError` (`_qa.py:80`) lets it propagate. The "unreachable in practice" annotation is wrong; fix the guard + comment. 4k+ is safe (verified), so not urgent — but a `stable` benchmark shouldn't ship a crash path.

**Align + explain the diff** — this is what earns `stable` over `experimental`:

- `ReferenceImpl.notes` (`ruler_0shot_gen.py:65`) must enumerate every divergence with why it's score-neutral: HotpotQA insertion-order vs upstream `sorted(set())` (`_qa.py:183`), essay concat order html-then-repo vs upstream repo-then-html (`gen_paul_graham_essays.py:143` — its "bytes match upstream" comment is currently false), and QA uncapped `max_tokens` (`string_match_part` can only inflate).
- Attach the repro table to the PR (the 16k Qwen3-8B run — cwe 93.5 vs 93.2, mean 88 vs 88.2). That's the "result explaining the diff" stable needs; today it only lives in review threads.
- CWE cap (`_cwe.py:119-125`): verified it does **not** bite ≤32k (fills 98–99%), only 64k/128k. Either drop the cap so all lengths stay faithful, or scope the stable claim to ≤32k in the notes.

Land these and `stable` is well-supported. (Checksums re-verified; `all`-path 13 subtasks + the `preprocess→feedback→report` chain all run clean.)

ethan-scitix (commented):
**`Bug:` `qa_hotpotqa` never reads the staged snapshot; silently falls back to a live online fetch.** `ruler.py:56` routes it to the base data dir, so `_qa.py:171` calls `load_dataset("<data_dir>", "distractor", …)` — which raises `ValueError: BuilderConfig 'distractor' not found` (reproduced against the real staged layout), caught at `_qa.py:172` → always drops to online `load_dataset("hotpotqa/hotpot_qa", …)`. So the pinned copy `download` stages is dead, and eval fails under `HF_HUB_OFFLINE=1` despite a completed download. Point the loader at `<data_dir>/hotpotqa/hotpot_qa` and keep online as an explicit last resort.

**`Nit:` unrelated churn in `leaderboard/commands.py:64-73`.** Loop→comprehension rewrite of `report()`, no RULER connection — leftover from the removed `ruler-avg` command. Please restore `origin/main`'s version so the PR touches no unrelated CLI file.

**`Nit:` two of the four external sources are downloaded but never consumed.** Besides HotpotQA above, `english_words.json` is loaded into `randle_words` (`_cwe.py:42-45`) but never sampled — the CWE cap keeps `num_words ≤ vocab`. This is the documented ≤32k `stable` scope, so it's fine; just add one line to the source docstring so nobody assumes it's load-bearing.

**`Nit:` `_read_hotpotqa` builds a dict it only uses as a set** (`_qa.py:180-189` — values discarded by `sorted(set(...))`). Use a set comprehension.

**Housekeeping:** rebase onto `main` to drop the closed-#9 downloader commits (branch is also behind #30/#34/#35/#37). `ruler-gen` generator deps stay undeclared — a conscious call, but running the script still `ImportError`s with no install hint; a one-line `pip install …` in the usage/error path would close it.

ethan-scitix (commented):
**To address**

1. `Question:` **Thinking long-context reproduction is unverified.** In the real runs, think-32k-YaRN collapsed (overall 6.89, everything 0 except fwe) and think 64k/128k have no report. This is *not* a port bug — the same code gives non-think-32k=92.3 and think-4k/8k are fine, so it's the 32k+YaRN serving config. But the "reproduces published recall" claim in `reference_impl.notes` (`sieval/tasks/ruler_0shot_gen.py:62`) is currently only backed by non-think + think-≤8k. Please **re-run think ≥32k to confirm**, or **scope the `status="stable"` / repro claim to the validated configs** and mark thinking long-context experimental.

2. `Nit:` **The divergence notes read as exhaustive but omit 4 (all score-neutral, confirmed by the runs)** — add to the notes or align to upstream:
   - `estimated_max` bounding formula differs from upstream (`sieval/datasets/ruler/_niah.py:210` / `_vt.py:246` / `_cwe.py:123` / `_qa.py:133`) → different binary-search probe trajectory → different global-RNG consumption → samples not byte-reproducible vs upstream. Prefer aligning to upstream (`max_seq_length -= model_template_token` up front, use the raw token count).
   - HotpotQA is sourced from the HF mirror, not upstream's derived json; questions are taken in native order and sliced to the first N, so per-index selection may differ (`sieval/datasets/ruler/_qa.py:182`). The notes only cover doc ordering.
   - essay `escape_snob=True` (`scripts/gen_paul_graham_essays.py:131`) escapes markdown, but upstream's `escape_all` is a no-op (unescaped — confirmed in the staged corpus) → diverges from upstream's actual output. Either revert to unescaped, or record it as a divergence and note the repro corpus must be regenerated.
   - VT ICL fragment is sized against 500, not 500−T (`sieval/datasets/ruler/_vt.py:128`, currently only in a helper docstring).

3. `Nit:` CWE drops sharply in thinking mode under the cap (non-think 8k 98.3 → think 8k 69.1); predictions are verbose lists truncated mid-answer. Upstream's cap=120 assumes a terse answer — worth noting its effect on thinking-mode CWE in the notes.

4. `Nit:` Inconsistent qwen3 detection: `thinking_prefill` uses substring `in` (`sieval/datasets/ruler/_shared.py:164`) while `get_template` / `tokens_to_generate` use `.startswith` (`:224` / `:112`). For non-standard model ids, packing and inference can disagree on the `<think>` block — align them.

ethan-scitix (commented):
Verified empirically: ran NVIDIA/RULER's own generators at `ab17b78` and byte-compared against this port — **`input`/`outputs`/`answer_prefix`/`length` identical for all five families** (CWE @32k+128k, VT @32k, NIAH multivalue+multikey_3 @32k, QA-squad @32k, FWE @32k). `7d05c37e`'s byte-reproducibility goal is achieved, and `d606fea2` fixes the CWE underfill (99.8% fill @128k). Checksums re-verified against live downloads. Preflight green, 110 ruler tests pass.

Also confirmed the thinking sweep is complete — `128k_yarn/ruler_128k/progress.json` reads `2109/6500, finished: false`, but that's a stale snapshot: all 6500 sample ids are present in `0/final/*.idx`, and the six thinking reports average to exactly 86.86.

### Must fix before merge

**1. `Bug:` the CWE thinking-mode note has the mechanism wrong** — `sieval/tasks/ruler_0shot_gen.py:88-94`

It says "reproduces upstream's own cap … not a bug", but upstream has no thinking mode — there is no upstream cap here. From the stored records: `max_tokens = gen_budget + think_budget = 124 + 8192 = 8316`, sglang's budget processor forces `</think>` at 8192, so **the answer gets 124 tokens** while a 10-item markdown list needs ~150. Measured on the six thinking runs:

| | 4k | 8k | 16k | 32k | 64k | 128k |
|---|---|---|---|---|---|---|
| truncated | 469/500 | 482/500 | 484/500 | 468/500 | 437/500 | 363/500 |
| recall completed / truncated | **100.0** / 83.1 | 88.9 / 68.3 | 95.6 / 53.8 | 82.8 / 56.9 | 55.9 / 42.9 | 13.0 / 14.4 |

Same runs: `niah_single_1` and `vt` truncate 0/500. So ≤32k CWE is budget-starved, not capability-limited; 128k genuinely is (no gap). Correcting ≤64k moves the thinking mean 86.86 → ~88.3, i.e. +2.91% → ~+4.6% vs the 84.4 reference — **outside the stated <3% band**. Please rewrite the note around the real mechanism, include the truncation rate, and split ≤32k from 128k. This is the one place a bespoke divergence has material score dependence, so it needs stating plainly rather than as expected behaviour. (Whether to actually give the answer a real budget is a separate call — see follow-ups.)

**2. `Bug:` `escape_snob=True` diverges from upstream's actual output** — `scripts/gen_paul_graham_essays.py:131`

Verified on html2text 2025.4.15: `escape_all` is not a real config attribute, so upstream's setting is a silent no-op and **upstream's corpus is unescaped**. Converting `paulgraham.com/avg.html` both ways gives 0 vs 39 escape sequences, and the staged corpus behind the repro numbers has 0 — i.e. it predates `46cb8920`. Please drop the line (keep the `html2text<2026` cap and a comment on why `escape_all` is not set) and remove the clause from note 1. With everything else byte-exact this is the only remaining avoidable divergence, and the shipped script currently cannot regenerate the corpus the repro table was measured on.

**3. `Bug:` `context_length != 131072` encodes the example's serving topology into the loader** — `sieval/datasets/ruler/_shared.py:121-123`

That is not a Qwen3 property; it is "the served window is larger than `max_seq_length`", which holds for 4k–64k in the example configs and fails at 128k (the config's own "Budget coupling" note says as much). Serve 32k natively and run `ruler_32k` and the loader skips the reserve → prompt + 8192 thinking tokens overflow the real window. Either make it an explicit `load()` parameter the YAML sets (behaviour-preserving, samples unchanged), or guard the `max_seq_length == context_length` case and document the assumption.

**4. `Bug:` qwen3 detection now misses the canonical HF id** — `sieval/datasets/ruler/_shared.py:164`

Aligning `thinking_prefill` to `.startswith` matched `:112`/`:224`, but all three now miss `Qwen/Qwen3-8B` and `/models/Qwen3-8B` — both common `--served-model-name` values, and `preprocess` feeds the *served* id (not the YAML `model_name: qwen3`) into `thinking_prefill`. A miss drops the empty `<think></think>` block in non-thinking prefill mode. Shipped examples are unaffected, so no run is invalidated. Substring match or basename normalisation would be the safer alignment; a small case table would pin it (there is currently no test for this function).

**5. `Nit:` `english_words.json` is now load-bearing but fails opaquely** — `sieval/datasets/ruler/_cwe.py:47-51`

With the cap gone, a missing file raises `ValueError: Sample larger than population` from inside `random.sample` at 128k (silently fine at ≤32k). Upstream opens it unconditionally and would at least name the path. One explicit `raise` naming the file and the download command closes it.

**6. `Nit:` PR body no longer matches what is shipped**

Still says "stacked on #9", "5 synthetic loaders", "bundled in-tree", "VT/CWE are `kshot` (`n_shot=1`)" — actual is one dataset + one `ruler_0shot_gen` (`n_shot=0`) with a BYO corpus. Please also attach the per-length table from the six thinking reports, add the item-1 caveat to the thinking `<3%` claim, and note the runs were a dev build (`meta.json`: `version 0.0.0`, `deterministic: false`). The CWE cells at 64k/128k were re-run after the cap removal (44.54→46.30, 13.98→13.46, headline 86.86→86.87) — worth one line so the table's provenance is unambiguous.

### Follow-ups (not blocking)

- **RFC on the dataset↔model boundary.** Upstream keeps its generators model-agnostic: `prepare.py:86-101,140` resolves the model into `--template` plus two integers. This port pushed `model_name` into the loader instead, which is where the three-way qwen3 check and the `131072` branch come from. RULER is the only one of 60 registered datasets taking model identity, so this sets a precedent worth deciding deliberately rather than inside a benchmark PR.
- Move `model_name` / `enable_thinking` / `think_budget` / `is_qwen3` / the tag constant out of `datasets/ruler/` once that is settled — item 4 disappears with it.
- `preprocess` sniffing `extra_body` for `continue_final_message`/`add_generation_prompt` → explicit knob.
- Whether to give the answer a real budget in thinking mode. Needs a re-run, and per item 1 it would push the result outside the <3% band — a product call, not a fix.
- `sieval/datasets/ruler/_vt.py:183-192` still expresses the template reserve the old way while `:236` uses upstream's form; converging lets `_count_prompt_tokens` and `apply_template` go away.
- Non-thinking 64k/128k CWE re-run post-cap-removal (estimated ±0.03 on the headline, from the thinking-side delta).

### Prior comments

Rounds 1–6: all resolved — spot-verified the `else: break` removal in all four retry loops, `_subtask_data_path`, `_read_hotpotqa` staged read + set comprehension, `transformers`/`nltk` in the extra, wonderwords `<3`, no internal branch names, no `leaderboard/` churn, pre-commit revert.

Round 7: 2a / 2b / 2d / 3 / 4 resolved. **Item 1 resolved** — the full 6-length thinking sweep exists and the earlier think-32k=6.89 collapse was indeed the serving config (32k now 88.15). **Item 2c responded, but the reasoning does not hold** → item 2 above.

ethan-scitix (approved):

PR #12, opened 2026-06-22
feat(mbpp): add dataset and few-shot base-model task

jack-scitix-ai:
## Type

## Summary

- Add MBPP benchmark: dataset loader + base-model few-shot generative task with pass@k execution scoring.
- `mbpp_kshot_base_gen` (base gen): few-shot prompt with `[BEGIN]`/`[DONE]` delimiters, model completes the function, candidate is executed against `test_list[:3]` via the SiEval code-eval API; pass@k accuracy metric.
- Few-shot examples (task_id 2/3/4), prompt template, `[DONE]` stop, and the `test_list[:3]` harness vendored byte-for-byte from lm-evaluation-harness `mbpp.yaml` @ `1dd9310`.
- Loader mirrors `hf:google-research-datasets/mbpp` config `full` (revision-pinned `@4bb6404`), keeping the native splits prompt/test/validation/train (10/500/90/374); evaluation runs on `test` (500). No id-range split rebuild — `full` ships the official splits.
- `n_shot=3` matches the lm-eval 3-shot reference; params follow the base-gen family (`n_shot` few-shot count, `k` pass@k, `n` sampling count). Reference: https://github.com/EleutherAI/lm-evaluation-harness/blob/1dd931087362abba74e0375c8c631295559f48b2/lm_eval/tasks/mbpp/mbpp.yaml

### Automated

### Manual

| --- | --- | --- | --- |
| Qwen2.5-72B-Base (3-shot) | 76.0 | 76.6 | +0.6 (+0.6%) |

Note: expected 76.0 is the Qwen3 report (Table 3) Qwen2.5-72B-Base MBPP 3-shot Pass@1; DeepSeek-V3 (Table 3) reports 72.6 under an unspecified protocol, so it isn't used as the target. Few-shot set, prompt template, `[DONE]` stop, and `test_list[:3]` harness are byte-exact vs lm-eval `mbpp.yaml` @ `1dd9310`. The earlier `url:`/jsonl run scored the same 76.6, and an offline diff confirms the HF `full`-config `test` split is byte-identical to the old task_id 11–510 slice — the migration didn't shift the evaluated set. Re-run at the upstream `timeout=3.0`: score 76.6, `fails=0`, `timeouts=0`.

### Required (all PRs)

### If: community/ Changes


ethan-scitix (changes requested):
Reviewed against lm-evaluation-harness MBPP @ `1dd9310`. Reproduction verified empirically (76.6, 0 fails) — the few-shot set, prompt template, `[DONE]` stop, `test_list[:3]` scoring, and splits all match the reference. No blocking correctness bugs. Requesting changes on one architectural point plus a few consistency/fidelity items.

### `sieval/datasets/mbpp.py` + `downloaders/url.py`

**Question (architecture):** lm-eval's `mbpp.yaml` loads `google-research-datasets/mbpp` config `full` from HF — parquet, no `trust_remote_code`, and it already provides the `prompt/test/validation/train` splits (10/500/90/374) that `mbpp.py:94` rebuilds by hand. Switching to `hf:google-research-datasets/mbpp@<sha>` aligns with the reference and drops `MBPP_JSONL_URL`, `_resolve_data_file`, the split rebuild, and the empty-split guard. It also makes the `url.py` `_has_compressed_content` change unnecessary — MBPP is its only caller (mmlu/gpqa are identity-served), so please drop it or split it into its own PR.

### `sieval/tasks/mbpp_kshot_base_gen.py`

**Question (`mbpp_kshot_base_gen.py:80-122`):** Collapse `num_shots` + `k`-alias + conflict-check into a single `k: int | None` per the theoremqa sibling (PR #3); keep `pass_k`. The alias re-introduces the `k` ambiguity it tried to remove.

**Nit (`mbpp_kshot_base_gen.py:172-178`):** Drop `dict[str, object]`, pass explicit kwargs like PR #6: `return await self.model.agenerate(pre, n=self._n, stop=list(self._stop))`.

**Nit (`mbpp_kshot_base_gen.py:86`):** Make `stop` non-optional (`tuple[str, ...] = STOP_SEQUENCES`, per PR #6) — `[DONE]` is prompt-required, and this removes the conditional above.

**Nit (`mbpp_kshot_base_gen.py:138`):** lm-eval doesn't strip the joined tests; `.strip()` drops a trailing space on 14/500 prompts (score-neutral, but an unannotated divergence in a strict-reproduction task). Drop it or annotate.

**Nit (`mbpp_kshot_base_gen.py:54-61`, reference_impl notes):** Add a half-sentence that DeepSeek-V3's 72.6 uses an unspecified protocol, so the +4.0 vs the Qwen-aligned 76.6 isn't read as an implementation error.

### `sieval/community/mbpp.py`

**Nit (`community/mbpp.py:5`):** Pin the URL to `1dd931087362abba74e0375c8c631295559f48b2` (the SHA the few-shot was copied from), matching the task's `reference_impl`.

jack-scitix-ai:
Thanks for the detailed review — addressed all points.

### `datasets/mbpp.py` + `downloaders/url.py` (architecture)

- Switched to `hf:google-research-datasets/mbpp@4bb6404fdc6cacfda99d4ac4205087b89d32030c` config `full`. The repo natively ships `prompt/test/validation/train` (10/500/90/374) with a schema identical to `MBPPDatasetSample`, so I dropped `MBPP_JSONL_URL`, `_resolve_data_file`, the task_id-range split rebuild, and the empty-split guard. The loader is now the minimal HF form (matches `human_eval`/`math_500`).
- With MBPP off the URL path, `_has_compressed_content` had no remaining caller (mmlu/gpqa are identity-served), so I reverted `url.py` to its pre-PR baseline (`git diff` against the base is empty) rather than splitting it out. Happy to reintroduce it as its own PR if you'd like the capability kept.

### `tasks/mbpp_kshot_base_gen.py`

- Collapsed `num_shots` + `k`-alias + conflict-check into a single `k: int` (kept `pass_k`), per the theoremqa sibling.
- `infer` now passes explicit kwargs: `agenerate(pre, n=self._n, stop=list(self._stop))`.
- `stop` is non-optional (`tuple[str, ...] = STOP_SEQUENCES`); removed the conditional.
- Dropped `.strip()` in `_format_tests` so the joined tests match lm-eval byte-for-byte; added a comment noting the retained trailing space.
- reference_impl notes now state DeepSeek-V3's 72.6 uses an unspecified protocol, so the gap to the Qwen-aligned number reads as a protocol difference, not an implementation error.

### `community/mbpp.py`

- Pinned the upstream `utils.py` URL to `1dd931087362abba74e0375c8c631295559f48b2`, matching the task's `reference_impl`.

### Tests / sync

- Removed `test_mbpp.py` — it only covered the now-deleted URL/split machinery (the other 12 HF loaders have no dataset unit test). Reverted the compressed-transport case in `test_url.py`. Task tests updated for the `k` rename.

ethan-scitix (commented):
Faithful port — few-shot data, prompt template, and matcher are byte-exact vs lm-eval-harness @1dd9310, and the reproduction checks out (Qwen2.5-72B-Base pass@1 = 76.6 vs published 76.0). Two small items before merge:

**Question — `timeout` default.** `sieval/tasks/mbpp_kshot_base_gen.py:86` defaults `timeout=5.0`, but lm-eval scores MBPP via HF `code_eval` whose default is **3.0s** (`pass_at_1` passes no timeout). It's a more-lenient, unannotated deviation — harmless here (`timeouts: 0`), but for a strict-repro task: either set it to `3.0`, or keep `5.0` with a one-line "intentional" comment. Please use the same value as `human_eval_0shot_base_gen` (#6, also `5.0`) so the family stays aligned.

**Dataset test coverage.** `e5080e9f` swapped the loader to the HF `full` config but deleted `tests/unit/datasets/test_mbpp.py` without replacing it, so `MBPPDataset.load` ships with no coverage. Worth re-adding a small test asserting the four splits/counts (prompt 10 / test 500 / validation 90 / train 374) — it also doubles as proof the new loader yields the same test split the 76.6 run used.

jack-scitix-ai:
Thanks — both addressed.

### `timeout` default

Kept `5.0` to stay aligned with the `human_eval` code-exec task family, and added a one-line comment at the param marking it as an intentional, more-lenient deviation from lm-eval's HF `code_eval` 3.0s default (observed `timeouts: 0`):

```python
# lm-eval scores MBPP via HF code_eval (default 3.0s). 5.0s is an
# intentional, more-lenient value shared with human_eval_0shot_gen to
# keep the code-exec task family aligned; observed timeouts stay at 0.
timeout: float = 5.0,
```

### Dataset test coverage

Re-added `tests/unit/datasets/test_mbpp.py` covering `MBPPDataset.load`. Unit tests here don't hit the network (repo convention — all loaders are mocked / `_hf_dict`-bypassed, no test downloads real HF data), so the test stubs `load_dataset` with the upstream split shapes **keyed by config**:

- `test_load_uses_full_config_and_preserves_official_splits` — asserts the loader yields the four official splits at published counts **prompt 10 / test 500 / validation 90 / train 374**. The stub raises on any config other than `full`/`sanitized`, so this assertion also proves `load` requests `full` (test=500), not `sanitized` (test=257) — i.e. the exact test split the 76.6 run used.
- `test_load_passes_explicit_config_override` — `config="sanitized"` flows through (test=257).
- `test_load_rejects_non_dataset_dict` — a single-split (`Dataset`) load is rejected by `ensure_dataset_dict`.

ethan-scitix (commented):
Reproduction confirmed against the author's runs: **76.6** on `test`=500 (0 timeouts, 0 fails), and the earlier `url:`/jsonl run scores the same — good evidence the HF `full`-config migration didn't shift the evaluated set. Few-shot examples, prompt template, `[DONE]` stop, and the `test_list[:3]` harness are byte-exact against lm-eval `mbpp.yaml` @ `1dd9310`. Two changes before merge, plus two nits.

---

**`Bug:` param naming diverges from the sibling family** — `sieval/tasks/mbpp_kshot_base_gen.py:82-83`

Here `k` is the few-shot count and `pass_k` is pass@k, but `livecodebench_code_generation_kshot_base_gen.py:130-131` and `human_eval_0shot_base_gen.py:96` use `n_shot`=few-shot and `k`=pass@k. Same YAML key, opposite meaning across the family — a real run config here sets `k: 3` to mean 3-shot. It also disagrees with this task's own `n_shot=3` metadata. Please rename `k`→`n_shot` and `pass_k`→`k` (params, validators/messages, the `pass@{...}` key, and tests).

**`Bug:` timeout doesn't match upstream, and the comment's justification is wrong** — `sieval/tasks/mbpp_kshot_base_gen.py:86-89`

lm-eval scores MBPP via HF `code_eval`, whose default `timeout` is `3.0` (lm-eval doesn't override it). The comment claims 5.0 is "shared with `human_eval_0shot_gen` to keep the family aligned," but both HumanEval tasks use 3.0 and LiveCodeBench uses 6.0 — nothing uses 5.0. Set `timeout=3.0` (matches upstream *and* the HumanEval siblings) and drop the alignment claim. Observed timeouts are 0 at 5.0s, so a quick re-run should confirm the score holds at 3.0.

**`Nit:` `report()` empty-input schema is inconsistent** — `sieval/tasks/mbpp_kshot_base_gen.py:222-223`

The `total == 0` branch returns `{score, fails}`, but the populated path returns `{score, fails, timeouts, pass@1}` (confirmed in `report.json`). Emit the same keys (zeros) in the empty branch. `human_eval_0shot_base_gen.py` has the same split — worth fixing both so the family doesn't drift.

**`Nit:` PR body is stale** — it describes a `url:`/`mbpp.jsonl` source, a task_id-range split rebuild, and a "bundled URL downloader fix," none of which are in the diff (switched to `hf:…@…` config `full` in `8b36b0da`). Please re-sync the description.

ethan-scitix (approved):

jack-scitix-ai:
Thanks — all four addressed. Re-ran at `timeout=3.0` to confirm the score holds.

**Bug 1 — param naming**  Renamed to match the base-gen family: few-shot count `k`→`n_shot`, pass@k `pass_k`→`k` (params, validators/messages, internal `self._n_shot`/`self._k`, the `pass@{...}` key, YAML run configs, and tests). Now consistent with `human_eval_0shot_base_gen` / `livecodebench_code_generation_kshot_base_gen` and with this task's own `n_shot=3` metadata.

**Bug 2 — timeout**  Set `timeout=3.0` to match lm-eval's HF `code_eval` default (and the HumanEval siblings), and dropped the incorrect "aligned with `human_eval_0shot_gen`" comment — you're right, nothing in the family uses 5.0. Comment now just states it matches upstream. Re-ran the full `test`=500: **score 76.6, fails 0, timeouts 0** — holds at 3.0.

**Nit 3 — empty `report()` schema**  Empty branch now emits the same keys as the populated path (`{score, fails, timeouts, pass@1}`, zeros). Applied the same fix to `human_eval_0shot_base_gen` so the family doesn't drift, and updated the empty-report unit test to assert the full schema.

**Nit 4 — stale PR body**  Re-synced the description to the `hf:…@…` config `full` source (native splits, no id-range rebuild, no URL-downloader change).

PR #13, opened 2026-06-23
feat(ifbench): add dataset and few-shot base-model task

jack-scitix-ai:
## Type

## Summary

- Add the IFBench dataset wrapper backed by `hf:allenai/IFBench_test`, pinned to revision `2e8a48de45ff3bf41242f927254ca81b59ca3ae2`.
- Add `ifbench_0shot_gen`, a 0-shot generative chat task for precise, verifiable instruction following (58 OOD constraints over held-out WildChat prompts).
- Vendor AllenAI's `evaluation_lib` + instruction registry/checkers under `sieval/community/ifbench/`; the task reports **prompt-level loose accuracy** as the headline `score`, the metric the IFBench paper reports.
- Register the IFBench dataset/task exports, stubs, and meta index entries for discovery; add the `[ifbench]` optional-dependency group (`emoji`, `nltk`, `syllapy`, `setuptools<81`).
- Reference paper: https://arxiv.org/pdf/2507.02833
- Reference repo: https://github.com/allenai/IFBench/blob/1091c4c3de6c1f6ed12c012ed68f11ea450b0117/evaluation_lib.py
- Comparison target: AllenAI leaderboard, **Qwen3-32B = 37.3** (prompt-level loose accuracy). The official hyperparameters are confirmed by the IFBench authors in [allenai/IFBench#5](https://github.com/allenai/IFBench/issues/5): `max_gen_toks: 32768`, `stop_sequences: ["</answer>"]`, `temperature: 0`, `process_output: "r1_style"`, thinking enabled.
- Known alignment caveat (thinking models): the headline score is driven almost entirely by how many responses come back empty, not by the scorer. Under greedy decoding (`temperature 0`) with Qwen3 thinking enabled, the model can enter deterministic repetition loops in the reasoning trace, exhaust the token budget, and emit an empty answer — which fails every instruction. Reasoning-chain stripping is handled by the chat backend separating `reasoning_content` from `content`, so `texts[0]` is already the answer without the trace.

### Automated

### Manual

- [x] Dataset loading tested (`sieval dataset download ifbench` succeeds).
- [x] Ran `ifbench_0shot_gen` on the full IFBench test split (300 prompts) with Qwen/Qwen3-32B via a Scitix OpenAI-compatible endpoint.

**1. Comparison to the leaderboard**

| --- | ---: | ---: | ---: |
| Qwen3-32B | 37.3 | 37.3 | 0.0 |

This 37.3 is **not an exact reproduction** — it happened to come out at `temperature 0.6 / top_p 0.95 / top_k 20 / max_tokens 38912` (thinking enabled), and since the temperature is not 0 the result is inherently random. Two temperature-0.6 runs landed at **37.33** and **37.00** (the second with `seed 42`), so the precise match to 37.3 is one fortunate sample, not a deterministic outcome. Validated report (from output):

```json
{"score": 37.33, "loose_prompt_level_accuracy": 37.33, "strict_prompt_level_accuracy": 30.00, "fails": 0}
```

**2. Why we did not run at the official `temperature 0`**

The IFBench authors posted their hyperparameters in [allenai/IFBench#5](https://github.com/allenai/IFBench/issues/5): `temperature 0`, `max_gen_toks 32768`, `stop ["</answer>"]`, `process_output "r1_style"`, thinking enabled. We could **not** reproduce the score under these settings: with Qwen3 thinking enabled and `temperature 0`, the model reliably enters a degenerate reasoning-repetition loop that exhausts the token budget and returns an empty answer, which fails every instruction. This is experimentally confirmed (from output):

| temperature | max_tokens | loose (score) | length-truncated | empty answers |
| --- | ---: | ---: | ---: | ---: |
| 0.0 (greedy) | 32768 | 34.0 | 22 | 18 |
| 0.0 (greedy) | 38912 | 36.3 | 20 | 16 |

Raising `max_tokens` does not help — the greedy loop is deterministic and never terminates, and `stop: ["</answer>"]` never fires because generation never reaches the answer boundary. So the author's posted settings appear problematic in practice on this endpoint, and we currently cannot explain how their implementation avoids the loop. We therefore fall back to non-greedy sampling, which breaks the loops (empty answers drop from 18 → 1).

**3. Stability check — fixed config, 5 runs averaged**

Because the 37.3 above is a noisy single sample, we separately verified the scoring/implementation is stable by fixing a clean config (thinking **off**, `temperature 0`, `max_tokens 8192` — no repetition bug when thinking is disabled) and running the full split 5 times (from output):

| run | loose (score) | strict |
| --- | ---: | ---: |
| 1 | 27.33 | 23.00 |
| 2 | 28.00 | 24.00 |
| 3 | 28.67 | 24.67 |
| 4 | 27.33 | 24.33 |
| 5 | 27.67 | 24.33 |

Mean **27.80**, stdev **0.56**, range **1.33**, with **0 empty answers** in all 5 runs. The small spread indicates the eval pipeline produces stable, reproducible scores, so the implementation itself is sound — the difficulty in matching 37.3 is a model/endpoint decoding issue, not a scorer problem.

### Required (all PRs)

- [x] Dataset loading tested (`sieval dataset download ifbench` succeeds)

### If: community/ Changes

### If: New Dependency


ethan-scitix (changes requested):
**Review — IFBench (0-shot generative).**

Verified empirically: the vendored scorer is byte-identical to upstream `1091c4c3` (only the import + NLTK-cache adaptations), `score`=loose prompt-level matches AllenAI's official headline metric, reasoning separation worked 300/300 on the author's runs (0 `<think>` leakage), float64 kwargs are checker-safe, deps/lock/registration are consistent, and the 4 new tests pass. Nice port. Two items need resolving before merge (1–2), plus two small follow-ups (3–4).

**1. `Question:` the reproduction story isn't aligned at the official decoding.** `sieval/tasks/ifbench_0shot_gen.py:11-13` documents "official reproduction: temperature=0, 32768". At exactly those settings the run scores **34.0**, not the 37.3 target. The only run that hits 37.3 uses temp=0.6 / top_p=0.95 / top_k=20 — which is **Qwen3's recommended sampling, not IFBench's leaderboard口径** (per allenai/IFBench#5 the 37.3 entry is temperature=0, no penalties). The gap is greedy degeneration (≈18/300 runaway generations truncate into empty answers), not a port defect. Please either report 34.0 as the matched-decoding number with that caveat, or drop the 37.3 alignment claim — the temp=0.6 run shouldn't be presented as reproduction.

**2. `Question:` drop the file/dir/JSONL branching in `sieval/datasets/ifbench.py:42-64`.** `source` is pinned to `hf:allenai/IFBench_test@<rev>`, so production always loads the staged parquet through the `load_dataset` fallback. The `IFBench_test.jsonl` branch loads an unpinned alternate artifact that bypasses the pinned-source contract and is never produced by the download pipeline — yet it's the only path the two unit tests exercise (the real parquet path is untested). Suggest matching the `ifeval.py` sibling (`load_dataset` + `apply_eval_split`) and testing that path.

**3. `Nit:` `Dockerfile` is missing two NLTK corpora.** IFBench's checkers need `stopwords` + `averaged_perceptron_tagger_eng` (pulled at import via `instructions_util.download_nltk_resources()`). `Dockerfile:11-13` pre-seeds only `punkt punkt_tab wordnet omw-1.4`, so the prepared image still triggers a runtime network download for those two. Add them to the `nltk.downloader` line so the image stays self-contained (the IFEval sibling relies on this pre-seed rather than import-time download).

**4. `Nit:` document the `setuptools<81` ceiling.** `pyproject.toml:33` — the pin exists because `syllapy` imports `pkg_resources` (removed in setuptools 81). A one-line comment will stop a future bump from breaking `import syllapy`.

jack-scitix-ai:
Thanks — all four addressed. The two-run data on point 1 confirms your read.

**1. Reproduction story → reporting 34.0 at official decoding, with a caveat.**

At the official settings (temperature 0, no penalties, `max_gen_toks` 32768) the run scores **34.0**, so that's the matched-decoding number I'll report; the 37.3 "reproduction" claim is dropped. The gap is greedy degeneration on this endpoint, not a port defect — evidence:

| decoding | loose (score) | empty answers | loose over non-empty |
|---|---:|---:|---:|
| temp 0, no penalty (official) | 34.0 | 18/300 | 36.2% |
| temp 0 + presence_penalty 1.0 | 35.0 / 36.0 | 6 / 3 | 35.7% / 36.4% |
| temp 0.6 / top_p 0.95 / top_k 20 | 37.3 / 37.0 | 1 / 3 | 37.5% / 37.4% |

Accuracy over non-empty responses is stable (~36–37.5%) across every config; the headline score tracks only the empty-answer count (18 → 1), which comes from runaway reasoning-repetition truncations under greedy decoding on scitix endpoint. Two points worth noting:

- I tried `presence_penalty` (Qwen3's documented remedy for endless repetition) to stay at temperature 0: it reduces empties (18 → 3–6) and lifts the score to 35–36, but does not fully clear them and still falls short of 37.3.
- Temperature 0 is **not** deterministic here either — two identical `presence_penalty=1.0` runs gave 36.0 vs 35.0 — so no config yields a bitwise-reproducible number. That's why I'm reporting the official-decoding 34.0 with this caveat rather than presenting any single lucky run as reproduction. The docstring is reworded to say "official leaderboard decoding" (no score claim).

**2. Loader.** Collapsed `load()` to `ensure_dataset_dict(load_dataset(name_or_path, **kwargs))` + `apply_eval_split(dataset, "train")` — the file/dir/JSONL branching and `_load_jsonl` are gone, so production only ever reads the pinned staged parquet. Both unit tests rewritten to exercise this path (train→test mapping via a stubbed `load_dataset`; asserting no `split=` is forced and kwargs pass through).

**3. Dockerfile.** Added `stopwords averaged_perceptron_tagger_eng` to the `nltk.downloader` line so the image is self-contained.

**4. setuptools pin.** Added a one-line comment: `syllapy` imports `pkg_resources` (removed in setuptools 81), so the `<81` ceiling must stay.

ethan-scitix (commented):
**Verdict: functionally correct — approve after one docstring fix.** Vendored evaluators are byte-faithful to `allenai/IFBench@1091c4c` (only the documented import/NLTK-path changes), the HF source matches upstream's jsonl, deps are complete, and the task reproduces the benchmark: on the Qwen3-32B runs the best config hits **37.33 loose prompt-level ≈ leaderboard 37.3** (300/300, 0 fails). `status="stable"` is warranted.

### [Must-fix] The "official decoding" cited in the docstring doesn't reproduce the number

`sieval/tasks/ifbench_0shot_gen.py:11` cites temp=0 / 32768 as official, but empirically:

| config | score | truncated / empty |
|---|---|---|
| temp=0, 32768 | 34.00 | 22 / 18 |
| temp=0, 38912 | 36.33 | 20 / 16 |
| **temp=0.6 / top_p 0.95 / top_k 20, 38912** | **37.33** | 2 / 1 |

Greedy decoding makes Qwen3's reasoning loop and overrun the budget (→ empty answers scored as fail); the token budget is secondary, sampling is the real lever — and temp=0 is exactly what Qwen3 warns against for thinking mode. Please (a) fix the docstring so temp=0/32768 isn't presented as the reproduction recipe, and (b) point `ReferenceImpl.notes` at the config that actually reproduced it, noting it's a stochastic band (the same config gave 35.33 / 37.00 / 37.33 across 3 runs — 37.3 is the top edge).

### [Question] Is there a temp=0 recipe that reaches the target?

`presence_penalty` at temp=0 cuts the repetition/truncation (22 → 4–7, score → ~36) but stayed at 35–36 in the runs on record — all capped at 32768. pp=1.0 is likely on the low side (1.1–1.2 is more typical), and temp=0 + pp + 38912 was never tried. If a deterministic recipe is preferred, one run would confirm it; otherwise just document that Qwen-recommended sampling is required.

### [Nits]

- `notes` attributes "37.3" to `allenai/IFBench#5`, but #5 only gives the (non-reproducing) hyperparameters — 37.3 is from the README leaderboard. Split the citation.
- `README.md:10` says "16 registered benchmark datasets"; the registry has 19 (missing HMMT Feb 2025, IMO-AnswerBench, TheoremQA — pre-existing, but fixable here since the line's already edited).

jack-scitix-ai:
Thanks — all four addressed.

**[Must-fix a] Docstring no longer presents temp=0/32768 as the reproduction recipe.**
Reworded to label it "official leaderboard hyperparameters (allenai/IFBench#5)", then state that greedy decoding degenerates on Qwen3 thinking mode (reasoning loops → empty answers scored as fail) and that reproduction requires Qwen3's recommended sampling (temperature=0.6 / top_p=0.95 / top_k=20). Decoding values only, no score claims.

**[Must-fix b] `ReferenceImpl.notes` points at the reproducing config + flags the band.**
Now names the config that reproduced (temp=0.6 / top_p 0.95 / top_k 20 / 38912) and states the score is a stochastic band with 37.3 at the top edge, not a deterministic value.

**[Nit] Citation split.** 37.3 is now attributed to the AllenAI README leaderboard; `allenai/IFBench#5` is cited only as the (non-reproducing) hyperparameter source.

**[Nit] README count.** Corrected against the live registry — now 21 (was 16), adding HMMT Feb 2025, IMO-AnswerBench, TheoremQA (pre-existing gaps) plus OpenBookQA/MBPP from the latest main, and IFBench.

**[Question] Is there a temp=0 recipe that reaches the target?**
Ran the untested combination — temperature=0 + presence_penalty>1 + no truncation (38912) — at pp=1.1 and pp=1.2, full split, thinking on, 0 fails:

| decoding (temp=0, thinking) | loose (score) | strict | truncated | empty | acc. over non-empty |
|---|---:|---:|---:|---:|---:|
| no penalty, 32768 | 34.00 | 26.33 | 22 | 18 | 36.2% |
| presence_penalty 1.0, 32768 | 36.00 | 28.67 | 4 | 3 | 36.4% |
| presence_penalty 1.1, 38912 | **33.33** | 26.33 | 3 | 3 | **33.7%** |
| presence_penalty 1.2, 38912 | **37.33** | 30.00 | 5 | 3 | 37.7% |
| temp=0.6 / top_p 0.95 / top_k 20, 38912 (ref) | 37.33 | 30.00 | 2 | 1 | 37.5% |

Numerically pp=1.2 hits 37.33, but it is **not a reproducible or defensible recipe**:

1. **Non-monotonic — it drops *below* the no-penalty baseline.** pp 1.0 → 1.1 → 1.2 gives 36.0 → 33.3 → 37.3; pp=1.1 (33.3) is worse than no penalty (34.0). No "higher pp → higher score" trend, just scatter.
2. **The swing is in answer quality, not truncation.** All three pp configs have the same empty count (3) — truncation is already solved — yet accuracy over non-empty responses moves 36.4% → 33.7% → 37.7%. presence_penalty is perturbing the answers themselves.
3. **presence_penalty is mechanistically adversarial to IFBench.** Many constraints are exact keyword-repetition counts ("include keyword X exactly N times"); penalizing already-emitted tokens works directly against them, consistent with the pp=1.1 dip.
4. **temp=0 isn't deterministic on this endpoint anyway** (the same pp=1.0 config gave 35.0/36.0 on two earlier runs), so pp=1.2's 37.33 is one draw of a band, and picking pp=1.2 over pp=1.1 post hoc would be fitting to the outcome.

Conclusion: no principled temp=0 recipe — low pp truncates, and tuning pp to "land" on 37.3 is parameter-fishing against a noisy, constraint-adversarial knob. Keeping Qwen3's recommended sampling as the documented reproduction path, score noted as a stochastic band.

ethan-scitix (approved):

PR #14, opened 2026-06-23
feat(livecodebench): add dataset and few-shot base-model task

jack-scitix-ai:
## Type

## Summary

- Add `livecodebench_code_generation_kshot_base_gen` — a **base-model** (completions / `GenModel`) few-shot counterpart to the existing instruct task `livecodebench_code_generation_0shot_gen`.
- Mirrors LiveCodeBench's official base-model prompt builder (`get_base_model_question_template_answer`), decomposed into a cacheable few-shot prefix + per-sample target block so the fixed prefix is built once in `setup()`, not per sample. For `n_shot=1` the rendered prompt is byte-identical to upstream.
- Default `n_shot=3` + `stop=("###",)` target **DeepSeek-V3 Table 3 "LiveCodeBench-Base (Pass@1), 3-shot"**. Upstream ships only 2 few-shot examples per pool, so a 3rd example per pool is sieval-authored (marked + documented).
- Decoding params (`temperature`/`top_p`/`max_tokens`) are **not** set by the task — only `n` (pass@k) and the prompt-coupled `stop` are forwarded; decoding comes from model config / `infer_args`.
- Pin the LiveCodeBench HF dataset to a fixed revision (`@0fe84c39…`), for reproducibility.
- Refs: [LiveCodeBench](https://github.com/LiveCodeBench/LiveCodeBench/blob/28fef95ea8c9f7a547c8329f2cd3d32b92c1fa24/lcb_runner/prompts/code_generation.py) · [DeepSeek-V3 report](https://arxiv.org/abs/2412.19437) (Table 3).

### Automated

### Manual

- **Model:** Qwen2.5-72B (base), served via sglang, greedy (`temperature=0`, `top_p=1`, `max_tokens=2000`), `n_shot=3`.
- **Window:** 2024-08-01 → 2024-11-01 (`0801-1101`), 83 problems.
- **Reproducibility:** identical score across 3 independent runs (`fails=0`, `timeouts=15`).

|---|---|---|---|
| Qwen2.5-72B-Base | 12.9( (DeepSeek-V3 Report)) | 13.25 | +0.35 (+0.35%) |

### Required (all PRs)

### If: community/ Changes


ethan-scitix (commented):
**Verdict: approve — faithful reproduction, no blocking issues.** Two non-blocking nits below.

Verified empirically:
- Prompt builder byte-identical to upstream `get_base_model_question_template_answer` at `n_shot=1` (both pools); `func.json`/`stdin.json` examples [0]/[1] are verbatim upstream, [2] is `_source`-tagged sieval-authored.
- `extract_code(lmstyle="GenericBase")`, `stop=("###",)`, `max_tokens=2000` all match the upstream `lcb_runner` defaults; `GenericBase` ↔ base-model template coupling confirmed in upstream `format_prompt_generation`.
- Reproduction: 3 deterministic Qwen2.5-72B-Base runs all land Pass@1 = 13.25 (22/166), 0 fails, 0 anomalies — vs DeepSeek-V3 Table 3 anchor 12.9 (+0.35 abs, slightly high, consistent with the documented sieval-authored 3rd few-shot). The 15 timeouts are genuine service-side subprocess timeouts (46–62s, CPU pegged ~100%), correctly scored as failures; no sample hit the HTTP-timeout/exception path.
- Upstream URL + pinned HF revision resolve; `check_tasks`/`check_datasets`/`check_dep_coverage` and 12/12 unit tests pass.

The deviation disclosure (n_shot generalization, the 3rd authored example, external eval sandbox) across the docstring, `reference_impl.notes`, `index.json`, and the dedicated pinning test is exemplary.

---

`Nit:` `sieval/community/livecodebench/prompts/code_generation.py:5` — use `importlib.resources` instead of `Path(__file__).parent`. The latter breaks under non-filesystem loaders (zipimport) and diverges from the repo pattern (`sieval/meta/loader.py:25` uses `files(__package__).joinpath(...)`). Since the path had to be adapted from upstream's cwd-relative `open("lcb_runner/...")` anyway, land the robust form:

```python
from importlib.resources import files
_FEWSHOT = files(__package__).joinpath("few_shot_examples", "generation")
func = json.loads((_FEWSHOT / "func.json").read_text(encoding="utf-8"))
stdin = json.loads((_FEWSHOT / "stdin.json").read_text(encoding="utf-8"))
```

`Nit:` `sieval/tasks/livecodebench_code_generation_kshot_base_gen.py:162-165` — dead cache-miss branch in `preprocess`; rely on `setup()`. `has_starter` is always a `bool`, and `setup()` (`:154-156`) pre-fills both keys before any sample runs (`sieval/core/runners/runner.py:337`), so this `if` never fires in the pipeline (only when `preprocess` is called without `setup()`, i.e. the standalone test). Drop it and index directly — also removes the in-`preprocess` mutation of shared state:

```python
@override
async def preprocess(self, raw, ctx):
    has_starter = bool(raw["starter_code"])
    return self._fewshot_prefix[has_starter] + get_base_model_target_block(
        raw["question_content"], raw["starter_code"]
    )
```

Tests then need a `setup()` call first (matches the framework contract). Or drop `setup()` and go lazy-only like the `gsm8k_kshot_base_gen` sibling — either way, pick one mechanism.

jack-scitix-ai:
Thanks, both addressed in the latest commit.
1. **`importlib.resources`**: switched the few-shot loader to `files(__package__).joinpath(...)` + `read_text`, matching `sieval/meta/loader.py`. Robust under zipimport now.
2. **Dead cache-miss branch**: dropped it; `preprocess` indexes `self._fewshot_prefix` directly (kept the `setup()` mechanism), which also removes the in-`preprocess` shared-state mutation. Updated the standalone test to call `setup()` first per the framework contract.

ethan-scitix (commented):
One consistency item to fold in on rebase (branch is behind main; #6 landed after the branch point):

`Nit:` Align the code-eval call with the post-#6 LCB 0-shot sibling. #6 made the per-execution budget client-owned by passing it in the request body; this task still omits it, so the base task and the 0-shot task would run the **same benchmark on different execution budgets** (this one falls back to the service's unspecified default). In `feedback()`, add to the POST body to match the sibling:

```python
"test": {"inputs": inputs, "outputs": outputs, "fn_name": fn_name},
# All N cases share one sequential budget, so scale by N.
# Approximates official per-case 6s within a single run.
"timeout": self._timeout + len(inputs) * 2.0,
```

and reword docstring deviation #3 accordingly (timeout is now client-owned, not service-owned).

`Nit:` Drop the dead `self._max_concurrency` attribute — it's stored but never read (the httpx client uses the local `max_concurrency` param directly); #6 already removed it from the 0-shot task.

Everything else (decoding params kept model-side, `stop`/`n` forwarding, pass@k semantics, `k`-vs-`n_shot` naming) is consistent with the current base-model family on `main`.

jack-scitix-ai:
Thanks — rebased onto latest `main` and folded in both, now consistent with the post-#6 family.

1. **Client-owned execution budget**: added `"timeout": self._timeout + len(inputs) * 2.0` to the code-eval POST body, byte-identical to the 0-shot sibling, so both tasks run the same benchmark on the same budget instead of falling back to the service default. Reworded docstring deviation #3 — the budget is now client-owned (sent in the request body, scaled by case count to approximate the official per-case 6s), no longer service-owned.
2. **Dead attribute**: dropped `self._max_concurrency` (the httpx client uses the local `max_concurrency` param directly), matching #6's cleanup of the 0-shot task.

ethan-scitix (changes requested):
Verified the port is faithful — prompt builder is byte-identical to upstream at `n_shot=1`, the two vendored few-shot examples match upstream verbatim, and the run reproduces (pass@1 **13.25** vs DeepSeek-V3's **12.9** Qwen2.5-72B-Base anchor over `0801-1101`). Two changes before merge.

### `Bug:` LCB dataset default `version_tag="release_v1"` makes the repro window unrunnable — `sieval/datasets/livecodebench_code_generation.py:67`

The task's window is `2024-08-01..2024-11-01`, but the default `release_v1` predates it, so the date filter yields **zero** problems → silent `score: 0.0` (there's no empty-dataset guard, so `report()` falls into its `total == 0` branch). Bump the default to the release the pinned revision and `leaderboards/sft_fast_202511.yaml` already use:

```python
version_tag: str = "release_v6",  # release_v1 predates the 2024-08+ windows
```

Shared infra — the 0-shot sibling benefits too.

### `Nit:` drop `dict[str, object]` in `infer` — `sieval/tasks/livecodebench_code_generation_kshot_base_gen.py:176`

Use explicit kwargs (the conditional stays — `stop=None` would clobber the model's configured stop via the `{**self._kwargs, **kwargs}` merge):

```python
if self._stop:
    return await self.model.agenerate(pre, n=self._n, stop=list(self._stop))
return await self.model.agenerate(pre, n=self._n)
```

Same pattern in `sieval/tasks/gsm8k_kshot_base_gen.py:169` — worth fixing both.

jack-scitix-ai:
Thanks — both are addressed:

1. **`Bug:` default `version_tag`**: bumped the shared loader default `release_v1 → release_v6` so the 2024-08+ windows aren't stranded to an empty dataset (matches the pinned HF revision and `leaderboards/sft_fast_202511.yaml`; the 0-shot sibling benefits too). Also added the empty-after-filter guard from `datasets.md` — an empty window now raises loudly instead of silently scoring 0.0.
2. **`Nit:` `dict[str, object]` in `infer`**: replaced with explicit `agenerate` kwargs in both the LCB base task and `gsm8k_kshot_base_gen.py`; kept the conditional so `stop` stays out of the call when unset (won't clobber the model's configured stop via the `{**self._kwargs, **kwargs}` merge).

ethan-scitix (approved):

PR #15, opened 2026-06-24
feat(c-eval): add dataset and few-shot base-model task

jack-scitix-ai:
## Type

## Summary

- Add **C-Eval** (Chinese 52-subject exam MCQ): `CEvalDataset` + few-shot `CEvalFewShotCLPTask` (`c_eval_kshot_clp`), extending delivery-quality verification to Chinese knowledge.
- Scoring: next-token log-prob over A/B/C/D → argmax (≡ `softmax(logits[A,B,C,D])`), via deterministic `alogprobs(echo=True)`. Classified as **`EvalMode.CLP`** (next-token option-token log-prob) per #23. A sieval-native implementation **aligned with the method** of hkust-nlp/ceval `evaluator_series`, **not a byte-for-byte port** of the official repo.
- Eval on the released **test** split; `score` = **macro-average over the 52 subjects** (C-Eval paper Table 3), distinct from lm-eval `ceval-valid`'s micro.
- `CEvalDataset` mirrors the source as-is (52 per-subject configs, injects the English `subject` key, HF revision pinned); split selectable via `apply_eval_split`.
- Refs: github.com/hkust-nlp/ceval · arXiv:2305.08322.

### Automated

### Manual

- [x] Ran `c_eval_kshot_clp` on the full C-Eval test split with Qwen2.5-72B Base
- [x] Run completed 12342/12342 samples with 0 failures
- Actual score: 90.19 macro accuracy; reference score: 89.20; diff: +0.99

| --- | --- | --- | --- |
| Qwen2.5-72B Base | 89.20 | 90.19 | +0.99 (+0.99%) |

Reference: DeepSeek-V3 report (89.2). Diff within the <3% target.

### Required (all PRs)

### DIFF
--- a/sieval/datasets/__init__.pyi
+++ b/sieval/datasets/__init__.pyi
@@ -13,6 +13,10 @@ from .aime_2026 import (
     AIME2026Dataset,
     AIME2026DatasetSample,
 )
+from .c_eval import (
+    CEvalDataset,
+    CEvalDatasetSample,
+)
 from .cmmlu import (
     CMMLUDataset,
     CMMLUDatasetSample,
@@ -105,6 +109,8 @@ __all__ = [
     "AIME2025DatasetSample",
     "AIME2026Dataset",
     "AIME2026DatasetSample",
+    "CEvalDataset",
+    "CEvalDatasetSample",
     "CMMLUDataset",
     "CMMLUDatasetSample",
     "DROPDataset",
--- a/sieval/datasets/c_eval.py
+++ b/sieval/datasets/c_eval.py
@@ -1,0 +1,152 @@
+"""C-Eval dataset loader.
+
+C-Eval is a Chinese multi-domain exam benchmark with 52 subjects, organised on
+the Hub as 52 per-subject configs (there is no combined ``all`` config), each
+with ``dev`` (few-shot exemplars with explanations), ``val``, and ``test``
+splits. C-Eval has since released the ``test`` answers, so all three splits
+carry labels. This loader mirrors the source as-is — it loads every subject (or
+a caller-supplied subset), tags each row with its ``subject`` (the English
+config key, which is absent from the raw rows), and concatenates them.
+
+The evaluation split is selected via :func:`apply_eval_split`; with the default
+(``eval_split=None``) the released ``test`` split is the eval target. Pass
+``eval_split="val"`` to evaluate on the validation split instead.
+
+AI-Generated Code - Claude Opus 4.8 (Anthropic)
+"""
+
+from typing import Any, TypedDict, override
+
+from datasets import Dataset as HFDataset
+from datasets import DatasetDict as HFDatasetDict
+from datasets import concatenate_datasets, load_dataset
+
+from sieval.core.datasets import Category, Dataset, Level1Category, sieval_dataset
+from sieval.core.utils.hf import apply_eval_split, ensure_dataset_dict
+
+# Pin the Hub revision for reproducibility (current `main` at integration time).
+CEVAL_REVISION = "617524a00b307ff6f9933702f724131fe12ca7ce"
+
+
+class CEvalDatasetSample(TypedDict):
+    question: str
+    A: str
+    B: str
+    C: str
+    D: str
+    answer: str
+    subject: str
+
+
+@sieval_dataset(
+    name="c_eval",
+    display_name="C-Eval",
+    description="C-Eval — Chinese multi-domain exam benchmark with 52 subjects.",
+    source=f"hf:ceval/ceval-exam@{CEVAL_REVISION}",
+    categories=(Category(Level1Category.KNOWLEDGE, "Multi-domain"),),
+    tags=("chinese", "multiple-choice"),
+    license="cc-by-nc-sa-4.0",
+)
+class CEvalDataset(Dataset[CEvalDatasetSample]):
+    """C-Eval dataset for Chinese model evaluation.
+
+    Row fields: ``question``, ``A``/``B``/``C``/``D`` (options), ``answer`` (the
+    correct option letter), and ``subject`` (the English config key, injected by
+    this loader since it is not present in the raw rows).
+    """
+
+    # All C-Eval subjects (52 total), used as the Hub config names.
+    SUBJECTS = [
+        "accountant",
+        "advanced_mathematics",
+        "art_studies",
+        "basic_medicine",
+        "business_administration",
+        "chinese_language_and_literature",
+        "civil_servant",
+        "clinical_medicine",
+        "college_chemistry",
+        "college_economics",
+        "college_physics",
+        "college_programming",
+        "computer_architecture",
+        "computer_network",
+        "discrete_mathematics",
+        "education_science",
+        "electrical_engineer",
+        "environmental_impact_assessment_engineer",
+        "fire_engineer",
+        "high_school_biology",
+        "high_school_chemistry",
+        "high_school_chinese",
+        "high_school_geography",
+        "high_school_history",
+        "high_school_mathematics",
+        "high_school_physics",
+        "high_school_politics",
+        "ideological_and_moral_cultivation",
+        "law",
+        "legal_professional",
+        "logic",
+        "mao_zedong_thought",
+        "marxism",
+        "metrology_engineer",
+        "middle_school_biology",
+        "middle_school_chemistry",
+        "middle_school_geography",
+        "middle_school_history",
+        "middle_school_mathematics",
+        "middle_school_physics",
+        "middle_school_politics",
+        "modern_chinese_history",
+        "operating_system",
+        "physician",
+        "plant_protection",
+        "probability_and_statistics",
+        "professional_tour_guide",
+        "sports_science",
+        "tax_accountant",
+        "teacher_qualification",
+        "urban_and_rural_planner",
+        "veterinary_medicine",
+    ]
+
+    @override
+    def load(
+        self,
+        name_or_path: str,
+        subjects: list[str] | None = None,
+        eval_split: str | None = None,
+        **kwargs: Any,
+    ) -> HFDatasetDict:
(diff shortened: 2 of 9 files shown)


ethan-scitix (commented):
**`Nit:`** Drop the `raw is None` guard in `feedback` (`c_eval_kshot_ppl.py:215-217`). The framework guarantees `ctx.raw_sample` at this stage (sibling `gsm8k_kshot_base_gen.py:195` accesses it unconditionally); as written a `None` finalizes with `subject=""`, fabricating a phantom 0% subject that skews the macro-average.

**`Nit:`** Unnecessary `Any` widening on stage signatures (`c_eval_kshot_ppl.py:95,173,179,228`). `Model[Any]` / `TaskContext[Any, Any, Any, Any, Any]` / `Sequence[Any]` discard the precise types the base `Task[...]` generics provide. Every other task in `sieval/tasks/` leaves these unannotated and inherits concrete types — and `feedback:213` here already uses the concrete first param, so the file is internally inconsistent. Drop the annotations to match siblings.

jack-scitix-ai:
Thanks, both are fixed in the latest commit:
- Dropped the raw is None guard in feedback — ctx.raw_sample is guaranteed here (matching gsm8k_kshot_base_gen), and the phantom subject="" it could fabricate would indeed skew the macro-average.
- Removed the Any widening on __init__/preprocess/infer/postprocess/report; the stages now inherit the concrete Task[...] generics like the sibling tasks, which also resolves the internal inconsistency with feedback.

ethan-scitix (commented):
**`Question:` (please resolve before merge) — cross-subject aggregation is unanchored.** `report()` (`c_eval_kshot_ppl.py:219-220`) uses a macro-average, but the docstring (`:16-19`, `:211`) justifies it as "matching the C-Eval paper's average accuracy over all the subjects." The official scoring (`summary.json` `"All" = correct/num`) and lm-eval `ceval-valid` (`weight_by_size: true`) are both **micro**, and the cited `llama.py` doesn't aggregate across subjects at all — so macro has no upstream anchor. This is about *alignment*, not the number (empirically macro 90.19 vs micro 90.46 — ~0.26 apart). Please (a) confirm which aggregation you're aligning to and fix the docstring claim, and (b) record the comparison target like the GSM8K sibling — DSV3 Table 3 lists C-Eval 5-shot EM = 89.2 for Qwen2.5-72B-Base; your run lands at ~90.2.

**`Question:` — reference choice.** GSM8K ports from lm-eval-harness; this ports from ceval's own `llama.py`. lm-eval's `ceval-valid` is the same base-model loglikelihood eval and already pins split + aggregation. Why diverge? If `llama.py` is intentional, note in the docstring that it leaves aggregation undefined (so the choice of aggregation is your own).

**`Nit:`** `report()` `:214-216` — `if fb is not None` is dead for `finals` and inconsistent with `gsm8k_kshot_base_gen.py`; drop it.

**`Nit:`** `_select_examples` `:126-129` — silently truncates to `[:k]` with no `len(dev) < k` check; the GSM8K sibling raises. Harmless today (dev is always 5), worth matching.

jack-scitix-ai:
Thanks — addressed all four, and also migrated to the new `EvalMode.CLP` (#23) + updated the PR body.

**(1) Aggregation — it *is* anchored, just to the paper, not `llama.py`.** You're right that `evaluator_series/llama.py` doesn't aggregate and that lm-eval `ceval-valid` (`weight_by_size`) and the repo's `summary.json` are micro. But the anchor for macro is the **C-Eval paper**: Tables 3/4/5 define "Average" as "the average accuracy over all the subjects," and it checks out numerically — GPT-4 zero-shot, the unweighted mean of the 52 per-subject accuracies = 66.37 → reported 66.4 (a question-weighted/micro mean, or a mean-of-4-categories = 66.8, do **not** match). So macro is the paper's own metric; `evaluator_series` being per-subject is consistent with it. Fixed the docstring + notes to say exactly this and to flag that `ceval-valid` uses micro (`weight_by_size`) on val, so a reader knows what `score` is and how it differs.

- (b) Recorded the comparison target like the GSM8K sibling: DSV3 Table 3, C-Eval 5-shot 89.2 (Qwen2.5-72B-Base); our test-split run lands at 90.19 (macro), +0.99, within the <3% target.
- Re `summary.json`: that micro is a code convenience; the number the paper reports (and everyone cites, incl. DSV3) is the macro "Average." If the team would rather anchor to lm-eval/`summary.json` (micro), it's a one-line change — but our reading is the paper is the canonical definition, so we went macro and documented the divergence.

**(2) Reference choice — ceval's `llama.py`, not `ceval-valid`.** The per-question method is identical in both (loglikelihood-argmax over A/B/C/D), so the choice only decides split + aggregation. `ceval-valid` pins **val + micro**; we target **test + macro** — test labels are now public, that's the official leaderboard target and what DSV3's 89.2 compares against, and `ceval-valid` can't produce a test-split eval (val-only). So we aligned to the primary source (ceval's own evaluator + paper) rather than a third-party harness that re-derives a different aggregation. Documented that `evaluator_series` defines no cross-subject aggregation.

**Nit — `if fb is not None`:** dropped; direct `feedback_result` access like `gsm8k_kshot_base_gen`.

**Nit — `_select_examples`:** now raises `ValueError` when a subject has fewer than `k` dev exemplars, matching the GSM8K sibling.

**Also:** this task is `EvalMode.CLP` (the "next-token option-token log-prob" mode #23 introduced) — renamed `c_eval_kshot_ppl` → `c_eval_kshot_clp` (file/name/class) per the `_clp.py` convention. PR body updated accordingly.

ethan-scitix (commented):
**1. Align failure behavior with the CMMLU sibling (non-blocking):**

- `sieval/tasks/c_eval_kshot_clp.py:196,204-206` — when all four option logprobs come back `-inf`, `postprocess`'s `max()` silently returns `"A"`. The sibling raises on missing option tokens. Near-unreachable in practice (this run was 0-fail at `logprobs=20`), but worth a loud guard for consistency.
- `sieval/tasks/c_eval_kshot_clp.py:124-125` — `_ensure_few_shot_pool` silently returns when the `dev` split is absent; the sibling raises a clear error at setup (`cmmlu_kshot_base_gen.py:339-343`). Suggest failing early here too.

**2. Scoring mechanic — maintainer's call (non-blocking):** this task scores via **4 `echo` calls/sample**; the CMMLU sibling uses **1 `top_logprobs` call**. Both yield identical argmax when all four options are covered. `echo` is top-k-independent (hence 0 fails here at `logprobs=20`) but 4× the requests; `top_logprobs` is 1× but needs `--max-logprobs 100` + a loud-fail guard. Whether to unify the two Chinese-MCQ tasks (and which way — plus the `clp` vs `ppl` label) is your decision. Note switching this task to `top_logprobs` would need `logprobs≈100` and a fresh validation run — both existing runs used `echo`, so it's not pre-validated.

jack-scitix-ai:
Refactored to fully align this task with the CLP family (cmmlu #32 / mmlu #33) and resolved the earlier review points.

## Scoring mechanic → `top_logprobs` (unified with the family)
Dropped the 4×`echo` approach. Now a single `alogprobs(max_tokens=1, logprobs=100, echo=False)` call scored via the shared `choice_scores_from_top_logprobs` helper — the same mechanic as `cmmlu_kshot_clp` / `mmlu_kshot_clp`. Requires all four option tokens in the top-k and raises `RuntimeError` otherwise (needs `--max-logprobs 100` on vLLM; SGLang serves 100 by default), so partial coverage fails loudly instead of argmax-ing a subset.

## Report → per-category breakdown
Now emits `score`/`overall` (macro over the 52 subjects) + per-category macro (`stem` / `social_science` / `humanities` / `other`) + `fails`, mirroring the cmmlu report shape and the C-Eval paper's Table 3 category columns; absent categories are omitted (subject-subset runs). Cited the validated score in `reference_impl.notes` alongside the DSV3 target, matching the mmlu sibling.

## Earlier review points, resolved
- **`postprocess` silent `"A"`** → a missing option token now raises loudly (all-present guard), no longer defaults to `max()`'s first key.
- **`_ensure_few_shot_pool`** raises `ValueError` when the `dev` split is absent; **`_select_examples`** raises when a subject has fewer than `k` dev exemplars — matching the CMMLU sibling's early-fail.
- **Dropped the dead `if fb is not None`** in `report()`.
- **Aggregation is macro, anchored to the C-Eval paper** (Table 3, "average accuracy over all the subjects"; GPT-4's reported 66.4 equals the unweighted mean of the 52 per-subject accuracies). The lm-eval `ceval-valid` micro (`weight_by_size`) divergence is documented in the docstring.
- **`detect_truncated_output` clp false-fire**: resolved by rebasing onto main — #33's `applies_to=["gen"]` scopes it to generation tasks, so this clp task (which returns a `ModelOutput` at `max_tokens=1`) is no longer flagged.

## Validation
Re-validated on Qwen2.5-72B-Base (5-shot, test split, `logprobs=100`): **overall 90.19 macro, 0 fails / 12342**, per-category STEM 87.34 / Social Science 94.19 / Humanities 89.20 / Other 92.75.

ethan-scitix:
Pushed a small follow-up as maintainer (d3905d08) — hope you don't mind. It adds one regression test (`test_setup_raises_when_subject_has_too_few_dev_exemplars`) covering the `_select_examples` `len(dev) < k` guard you added earlier: that raise had **0 coverage** across the suite (confirmed via line trace), so this just pins it. No production code touched — test-only.

Everything else checks out on my end:
- Port is faithful to `llama.py` at the pinned SHA — English-key header, `format_example`/`\n答案：` marker, dev first-`k` same-subject shots, and next-token A/B/C/D argmax (equivalent to the reference `softmax(logits[A,B,C,D])`).
- The 52-subject category map matches C-Eval's official `subject_mapping.json` exactly (0 mismatches).
- HF revision resolves; all prior review points resolved.

LGTM once CI is green.

ethan-scitix (approved):

PR #17, opened 2026-06-28
fix(dataset): compare on-the-wire bytes so gzip responses aren't flag…

Mea1Ma:
## Type

## Summary

- **`url:` gzip truncation fix** — the download verifier compared bytes
  *written* against `Content-Length`, but when the server sends
  `Content-Encoding: gzip`, `Content-Length` is the compressed size while the
  decoded body is larger, so valid downloads were falsely flagged as truncated.
  Now compares against `num_bytes_downloaded` (on-the-wire bytes).

### Automated

### Required (all PRs)

### DIFF
--- a/sieval/datasets/downloaders/url.py
+++ b/sieval/datasets/downloaders/url.py
@@ -46,7 +46,7 @@ class URLHandler:
                 received = r.num_bytes_downloaded
                 if expected is not None and received != expected:
                     raise RuntimeError(
-                        f"truncated download from {url}: "
+                        f"size mismatch on download from {url}: "
                         f"Content-Length={expected} but received {received} bytes"
                     )
             tmp.replace(target)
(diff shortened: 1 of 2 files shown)


ethan-scitix:
Heads-up: I'm pushing two small maintainer-edit cleanups directly onto this branch (thanks for leaving "Allow edits by maintainers" on). The gzip fix itself is correct and approved — these are just polish:

1. **Reword the size-check error.** The guard rejects *any* `num_bytes_downloaded != Content-Length` mismatch — including an over-read (`received > expected`) — so `"truncated download"` was a slight misnomer for that direction. Changed to `"size mismatch on download from …"`; the test `match=` is updated to suit.

2. **Pin the gzip test docstring to the file RULER actually downloads.** `test_download_accepts_compressed_response` cited `train-v2.0.json` (`Content-Length=9551051`, ~42MB), but #11 fetches `dev-v2.0.json`. Both are gzip-served; I verified the dev figures (`Content-Length=800683`, decoded ~4.4MB) so the `Regression:` note now describes the real download path.

One follow-up for **#11**: it currently carries a copy of this same docstring with the old `train-v2.0.json` wording. Once this lands, please rebase #11 so it picks up the `dev-v2.0.json` correction (otherwise it'll reintroduce the stale numbers).

PR #19, opened 2026-06-29
feat(openbookqa): add dataset and k-shot generative tasks

jack-scitix-ai:
## Type

## Summary
- Add OpenBookQA benchmark: dataset loader + chat/instruct k-shot generative task.
- `openbookqa_kshot_gen` (chat): prompt lists A–D, model generates, letter extracted; accuracy metric.
- Prompt template and `first_option_postprocess` vendored byte-for-byte from OpenCompass `obqa_gen_9069e4` (`main` variant).
- Loader mirrors `allenai/openbookqa` as-is (revision-pinned), keeps `choices` native; A–D mapping by position lives in the task.
- `k=0` matches the OpenCompass 0-shot reference; `k>0` is a sieval extension (fixed first-k train rows). Reference: https://github.com/open-compass/opencompass/blob/5767b74899806c0c37efdc5529ffea01e7340e48/opencompass/configs/datasets/obqa/obqa_gen_9069e4.py

### Automated

### Manual
| --- | --- | --- | --- |
| Phi-4-mini-instruct (10-shot) | 79.2| 84.2 | +5.0 (+5.0%) |
| Qwen2.5-3B-Instruct (10-shot) | 79.3 | 82.6 | +3.3(3.3%) |

Note: the ReferenceImpl is OpenCompass `obqa_gen`, which publishes no OBQA score; the only published instruct comparison is the [Phi model card's number](https://huggingface.co/microsoft/Phi-4-mini-instruct), produced by a different (internal) harness, so the >3% gap is expected rather than a reproduction miss. `fails=0` in both runs.

### Required (all PRs)

### If: community/ Changes

### DIFF
--- a/sieval/tasks/openbookqa_kshot_gen.py
+++ b/sieval/tasks/openbookqa_kshot_gen.py
@@ -1,0 +1,185 @@
+"""OpenBookQA k-shot generative task (instruct/chat models).
+
+Accuracy metric; the predicted option letter is extracted with OpenCompass's
+``first_option_postprocess(options="ABCD")``. That extractor and the prompt
+template are vendored from OpenCompass ``obqa_gen_9069e4`` (``main`` variant)
+in ``sieval.community.openbookqa``. This task targets implementation parity
+with that config (prompt + extraction), not a specific published accuracy.
+
+Deviations from the OpenCompass reference (``obqa_gen_9069e4``):
+  - OpenCompass uses ``ZeroRetriever`` (0-shot). At ``k=0`` the prompt and
+    extraction match upstream; ``k>0`` is a sieval extension with no upstream
+    counterpart — the few-shot block is the first ``k`` ``train`` rows (fixed
+    indices), each with its ``answerKey`` appended. By default they are packed
+    into one user turn (the lm-eval/OpenCompass default); when
+    ``fewshot_as_multiturn`` is set they are rendered as alternating
+    user/assistant turns instead (lm-eval's ``fewshot_as_multiturn``).
+    At ``k>0`` generation is bounded by a stop sequence (the next example's
+    ``Question:`` header) so a verbose run-on cannot emit a later high-priority
+    extractor match that overrides the real answer; ``k=0`` is left unbounded to
+    match the upstream 0-shot config.
+  - Only the ``main`` variant is implemented; the ``additional``/``fact1``
+    ("Given the fact: ...") prompt variant is not used.
+  - Choices map to A–D by position (``choices["text"][0..3]``), matching
+    OpenCompass ``OBQADataset``.
+
+Repro decoding: greedy ``temperature=0``, ``top_p=1``. ``obqa_gen_9069e4`` sets
+no ``max_out_len``, so ``max_gen_toks`` follows the model/run config rather than
+a task-pinned value.
+
+AI-Generated Code - Opus 4.8 (Anthropic)
+"""
+
+from typing import TypedDict, override
+
+from openai.types.chat import ChatCompletionMessageParam
+
+from sieval.community.openbookqa import (
+    OBQA_OPTIONS,
+    OBQA_PROMPT_TEMPLATE,
+    first_option_postprocess,
+)
+from sieval.core.models import ModelOutput
+from sieval.core.tasks import (
+    EvalMode,
+    ReferenceImpl,
+    Task,
+    sieval_task,
+)
+from sieval.datasets import OpenBookQADatasetSample
+
+DEFAULT_N_SHOT = 0
+FEWSHOT_SEP = "\n\n"
+# Coupled to the few-shot block: each packed example begins with "Question:".
+# Applied only at k>0 to bound verbose run-on (see infer); k=0 stays unbounded.
+STOP_SEQUENCES = ("\nQuestion:",)
+
+
+class Feedback(TypedDict):
+    correct: bool
+    pred: str
+    answer: str
+
+
+def _format_question(sample: OpenBookQADatasetSample) -> str:
+    texts = sample["choices"]["text"]
+    return OBQA_PROMPT_TEMPLATE.format(
+        question_stem=sample["question_stem"],
+        A=texts[0],
+        B=texts[1],
+        C=texts[2],
+        D=texts[3],
+    )
+
+
+@sieval_task(
+    name="openbookqa_kshot_gen",
+    display_name="OpenBookQA (k-shot, generative)",
+    description="OpenBookQA elementary-science MCQ, generative letter extraction.",
+    eval_mode=EvalMode.GEN,
+    n_shot=DEFAULT_N_SHOT,
+    tags=("english", "science", "multiple-choice"),
+    model_type="chat",
+    reference_impl=ReferenceImpl(
+        source="opencompass",
+        url="https://github.com/open-compass/opencompass/blob/5767b74899806c0c37efdc5529ffea01e7340e48/opencompass/configs/datasets/obqa/obqa_gen_9069e4.py",
+        notes=(
+            "Prompt template (main variant) and first_option_postprocess "
+            "vendored from OpenCompass. At k=0 the prompt and extraction match "
+            "the upstream 0-shot config; k>0 is a sieval extension (fixed "
+            "first-k train rows)."
+        ),
+    ),
+)
+class OpenBookQAFewShotGenTask(
+    Task[
+        OpenBookQADatasetSample,
+        list[ChatCompletionMessageParam],
+        ModelOutput,
+        str,
+        Feedback,
+        dict[str, float],
+    ]
+):
+    def __init__(
+        self,
+        dataset,
+        model,
+        name: str | None = None,
+        *,
+        k: int = DEFAULT_N_SHOT,
+        fewshot_split: str = "train",
+        fewshot_as_multiturn: bool = False,
+        stop: tuple[str, ...] = STOP_SEQUENCES,
+    ):
+        if k < 0:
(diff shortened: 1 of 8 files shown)


sieval/tasks/openbookqa_kshot_gen.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+    @override
+    async def preprocess(self, raw, ctx):
+        prefix = self._fewshot_prefix if self._fewshot_prefix is not None else ""
+        return [{"role": "user", "content": prefix + format_question(raw)}]
+
+    @override
+    async def infer(self, pre, ctx):
+        return await self.model.agenerate(pre)
```
ethan-scitix, line 156: **`Bug:` (k>0 chat) run-on corrupts extraction for verbose models**

At `k>0`, few-shot is packed into one user turn (`openbookqa_kshot_gen.py:138`) with no stop here. Verbose models answer, then keep generating (re-answering the bundled examples). Since `first_option_postprocess` matches ~50 patterns **by priority over the whole text** — not first-by-position — a trailing match can override the model's leading answer.

Verified on a Qwen2.5-3B 10-shot run (temp=0, max_tokens=256, 500-sample test split): **7/500** samples where the model's leading answer is correct but the extractor picks a letter from the run-on → reported **82.6** vs **~84.0**. Phi-4-mini (terse): **0** — a silent, model-dependent bias against verbose models. (sample 40, answerKey D: `Answer: D … C. you will feel the h…` → returns C.)

Default `k=0` is unaffected (no bundled block → no run-on), so the 0-shot path is fine. The extractor must stay as-is (upstream parity), so bound generation at `k>0` — a stop sequence, or multi-turn few-shot. Both are fine; single-turn packing isn't wrong (it's the lm-eval/OpenCompass default), it just needs a bound at `k>0`.

Cross-check: DROP shares the single-turn/no-stop setup but isn't affected — its `re.search(ANSWER_PATTERN)` takes the first positional match. So this is specific to OBQA's position-agnostic extractor at `k>0`, not the few-shot format itself.

sieval/tasks/openbookqa_kshot_base_gen.py
```diff
@@ -0,0 +1,128 @@
...
+from sieval.community.openbookqa import OBQA_OPTIONS, first_option_postprocess
+from sieval.core.models import ModelOutput
+from sieval.core.tasks import (
+    EvalMode,
+    ReferenceImpl,
+    Task,
+    sieval_task,
+)
+from sieval.datasets import OpenBookQADatasetSample
+from sieval.tasks.openbookqa_kshot_gen import (
```
ethan-scitix: **`Nit:` task-to-task import**

This imports `DEFAULT_N_SHOT, Feedback, build_fewshot_prefix, format_question` from `openbookqa_kshot_gen`, so the completion task module depends on the chat task module at import time. Nothing flags it (no import-linter; `check_dep_coverage` skips `sieval.*`), and it's benign now (both `deps_group=null`) — but if either task later gains a `deps_group`, the sibling silently inherits it at import time and the dep check won't warn.

The shared bits are small, so either inline them in each task (duplication is fine if the tasks may diverge) or pull them into a neutral `tasks/_openbookqa.py` — just so neither task imports the other. (Can't live in `community/`: `build_fewshot_prefix` needs `datasets/`.)

ethan-scitix (commented):
Reviewed the port + ran your three `outputs/` 10-shot runs. The port is faithful (prompt template + `first_option_postprocess` are byte-for-byte against `obqa_gen_9069e4`, incl. the missing-comma quirk; `OBQADataset` A–D-by-position mapping matches), the pipeline is validated end-to-end, and the **0-shot default path is solid**. Two non-blocking notes below — neither affects `k=0`. Approve.

jack-scitix-ai:
Thanks for the careful review — both addressed in the latest commit.

## `Bug:` k>0 run-on corrupts extraction

Fixed by bounding generation at `k>0`, leaving the extractor and the `k=0` path untouched (per your constraints):

- Added `STOP_SEQUENCES = ("\nQuestion:",)`, coupled to the packed few-shot block (each example begins with `Question:`).
- `infer()` applies the stop **only when `k>0`** (`if self._k > 0 and self._stop`). At `k=0` there is no bundled block, so generation stays unbounded and byte-for-byte faithful to the upstream 0-shot config.
- `first_option_postprocess` is unchanged — still the verbatim upstream extractor. The stop just prevents the verbose run-on from emitting the later end-anchored match you traced (e.g. Qwen sample 40: `Answer: D` … run-on … `Answer: C` → the trailing `C` is cut before extraction).

Re-ran with the stop toggled (10-shot, temp=0, max_tokens=256, 500-sample test):

| Model | no stop | with `k>0` stop |
| --- | --- | --- |
| Qwen2.5-3B (verbose) | 82.6 | **83.8** |
| Phi-4-mini (terse) | 84.6 | **84.6** |

On Qwen the stop recovers 6 of the 7 parse-victims you flagged; the 7th is an *inline* continuation with no `Question:` boundary, so a stop can't bound it — an inherent limit of a position-agnostic extractor on inline run-on, not something a stop addresses. Phi is unchanged, confirming terse models are unaffected as you predicted.

## Single-turn vs multi-turn (follow-up on your "stop or multi-turn" note)

I added an optional `fewshot_as_multiturn` and compared it against the single-turn default on Phi-4-mini. Counterintuitively multi-turn is **lower**, 84.6 → 82.8, and it's **not** an extraction artifact — both formats emit a clean single letter for every sample (0 unparseable in either); multi-turn simply produces more genuinely-wrong answers (86 vs 77 wrong / 500). So single-turn packing (the lm-eval/OpenCompass default) is both upstream-faithful and empirically stronger here; multi-turn stays available but off by default.

Tests added:
- `test_stop_sequences_pinned`
- `test_infer_bounds_generation_at_kshot_but_not_zero_shot` — asserts `stop` is forwarded at `k>0` and **not** at `k=0`.
- `test_multiturn_renders_alternating_user_assistant_turns`.

## `Nit:` task-to-task import

Removed — the completion (`_base_gen`) task is dropped from this PR, which is now scoped to the instruct/chat task only. That deletes the cross-module import entirely; the shared helpers are folded back into the chat task as private (`_format_question`, `_retrieve_fewshot`). If we reintroduce the completion task later, it'll go through a neutral `tasks/_openbookqa.py` rather than importing its sibling, as you suggested.

ethan-scitix (commented):
**Reproduction verified** against OpenCompass `obqa_gen_9069e4@5767b748`: the `main` prompt template is byte-identical, `first_option_postprocess` is vendored verbatim (incl. the upstream missing-comma quirk at `sieval/community/openbookqa.py:44-45`), decoding params are correctly left to the model config, `deps_group` is correct, preflight passes, and 13/13 tests pass. The pinned HF revision `388097ea` resolves to the same 500-example OBQA `main` test split OpenCompass scores. Two non-blocking nits below.

---

`Nit:` `sieval/datasets/openbookqa.py:33` — annotate the data-source substitution. This is OpenCompass's code (prompt + extractor) run against a different data path: upstream's `OBQADataset` reads `opencompass/openbookqa_test`, this loads `allenai/openbookqa` (`main`). I verified they're equivalent for scoring — same 500-example test split, both position-map `choices.text[0..3]`→A–D, and every record's `choices.label` is ordered `[A,B,C,D]` so `answerKey` aligns; no `" what?"` rewrite (that's `OBQADatasetV2`). Since it's one project's code on another's data, please add one line (dataset docstring or `ReferenceImpl.notes`) noting the source repo differs from upstream's mirror but resolves to the same test set — so the next reader doesn't have to re-verify.

`Nit:` `sieval/tasks/openbookqa_kshot_gen.py:149` — match the sibling stop-forwarding convention. `kwargs: dict[str, object]` is the only `dict[str, object]` in `sieval/tasks/`. Every sibling that conditionally forwards `stop` branches and passes it inline instead of building an intermediate dict — `gsm8k_kshot_base_gen.py:172-173`, `theoremqa_kshot_base_gen.py:457`, `livecodebench_code_generation_kshot_base_gen.py:178-179`:

```python
if self._k > 0 and self._stop:
    return await self.model.agenerate(pre, stop=list(self._stop))
return await self.model.agenerate(pre)
```

Drops the `object` widening and aligns with the family (behavior-preserving — existing `infer` tests still pass).

---

Both nits are optional-but-recommended before merge. The branch is 1 commit behind `origin/main` — worth a rebase before merging.

jack-scitix-ai:
Thanks — both nits addressed, and rebased onto latest `origin/main`.

**Nit 1 (data-source annotation):** Added a "Source note" to the `openbookqa.py` module docstring: OpenCompass's `obqa_gen_9069e4` runs against its own mirror `opencompass/openbookqa_test`, while this loader points at `allenai/openbookqa` (`main`); the note records that they resolve to the same 500-example `main` test split (position-map `choices.text[0..3]`→A–D, `choices.label` ordered `[A,B,C,D]` so `answerKey` aligns, and the `" what?"` rewrite is `OBQADatasetV2`, not this config) — so the next reader doesn't have to re-verify.

**Nit 2 (stop-forwarding convention):** Dropped the intermediate `kwargs: dict[str, object]`; `infer` now branches and passes `stop` inline, matching `gsm8k_kshot_base_gen` / `theoremqa_kshot_base_gen` / `livecodebench_code_generation_kshot_base_gen`:

​```python
if self._k > 0 and self._stop:
    return await self.model.agenerate(pre, stop=list(self._stop))
return await self.model.agenerate(pre)
​```

Behavior-preserving — the `infer` tests (`test_infer_bounds_generation_at_kshot_but_not_zero_shot`) still pass, and `ty` is clean with the `object` widening gone.

PR #21, opened 2026-06-30
feat(models): add SglangGenModel for echoed-input logprobs via sglang /generate

jack-scitix-ai:
## Type

- feature — new capability

## Summary
- Add `SglangGenModel(Model[str])` speaking sglang's native `/generate` protocol for both generation and logprob extraction. sglang's OpenAI `/v1/completions` rejects `echo=True + logprobs`, blocking echo-based PPL scoring (ARC/HellaSwag) and top-k scoring (CMMLU/MMLU-Base) on the sglang backend.
- Overrides both `_agenerate_impl` (native `/generate`) and `_alogprobs_impl` (echoed-input logprobs + first-token top-k). Extends `Model[str]` rather than `GenModel` — inheriting `GenModel` would only add its OpenAI-completions `_agenerate_impl`, a different wire protocol (incidental reuse, not coupling).
- Parses both `*_token_logprobs` (echo) and `*_top_logprobs` (top-k, CMMLU); normalizes GPT-2 byte-level token text (`Ġ`→space) — SentencePiece `▁` (U+2581) passes through unchanged (fails loud, not mis-scored).
- Radix-cache guard: fails loud when sglang's prefix cache truncates echoed input logprobs (advises `--disable-radix-cache`).
- Dispatch via new model-config field `engine: sglang`; rejected on chat and derived models. `Task._validate_model_type` recognizes it as a "gen" model.

### Automated

### Manual

- [x] Live vs Qwen2.5-72B (bf16, tp=2) sglang server: echo (ARC-style) and
      top-k (CMMLU-style) agree (`B`, -0.007); native generation OK.
- [x] Full `cmmlu_5shot_base_gen` with `engine: sglang`, 200 samples:
      score 87.2% (stem 83.9 / other 90.5), 0 fails.

### Required (all PRs)

### DIFF
--- a/sieval/cli/leaderboard/session.py
+++ b/sieval/cli/leaderboard/session.py
@@ -24,7 +24,7 @@ from loguru import logger
 
 from sieval.cli.leaderboard.card import AlignmentCard, load_card
 from sieval.core.datasets import Dataset
-from sieval.core.models import ChatModel, GenModel, Model
+from sieval.core.models import ChatModel, GenModel, Model, SglangGenModel
 from sieval.core.runners import MultiTaskRunner, TaskRunnerConfig
 from sieval.core.tasks.context import TaskAction
 from sieval.core.types import JSONValue
@@ -62,6 +62,7 @@ class _InferMetaDict(TypedDict, total=False):
 class ModelConfigDict(TypedDict, total=False):
     name: str  # For base models
     type: Literal["chat", "gen"]  # "chat" or "gen" (default: "chat")
+    engine: Literal["vllm", "sglang"]  # gen backend (default: "vllm")
     base: str  # For derived models
     args: dict[str, Any]
     api_key: str
@@ -978,7 +979,16 @@ class EvalSession:
             model_type = self._infer_model_type(name, explicit_type)
 
             if model_type == "gen":
-                self.models[name] = GenModel(model=model_name, **args)
+                engine = cfg.get("engine", "vllm")
+                if engine == "sglang":
+                    self.models[name] = SglangGenModel(model=model_name, **args)
+                elif engine == "vllm":
+                    self.models[name] = GenModel(model=model_name, **args)
+                else:
+                    raise ValueError(
+                        f"Model '{name}' has invalid engine '{engine}'. "
+                        "Expected 'vllm' or 'sglang'"
+                    )
             elif model_type == "chat":
                 self.models[name] = ChatModel(model=model_name, **args)
             else:
--- a/sieval/cli/leaderboard/session.py
+++ b/sieval/cli/leaderboard/session.py
@@ -978,6 +978,13 @@ class EvalSession:
             explicit_type = cfg.get("type")
             model_type = self._infer_model_type(name, explicit_type)
 
+            # `engine` selects the gen backend; it is meaningless for chat.
+            if "engine" in cfg and model_type != "gen":
+                raise ValueError(
+                    f"Model '{name}': 'engine' is only valid for type: gen, "
+                    f"but this model is '{model_type}'."
+                )
+
             if model_type == "gen":
                 engine = cfg.get("engine", "vllm")
                 if engine == "sglang":
@@ -1030,6 +1037,12 @@ class EvalSession:
                     raise ValueError(
                         f"Derived model '{name}' cannot override api_key/api_base from "
                         f"base model '{base_name}'. Create a new base model instead."
+                    )
+
+                if "engine" in cfg:
+                    raise ValueError(
+                        f"Derived model '{name}' cannot set 'engine'; it is inherited "
+                        f"from base model '{base_name}'. Set it on the base instead."
                     )
 
                 # Extract concurrency_limit separately for with_args
--- a/sieval/cli/leaderboard/session.py
+++ b/sieval/cli/leaderboard/session.py
@@ -1050,23 +1050,35 @@ class EvalSession:
 
                 # Check if type conversion is needed
                 target_type = cfg.get("type")
-                if target_type:
-                    # Convert to target type
-                    if target_type == "gen":
-                        new_model = base_model.as_type(GenModel)
-                    elif target_type == "chat":
-                        new_model = base_model.as_type(ChatModel)
+                if target_type == "gen":
+                    # An sglang-backed base is already "gen"; as_type(GenModel)
+                    # would swap it to the OpenAI /v1/completions path (which
+                    # rejects echo+logprobs), silently defeating engine: sglang.
+                    # Preserve it. A plain GenModel base is unaffected.
+                    if isinstance(base_model, SglangGenModel):
+                        new_model = base_model
                     else:
-                        raise ValueError(
-                            f"Model '{name}' has invalid type '{target_type}'. "
-                            "Expected 'chat' or 'gen'"
-                        )
+                        new_model = base_model.as_type(GenModel)
                     logger.info(
                         "Created derived model '{}' from '{}' "
                         "with type conversion to '{}'",
                         name,
                         base_name,
                         target_type,
+                    )
+                elif target_type == "chat":
+                    new_model = base_model.as_type(ChatModel)
+                    logger.info(
+                        "Created derived model '{}' from '{}' "
+                        "with type conversion to '{}'",
+                        name,
+                        base_name,
+                        target_type,
+                    )
+                elif target_type:
+                    raise ValueError(
+                        f"Model '{name}' has invalid type '{target_type}'. "
+                        "Expected 'chat' or 'gen'"
                     )
                 else:
                     # No type conversion, just derive
--- a/sieval/cli/validation.py
+++ b/sieval/cli/validation.py
@@ -153,6 +153,25 @@ def _validate_models(cfg: dict, result: ValidationResult) -> None:
                 f"Model '{name}': type must be 'chat' or 'gen', got '{model_type}'"
             )
 
+        # `engine` selects the gen backend (mirrors _setup_models' guards; the
+        # engine-on-non-gen check there uses the *inferred* type, so here we
+        # can only flag the statically-decidable explicit `type: chat` case).
+        if "engine" in mcfg:
+            engine = mcfg.get("engine")
+            if has_base:
+                result.errors.append(
+                    f"Model '{name}': derived models cannot set 'engine'; it is "
+                    "inherited from the base model. Set it on the base instead."
+                )
+            elif engine not in ("vllm", "sglang"):
+                result.errors.append(
+                    f"Model '{name}': engine must be 'vllm' or 'sglang', got {engine!r}"
+                )
+            elif model_type == "chat":
+                result.errors.append(
+                    f"Model '{name}': 'engine' is only valid for type: gen, not 'chat'"
+                )
+
         if has_base:
             base_ref = mcfg["base"]
             if not isinstance(base_ref, str) or not base_ref:
--- a/sieval/core/models/__init__.py
+++ b/sieval/core/models/__init__.py
@@ -1,6 +1,7 @@
 from .chat_model import ChatModel
 from .gen_model import GenModel
 from .model import Model, ModelCallMeta, ModelMeta, ModelOutput, ModelUsage
+from .sglang_gen_model import SglangGenModel
 
 __all__ = [
     "ChatModel",
@@ -10,4 +11,5 @@ __all__ = [
     "ModelMeta",
     "ModelOutput",
     "ModelUsage",
+    "SglangGenModel",
 ]
--- a/sieval/core/models/model.py
+++ b/sieval/core/models/model.py
@@ -66,6 +66,11 @@ class ModelOutput:
     texts: list[str]
     finish_reasons: list[str] | None = None
     reasoning_texts: list[str] | None = None
+    # Token texts follow the OpenAI / literal-whitespace convention (leading
+    # spaces preserved, e.g. " A"). Backends that emit other markers (e.g.
+    # sglang byte-level "ĠA") normalize to this contract before populating it,
+    # so consumers can match on literal whitespace rather than per-tokenizer
+    # markers.
     logprobs_tokens: list[str] | None = None
     logprobs: list[float | None] | None = None
     top_logprobs: list[dict[str, float]] | None = None
--- a/sieval/core/models/sglang_gen_model.py
+++ b/sieval/core/models/sglang_gen_model.py
@@ -1,0 +1,135 @@
+"""SglangGenModel: echoed-input logprobs via sglang native /generate.
+
+sglang's OpenAI ``/v1/completions`` endpoint rejects ``echo=True`` together
+with ``logprobs``, so PPL-style tasks (ARC, MMLU-Base, HellaSwag) that read
+the logprob of an answer token appended to the prompt cannot use it. This
+variant overrides only logprob extraction to call the native ``/generate``
+endpoint (``return_logprob=True`` + ``logprob_start_len=0``), which returns
+per-token logprobs over the full echoed input sequence.
+
+AI-Generated Code - Claude Opus 4.8 (Anthropic)
+"""
+
+from typing import override
+
+from sieval.core.types import JSONValue
+
+from .gen_model import GenModel
+from .model import ModelOutput, ModelUsage
+
+
+def _normalize_token_text(text: str) -> str:
+    """Map GPT-2 byte-level BPE markers back to literal whitespace.
+
+    sglang detokenizes when ``return_text_in_logprobs=True``, but some
+    tokenizers (e.g. Qwen) surface the raw byte-level markers ``Ġ`` (space)
+    and ``Ċ`` (newline). ``extract_option_logprob`` matches ``" A"`` /
+    ``A``, so an un-normalized ``"ĠA"`` would silently never match and the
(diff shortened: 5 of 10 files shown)


ethan-scitix:
**Verdict: request changes.** Verified: sglang `/v1/completions` really does reject `echo`+`logprobs` (sgl-project/sglang#782); the `/generate` format matches the parser; 27 tests + `ty` pass.

**Blocking — `top_logprobs` requested but never parsed → CMMLU breaks on `engine: sglang`.** `sglang_gen_model.py:114` sends `top_logprobs_num` and accepts `echo=False`, but `_parse_logprobs` (`sglang_gen_model.py:49-66`) ignores `output_top_logprobs`, leaving `ModelOutput.top_logprobs=None`. CMMLU — the only `alogprobs` consumer, `cmmlu_kshot_base_gen.py:402-410` — then raises on every sample. Fix: parse `output_top_logprobs[0]` into the existing `list[dict[str, float]]` shape (normalize token text the same way); or, if this backend is echo-only, drop the `echo=False` branch and `raise` on it. The current middle ground (accept `echo=False`, request top-k, return none) is the worst case.

**Architecture — extend `Model[str]`, not `GenModel`.** As written it's a chimera: generation via the inherited OpenAI `/v1/completions` path, logprobs via native `/generate` — one object speaking two wire protocols. The genuinely shared infra (client, limiters, `with_args`/`meta`, the public wrappers) lives in `Model`; inheriting `GenModel` only buys the OpenAI-completions `_agenerate_impl` we don't want here (incidental reuse, not coupling). Recommend moving to `SglangGenModel(Model[str])` with a native `_agenerate_impl` over `/generate`.

**Nits:**
- `sglang_gen_model.py:118` reaches openai's private `_client` httpx wrapper, bypassing the configured `max_retries` — the completions path gets retries, this raw POST doesn't. Prefer the public `self._client.post(...)` or an explicit client, and note the deviation.
- `finish_reasons` not set on the returned `ModelOutput` (`sglang_gen_model.py:127`) — base `alogprobs` sets it; harmless for current consumers but a silent divergence.
- `engine` is silently ignored for `type: chat` and derived (`base:`) models (`session.py:829-839` only branches inside `gen`) — consider rejecting or documenting.

**Deferred (not this PR):** the `token_ids` output field and the OpenAI-shaped `Model.alogprobs` interface are tracked in RFC #25 (capability-based Model IR); the resume version gate that makes that redesign safe to defer is RFC #24.

Nice: `_normalize_token_text` and its deviation comment are exactly right — the `Ġ`/`Ċ` handling is correct and the rationale is documented.

ethan-scitix (changes requested):
Requesting changes — full review in the comment above (blocking: `top_logprobs` dropped → CMMLU breaks on `engine: sglang`).

jack-scitix-ai:
Thanks for the thorough review — all points addressed in the latest commit. Verified live against a Qwen2.5-72B sglang server: native generation, echo-logprob (ARC-style), and top-k (CMMLU-style) all work; the echo and top-k paths produce identical A/B/C/D logprobs on the same prompt (`B`, `-0.007`).

**Blocking — `top_logprobs` dropped → CMMLU breaks.** Fixed. Added `_parse_top_logprobs` flattening `input_top_logprobs`/`output_top_logprobs` into the existing `list[dict[str, float]]` shape (index-aligned with the token list; input segment first under `echo`; `None`/empty per-token entry → `{}`; normalized token text). Unit-tested a real edge case seen live: two tokens stripping to the same letter (`' B'` at `-0.007` vs `'\tB'` at `-11.94`) stay as distinct keys, so CMMLU's `max`-over-strip recovers the high logprob.
- **End-to-end proof:** ran the actual `cmmlu_5shot_base_gen` task with `engine: sglang` (Qwen2.5-72B, 200 samples): **score 87.2% (stem 83.9 / other 90.5), 0 fails**. The `all_present` guard never tripped — every sample had A/B/C/D within top-100.

**Architecture — extend `Model[str]`, not `GenModel`.** Done. `SglangGenModel(Model[str])` with a native `_agenerate_impl` over `/generate` (sampling-param translation, `n>1` list response, usage aggregation) — one wire protocol end-to-end. Ripple: `Task._validate_model_type` also accepts `SglangGenModel` as "gen" now (with a test).

**Nit — private `_client._client` bypasses `max_retries`.** Fixed — now `self._client.post(<abs-url>, cast_to=object, body=...)`, which preserves auth and retries and correctly hits `/generate` (absolute URL, since the client appends to the `/v1` base otherwise).

**Nit — `finish_reasons` not set.** Fixed — extracted from `meta_info.finish_reason` on both generation and logprob outputs.

**Nit — `engine` silently ignored for chat/derived models.** Now rejected with a clear message on `type: chat` and on derived (`base:`) models (tests added).

ethan-scitix:
### Review follow-ups on `SglangGenModel`

Verified empirically: the sglang `/generate` protocol matches (field names, `[logprob, token_id, token_text]` entry shape, `logprob_start_len` 0/-1 semantics), all 61 new tests pass, engine dispatch is correctly guarded, and CMMLU (`echo=False`, the only current consumer) is correct. A few things to address:

**1. `Bug:` (blocking) — `echo=True` silently returns partial logprobs under sglang's radix cache.** sglang's prefix cache (on by default) truncates `input_token_logprobs` to `prompt_tokens − cached_tokens` on a cache hit — it does not recompute cached positions. So the input-logprob path (`logprob_start_len=0`, `sglang_gen_model.py:240`) can return a partial, non-reproducible token set once a prefix is warm. CMMLU (`echo=False`, output-only) is unaffected, but the echo-based PPL tasks this model is built for (HellaSwag/ARC via `total_logprob`) would score wrong silently. Add a fail-loud guard in `_alogprobs_impl`: when `echo=True`, require the returned input-logprob count to equal `prompt_tokens` (or `meta_info.cached_tokens == 0`); otherwise raise, telling the operator to launch with `--disable-radix-cache`.

> Note: vLLM behaves differently here (V1 makes `prompt_logprobs`/echo mutually exclusive with prefix caching and *errors* rather than truncating), so this is a sglang-specific silent-failure mode — hence the loud guard. The serving-recipe side (threading a "prefix-cache off" requirement through the recipe system) is tracked separately and is out of scope for this PR.

**2. `Nit:` — document the `_client.post` tech debt.** Reusing the OpenAI SDK's low-level `.post` to speak native `/generate` (`sglang_gen_model.py:67-76`) is fine for now (keeps auth/retries) but couples us to an SDK-internal surface. Add a comment pointing to the decoupling RFC #25.

**3. `Nit:` — make the token-text rewrite explicit.** `_normalize_token_text` (`sglang_gen_model.py:42`) only handles GPT-2 byte-level markers (`Ġ`/`Ċ`); SentencePiece (`▁`) and other markers pass through unchanged — state this limitation in the docstring. Also note on `ModelOutput.logprobs_tokens` (`model.py:69`) that token text is normalized to the OpenAI/literal-whitespace convention, so consumers rely on a stated contract rather than a silent transform.

**4. `Nit:` — drop explicit `Any`.** `dict[str, Any]` → `dict`, `list[Any]` → `list`; and tighten `_post(...) -> Any` to `-> dict | list` (the callers already narrow with `isinstance`).

jack-scitix-ai:
All four addressed.

**1. Bug (blocking) — radix-cache truncation.** Added a fail-loud guard in `_alogprobs_impl`: when `echo=True`, if `meta_info.cached_tokens != 0` (or the returned input-logprob count != `prompt_tokens`), it raises and tells the operator to launch sglang with `--disable-radix-cache`. This turns the silent partial-logprob failure into a startup-visible one. CMMLU (`echo=False`) is unaffected — verified by a test that a cache hit under `echo=False` does not raise. Tests cover: cache-hit raise, count-mismatch raise, full-input pass, echo=False ignores cache.

**2. Nit — `_client.post` tech debt.** Documented in `_post`'s docstring: reuses the SDK's low-level `.post` (keeps auth/retries) at the cost of coupling to an SDK-internal surface; points to RFC #25 for the decoupling.

**3. Nit — token-text rewrite scope.** `_normalize_token_text` docstring now states it only handles GPT-2 byte-level markers (`Ġ`/`Ċ`); SentencePiece `▁` and others pass through. Added a contract note on `ModelOutput.logprobs_tokens` that token text follows the OpenAI/literal-whitespace convention so consumers match on whitespace, not per-tokenizer markers.

**4. Nit — drop explicit `Any`.** `dict[str, Any]`→`dict`, `list[Any]`→`list`, and `_post(...) -> Any` tightened to `-> dict | list` (callers narrow with `isinstance`; `_alogprobs_impl` now rejects a non-dict response explicitly).

ethan-scitix (commented):
**Verdict: approve with follow-ups.** Verified the happy path end-to-end — 41 tests pass, 100% line coverage on `sglang_gen_model.py`, and the radix-cache guard's `logprob_start_len=0 → len(input_token_logprobs) == prompt_tokens` invariant matches sglang's documented behavior. None of the below block the core scoring path; they're consistency/robustness items.

**1. `Question:` — `request_params` carries the full prompt and diverges in shape** · `sieval/core/models/sglang_gen_model.py:220`, `:296`
Both impls set `request_params=body`, which `build_model_call_meta` (`sieval/core/utils/meta.py:15`) copies verbatim into every persisted record. Vs `GenModel`/`ChatModel` (which store `dict(final_kwargs)`): (a) `body["text"]` duplicates the full prompt into per-call metadata on every call — please keep the prompt out; (b) the remaining shape differs (`sampling_params.max_new_tokens` vs flat OpenAI knobs). Nothing reads `request_params` today, so full normalization is arguably RFC #25's job — either normalize to the `GenModel` shape or leave a one-line comment marking this as the pre-#25 wire shape.

**2. `Question:` — silent engine downcast via `as_type`; should fail loud** · `sieval/cli/leaderboard/session.py:1056`
An explicit `type: gen` on an sglang-based derived model hits `as_type(GenModel)`, which rewrites `__class__` to `GenModel` — silently swapping to OpenAI `/v1/completions` (the endpoint the docstring says rejects `echo`+`logprobs`). This contradicts the engine-inheritance guard added at `:1042`. Suggest: `raise` when the base is engine-specific and `type` would re-convert it — or only call `as_type` when the target kind actually differs and keep the base instance otherwise (preserves sglang, no-op for vanilla gen).

**3. `Nit:` — `_agenerate_impl` response handling should match the logprobs path** · `sieval/core/models/sglang_gen_model.py:198-199`
`_alogprobs_impl` validates its response (`:256`); `_agenerate_impl` doesn't — `metas = [r["meta_info"] for r in results]` surfaces a bare `KeyError` on a malformed response while `texts` uses `.get`. Mirror the sibling's explicit check.

**4. `Question:` — confirm the `cached_tokens` guard intent + `prompt_tokens`-absent gap** · `sieval/core/models/sglang_gen_model.py:271`
The length check (`:272`) already catches truncation; the `cached_tokens` clause additionally rejects a full-length hit. If that's a deliberate "any cache touch = untrusted" stance (guards full-length-but-stale logprobs), please note it and add a test pinning that branch — there's none today. Separately: if a response omits `prompt_tokens` (`:269`), both this guard and the empty-check at `:285` are bypassed and a truncated-but-non-empty set slips through → consider raising on missing `prompt_tokens` rather than passing.

jack-scitix-ai:
Thanks — all four follow-ups addressed in the latest commit.

**1. `request_params` duplicated the full prompt / shape differed.** Added a `_request_params(body)` helper that strips `text` from the persisted params (both `_agenerate_impl` and `_alogprobs_impl`), so the prompt — already recorded as the sample input — no longer gets copied into every per-call record. Documented that the remaining shape is intentionally sglang-native (differs from the OpenAI-flavoured `GenModel`/`ChatModel`), with a pointer to RFC #25 for the client/protocol decoupling that would unify it.

**2. `type: gen` on an sglang-backed derived model silently swapped to `/v1/completions`.** Fixed in `_setup_models`: the `gen` branch now preserves the base when it's a `SglangGenModel` instead of calling `as_type(GenModel)` (which would route to the endpoint that rejects echo+logprobs). Plain `GenModel` bases and chat↔gen conversions are unchanged. Test added (`test_derived_type_gen_preserves_sglang_base`).

**3. `_agenerate_impl` surfaced a bare `KeyError` on a malformed response.** Now mirrors `_alogprobs_impl`'s explicit check: raises `RuntimeError("... missing meta_info")` if the response (or any element of the `n>1` list) isn't a dict carrying `meta_info`, before touching `r["meta_info"]`. Test added.

**4. Radix-cache guard intent + missing `prompt_tokens`.** Confirmed the deliberate stance and made it explicit in the comment: **any** cache touch — or a response we can't verify — is treated as untrusted, and echo-based scoring requires `--disable-radix-cache`. Also closed the bypass you flagged: when `echo=True` and the response omits `prompt_tokens`, completeness can't be verified, so it now raises rather than letting possibly-truncated logprobs through. Added guard tests for the cache-hit, count-mismatch, and missing-`prompt_tokens` cases (plus an `echo=False` case that's correctly unaffected).

ethan-scitix (approved):
**Verdict: Approve.** No blocking bugs — careful, well-documented work. What I verified empirically:

- All 277 tests in the touched files pass; `sglang_gen_model.py` has **100% line + branch coverage** (118 stmts / 44 branches, 0 missed).
- `openai==2.9.0` exposes `.post(path, *, cast_to, body)` exactly as used, and httpx's absolute-URL join rewrites the base — so `_generate_url()` correctly routes to `/generate`, not `/v1/generate`.
- Every non-trivial sglang claim checks out against sglang's docs/issues: `logprob_start_len` −1/0 semantics, the `(logprob, token_id, token_text)` tuple shape, the leading-`None` first token, and the radix-cache truncation (`cached_tokens ≈ prompt_tokens`, logprobs not recomputed for cached positions, `--disable-radix-cache` as the fix). The guard's `len(input_lps) == prompt_tokens` invariant matches the cold-cache output. I also confirmed the `Ġ→space` normalization is load-bearing: without it, CMMLU on Qwen+sglang wouldn't match (`"ĠA".strip() != "A"`).

The core architectural decision (extend `Model[str]` + translate params internally, rather than subclass `GenModel`) is exactly the interim path RFC #25 sanctions, and `_post` / `_request_params` already carry the RFC #25 pointer. Right posture for scaffolding.

---

### Worth doing in this PR (both cheap, both outside RFC #25's scope)

**Nit — PR description is stale.** It says `SglangGenModel(GenModel)` / "overrides only `_alogprobs_impl`", but the rebased code extends `Model[str]` and overrides both impls. Please sync it. While there, add a one-line note that `_normalize_token_text` handles GPT-2 byte-level BPE only — SentencePiece `▁` (U+2581) passes through. It's documented in the docstring and contained in practice (CMMLU fails loud on missing option tokens rather than mis-scoring), so this is just to set expectations.

**Question — cross-engine default generation length differs.** In `sieval/core/models/sglang_gen_model.py:201`, when `max_tokens` is unset no `max_new_tokens` is sent, so sglang applies its own default (128) while the completions path applies the OpenAI default. Flipping `engine: vllm ↔ sglang` on a gen task with an unset `max_tokens` silently changes output length. RFC #25's IR doesn't define a canonical default, so this parity gap survives the eventual refactor — worth a one-line "set `max_tokens` explicitly for cross-engine parity" note. CMMLU is unaffected (`alogprobs`, `max_tokens=1`).

### Deliberately left as-is (RFC #25's territory)

The `isinstance`/`as_type` special-casing in `session.py` (`as_type` is slated for removal), the SDK-internal `_post` coupling, the sglang-native `request_params` shape, and the `Response`-shape nits (e.g. `response_model` hardcoded to the request model, the minor `usage`-when-`completion_tokens`-missing asymmetry) are all owned by RFC #25 — and the `Response` IR is co-versioned with #24's resume gate. Acting on them now would be work #25 discards, so I'd keep them behind the existing RFC pointers. I verified the `as_type` sglang special-case is correct transitional scaffolding (it preserves the sglang instance instead of downgrading to `/v1/completions`).

### Nice work
The radix-cache guard (`sglang_gen_model.py:289-307`) is the standout — it refuses to score silently-truncated echoed logprobs, covers the `cached_tokens>0`, count-mismatch, and missing-`prompt_tokens` cases, points the operator at `--disable-radix-cache`, and only fires on real cache hits (no false positives on cold prompts). Test suite is thorough and discriminating (echo ordering, normalization on both token and top-k keys, the real Qwen `\tB`-vs-` B` edge case, all guard branches).

jack-scitix-ai:
Thanks for the thorough approval. Both follow-ups done; left the RFC #25 items untouched as you suggested.

**1. Stale PR description / SentencePiece note.** Synced the description — it now says `SglangGenModel(Model[str])` and "overrides both `_agenerate_impl` and `_alogprobs_impl`". The `_normalize_token_text` docstring already carries the one-line limitation (GPT-2 byte-level markers only; SentencePiece `▁`/U+2581 passes through), so no code change there.

**2. Cross-engine default generation length.** Added a one-line note at the `_agenerate_impl` sampling construction: with `max_tokens` unset, no `max_new_tokens` is sent, so sglang applies its own default (128) while the vllm/OpenAI path uses the OpenAI default — set `max_tokens` explicitly for `engine: vllm ↔ sglang` output-length parity. No behavior change (RFC #25 owns the canonical default). CMMLU is unaffected (`alogprobs`, `max_tokens=1`).

Left as-is per your call — all behind the existing RFC #25 pointers: the `isinstance`/`as_type` special-casing, the SDK-internal `_post` coupling, the sglang-native `request_params` shape, and the `response_model`/`usage` Response-shape nits.

PR #23, opened 2026-07-01
feat(tasks): add clp eval mode + naming category

jack-scitix-ai:
## Type

## Summary

- Adds `EvalMode.CLP` ("clp") — next-token conditional log-prob over a fixed set of option tokens (single inference), distinct from `PPL` (full-sequence perplexity, n inferences); aligns with [OpenCompass's ppl/clp split](https://opencompass.readthedocs.io/en/latest/get_started/faq.html#:~:text=Similar%20to%20ppl,used%20in%20OpenCompass.).
- Wires `clp` through the seam: preflight naming regex + docs (naming table, `ppl`/`clp` distinction). No consumer task, no scoring helper, and no anomaly routing here — those land with the first `clp` task to keep the seam pure and avoid unreachable code.
- No consumer task and no scoring helper here — these ship with the first clp task to avoid uncalled code.
- No behavior change for existing tasks; `meta/index.json` is unchanged.
- Frozen-schema note: adding an enum value is additive (not a rename/removal/type-change) → no `schema_version` bump required.

### Automated

### Manual

- [x] Flipped cmmlu to `EvalMode.CLP` locally (reverted, not committed): the
  staged pipeline ran end-to-end via a mock GenModel and `cls.tags` synthesized
  `clp`.

### Required (all PRs)

### DIFF
--- a/.claude/rules/tasks.md
+++ b/.claude/rules/tasks.md
@@ -11,7 +11,9 @@ paths:
     - `_gen.py` → `model_type = "chat"`
     - `_base_gen.py` → `model_type = "gen"` (base model, uses GenModel)
     - `_ppl.py` → `model_type = "gen"` (perplexity, uses GenModel)
+    - `_clp.py` → `model_type = "gen"` (conditional next-token log-prob, uses GenModel)
 - Class naming: `<Benchmark><ShotType><Mode>Task` — words for shot count (`ZeroShot`, `FewShot`)
+- `ppl` vs `clp` distinction: see `sieval/tasks/CLAUDE.md`.
 
 ## Checklist for New Benchmarks
 
--- a/scripts/check_preflight.py
+++ b/scripts/check_preflight.py
@@ -44,7 +44,7 @@ _MD_RELATIVE_LINK = re.compile(r"\[([^\]]*)\]\(([^)]+)\)")
 _GH_NON_PERMANENT = re.compile(r"github\.com/[^/]+/[^/]+/blob/(main|master|develop)/")
 
 _TASK_FILE_PATTERN = re.compile(
-    r"^[a-z][a-z0-9_]*_(\d+|k)shot_(gen|base_gen|ppl|llmjudge_gen)\.py$"
+    r"^[a-z][a-z0-9_]*_(\d+|k)shot_(gen|base_gen|ppl|clp|llmjudge_gen)\.py$"
 )
 _DATASET_SUFFIX_PATTERN = re.compile(r"(Dataset|DatasetSample|CSVSample)$")
 
--- a/sieval/cli/task/commands.py
+++ b/sieval/cli/task/commands.py
@@ -37,7 +37,8 @@ def list_cmd(
         str | None, typer.Option("--domain", help="Filter by Level1Category.")
     ] = None,
     eval_mode: Annotated[
-        str | None, typer.Option("--eval-mode", help="Filter by eval mode (gen/ppl).")
+        str | None,
+        typer.Option("--eval-mode", help="Filter by eval mode (gen/ppl/clp)."),
     ] = None,
     data_dir: Annotated[
         str | None, typer.Option("--data-dir", help="Override data directory.")
--- a/sieval/core/tasks/meta.py
+++ b/sieval/core/tasks/meta.py
@@ -14,7 +14,7 @@ Not frozen (may change within schema_version=1):
     Python-side internals.
 
 Frozen enum values:
-    EvalMode: gen, ppl.
+    EvalMode: gen, ppl, clp.
     Status: stable, experimental, deprecated.
 
 AI-Generated Code - Claude Opus 4.6 (Anthropic)
@@ -40,6 +40,9 @@ class EvalMode(StrEnum):
     # PPL covers perplexity-based (logprobs).
     GEN = "gen"
     PPL = "ppl"
+    # CLP: next-token conditional log-prob over a fixed set of option tokens
+    # (single inference); vs PPL = full-sequence perplexity (n inferences).
+    CLP = "clp"
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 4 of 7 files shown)


ethan-scitix (commented):
**Comment — solid groundwork; recommend one cut + a doc trim before merge.**

Verified: the 187 touched tests pass; `EvalMode.CLP` is additive (the frozen-enum contract only gates rename/removal, and the docstring was updated); no exhaustive dispatch on `EvalMode` breaks; the OpenCompass anchor resolves and the citation is faithful. One process note: the branch is 1 commit behind `origin/main` — rebase before merge.

### Drop the `top_logprobs` anomaly change; keep this PR to enum + naming + docs

The new `top_empty` branch in `sieval/core/tasks/anomaly.py:443` can't fire in practice: `GenModel` normalizes empty top-k to `None` (`gen_model.py:261`, `top_token_logprobs or None`) and `ChatModel` never sets `top_logprobs`, so no backend emits `top_logprobs == []` — the test only reaches it by building that state by hand. It's also a no-op for the only current consumer: cmmlu is `ppl`-tagged and `GenModel`-backed, and its empty case is already caught by the existing `logprobs` branch.

More to the point, the branch is coupled to the `ModelOutput.top_logprobs: list[dict]` shape that RFC #25 is going to restructure — the same OpenAI-`/v1/completions` leak as the `SglangGenModel` discussion on #21. Suggest landing only the transport-independent part here and adding capability-aware `clp` empty-detection as part of the #25 implementation, against the real `Response` shape. (Keeping `applies_to` unchanged too — the routing can come with that work.)

### `Nit:` `ppl` vs `clp` docs — simplify and drop the pinned link

The `zh-cn/latest` OpenCompass link isn't version-stable, and the explanation is duplicated across `.claude/rules/tasks.md` and `sieval/tasks/CLAUDE.md`. Suggest one concise definition in `sieval/tasks/CLAUDE.md`:

```
### `ppl` vs `clp`
- `ppl` — pick the answer whose full `context + candidate` has the highest
  sequence likelihood; one inference per candidate, any answer length.
- `clp` — pick the answer by the next single token's log-prob over a fixed set
  of option tokens, read from the API's `top_logprobs` in one inference.
  Single-token / labeled-choice answers only; tokenizer-sensitive.
```

and a one-liner in `.claude/rules/tasks.md` that defers to it.

### cmmlu staying `ppl` — fine

Transitional; the `cmmlu_kshot_clp.py` preflight fixture already signals the planned migration.

jack-scitix-ai:
Thanks for the thorough review — all three actioned.

**Dropped the `top_logprobs` anomaly change.** You're right that the `top_empty` branch was unreachable: `GenModel` collapses empty top-k to `None` (`gen_model.py:261`) and `ChatModel` never sets `top_logprobs`, so the only way to hit it was hand-built test state. Reverted `detect_empty_infer_ppl` (and its tests) to the pre-clp baseline — `applies_to` stays `["ppl"]`, no new branch. Capability-aware `clp` empty-detection will land with the #25 implementation, against the real `Response` shape rather than the `list[dict]` leak.

**Simplified the `ppl` vs `clp` docs.** Single definition now lives in `sieval/tasks/CLAUDE.md` (used your wording), and the `zh-cn/latest` OpenCompass link is gone. `.claude/rules/tasks.md` is a one-liner that defers to it — no more duplication.

**cmmlu stays `ppl`** — agreed, transitional; the `cmmlu_kshot_clp.py` preflight fixture already marks the intended migration.

ethan-scitix (commented):
**Question / blocking:** The Summary and Test Plan describe anomaly routing that isn't in this diff. The body says `empty_infer_ppl` is now `applies_to=["ppl","clp"]` and flags empty `top_logprobs`, and the Manual section checks that off — but the commit touches no anomaly code:

- `anomaly.py:420` still reads `applies_to=["ppl"]`, and `detect_empty_infer_ppl` checks `logprobs`/`logprobs_tokens`, not `top_logprobs`.
- `_rule_applies(["ppl"], {"clp","zero_shot"})` → `False` (ran it); no rule has `clp` in `applies_to` or references `top_logprobs`.

So a `clp` task returning empty `top_logprobs` currently has no anomaly coverage, and that manual test couldn't have passed against this commit. Note `ppl` already has its detector routed with no consumer task, so `clp` is the odd one out here.

Can you confirm which you intend?

1. **Restore it** (looks dropped in a rebase): `applies_to=["ppl","clp"]` + a `top_logprobs`-emptiness branch, and update the detector's `description`/`tags` which still say "perplexity"/`logprobs=[]`. Preferred — matches the `ppl` sibling and makes the body true.
2. **Defer it** to the first clp task (like the consumer/scoring): then please drop that claim from the Summary and uncheck the Manual item.

Rest of the PR (enum, preflight regex, docs, tests) is correct and consistent — no other changes needed.

jack-scitix-ai:
Thanks — you're right the body was stale, and that's the real defect here. Fixed: dropped the anomaly-routing line from the Summary and trimmed the Manual item.

On direction: I've kept the code **deferred** rather than restoring, to stay consistent with the earlier review on this PR, which asked to drop the anomaly change and explicitly keep `applies_to` unchanged ("the routing can come with that work"). Two supporting points:

- The `top_logprobs`-emptiness branch specifically is unreachable with current backends — `GenModel` normalizes empty top-k to `None` (`gen_model.py:261`) and `ChatModel` never sets `top_logprobs` — so it'd be dead code today, coupled to the `ModelOutput.top_logprobs` shape RFC #25 will restructure.
- There's no `clp` consumer yet (cmmlu is still `ppl`), and when one lands via `GenModel.alogprobs` its empty case is already caught by the existing `logprobs`/`logprobs_tokens` branch the moment we route it — so today's coverage gap is nil.

Plan: add `clp` anomaly coverage with the first `clp` task / the #25 work, against the real `Response` shape. If you'd rather land routing now for parity with the `ppl` sibling, the minimal reachable change is just `applies_to += "clp"` (no `top_logprobs` branch) — happy to do that instead, but flagging the branch itself wouldn't fire. Which do you prefer?

PR #22, opened 2026-06-30
feat(tasks): add HMMT Feb 2025 and IMO-AnswerBench benchmarks

peter-scitix:
## Type

- feature — new benchmark, task, or capability

## Summary

- Add two final-answer math benchmarks, each aligned to its dataset author's reference implementation:
  - **HMMT Feb 2025** (30 problems) — MathArena-aligned (eth-sri/matharena @`a11194de` `configs/competitions/hmmt/hmmt_feb_2025.yaml`): boxed prompt + last-boxed extraction (reuses `community/matharena.py`), equivalence via `math-verify`. Same pipeline as `hmmt_feb_2026`.
  - **IMO-AnswerBench** (400 short-answer Olympiad problems; Google DeepMind IMO-Bench, arXiv 2511.01846) — grading **vendored from the official `answer_verification.py` @`66b014f1`** into `community/imo_bench.py`: `math-verify` equivalence with a normalized-string fallback (NOT an LLM judge). The reference harness is agentic (tool-call answer, "reason step by step"); for a non-agentic generative run we add a boxed answer format and reuse the matharena last-boxed extractor. Ships **`status="experimental"`** — a non-strict port (generative vs upstream's agentic tool-call); every deviation is enumerated in the task's `reference_impl.notes`.
- Datasets pinned to specific HF commits (`MathArena/hmmt_feb_2025@6fdc427`, `Hwilner/imo-answerbench@0258becb`). No `core/` changes; no new dependency (reuses the `[math]` extra).

## Related Issues

Refs the MathArena AIME/HMMT 2026 work (#16) — this PR builds on the same `community/matharena.py`.

### Automated

- [x] Lint/format clean (`ruff check && ruff format --check`)
- [x] Type check clean (`ty check`)
- [x] Unit tests pass (`pdm run pytest`) — verifier tests (integer / LaTeX-equiv / string fallback) + import-discipline (math-verify stays lazy)

### Manual

All evaluations run **through sieval** (`sieval eval`) against the scitix model API.

**HMMT Feb 2025 — grader+extraction parity** (regrading MathArena's published outputs `MathArena/hmmt_feb_2025_outputs` through this pipeline): reproduces their per-sample `correct` labels with **~100% agreement across 55 models** (exact reproduce for Gemini 3 Pro 97.5%, Grok 4 95.0%, Step 3.5 Flash 98.3%, Kimi K2.5 93.3%, …).

**HMMT Feb 2025 — live run** (sieval, avg@4):

[...]

PR #28, opened 2026-07-02
fix(imo-answerbench): normalize in extraction + verbatim grader, promote to stable

### LINKED ISSUE #26: [RFC]: IMO-AnswerBench — normalize in extraction + verbatim grader (fix format false-negatives, unblock stable)
### Motivation.

`imo_answer_bench_0shot_gen` (added in #22, currently `status="experimental"`) grades generative boxed answers, and it currently **scores genuinely-correct answers as wrong** — a correctness defect rooted in two coupled issues:

**1. Not aligned with upstream's *effective* grading.**
Upstream (OpenReward IMO-Bench `answer_verification.py` @`66b014f1`) is *agentic*: the model submits a clean answer via the `answer` tool, so upstream's `verify_math_answer` receives a normalized string and `math_verify` does the equivalence. Our generative port instead extracts the last `\boxed{}` (often verbose) and grades that. To compensate, #22 added a bespoke `verify_answer_gen` that normalizes + does **string-identity-only** atom/set matching — deliberately *not* calling `math_verify` per split-item, to dodge the `n=3k → 3` false positive.

**2. The string-identity restriction + incomplete normalization still under-count.**
Because `_atom_equiv` never calls `math_verify`, and `_normalize` doesn't strip set-braces `\{\}` / degree `^\circ` / all `\text{…}` qualifiers, answers that `math_verify` would trivially accept never reach it in a parseable form.

**Evidence** (independently re-graded from an internal DeepSeek-V4-Pro run's stored predictions):

- Current grader: **293/400 = 73.25%**; raw upstream `verify_math_answer` alone: 260/400 = 65.00%.
- Of the 94 "wrong & finished (non-truncated)" samples, **≥12 are mathematically equal but scored wrong**. Self-contained repro (only public functions):

```python
from sieval.community.imo_bench import verify_answer_gen
# (gold, pred) pairs currently graded WRONG that are mathematically equal:
pairs = [
    (r"$\sqrt{2}+1$",                 r"1+\sqrt{2}"),                       # commutative
    (r"$12^{10}$",                    r"2^{20} \cdot 3^{10}"),             # 12^10 = 2^20*3^10
    (r"847288609444",                 r"3^{25} + 1"),                      # 3^25 = 847288609443
    (r"$2\sqrt[3]{\frac{196}{13}}$",  r"\sqrt[3]{\frac{1568}{13}}"),       # 2^3*196 = 1568
    (r"$A(x)=\frac{1}{2}(x^2-x-4)$",  r"\frac{1}{2}x^2 - \frac{1}{2}x - 2"),
    (r"997008, 995026, 995018",       r"\{995018,\ 995026,\ 997008\}"),    # reordered set
    (r"1/2,1,2",                      r"\left\{\frac{1}{2}, 1, 2\right\}"),
    (r"$180 - 2\alpha$",              r"180^\circ - 2\alpha"),             # degree symbol
]
for g, p in pairs:
    assert verify_answer_gen(g, p) is False   # all currently FALSE (i.e. scored wrong)
```

- Probe: strip `$` / `\text{}` / `^\circ`, unwrap `\{\}`, then feed the **verbatim** `verify_math_answer` → all 12 flip to correct (`math_verify` natively handles commutativity, factoring, set-equality). Projected **~305/400 = 76.25%**, a lower bound (there are more that need semantic judgment).

The upshot: the "65% raw vs 73.25% bespoke" framing is misleading — *neither* is the faithful-to-upstream number. Upstream's agent submits clean, so it enjoys clean-answer grading; our true comparable number is higher (~76%+). The paper-match (best-non-Gemini 73.1%) is partly coincidental.

### Proposed Change.

Move answer normalization to the **correct layer** and keep the grader verbatim:

1. **Normalize in the extraction/postprocess layer** — reconstruct the clean answer an agent would submit: strip `$`, trailing `\text{for any/all/some …}` qualifiers, `^\circ`, `\left/\right`, normalize spacing; unwrap `\{…\}` into a form `math_verify` parses as a set.
2. **Grade with the verbatim upstream `verify_math_answer`** (unchanged, gold-first) and let `math_verify` do *all* equivalence (commutativity, factoring, set-equality).
3. **Delete** the bespoke `_atom_equiv` + hand-rolled set-matching — unnecessary once inputs are clean and `math_verify` handles sets.
4. Align parse/wrap with the HMMT sibling (`verify(parse(f"${gold}$"), parse(f"${pred}$"))`), whose symmetric `$`-wrapping already avoids some of these misses.

**Outcome:** grading becomes 100% verbatim upstream → removes the *material score dependence on bespoke grading logic* that keeps this task experimental. Once the acceptance criteria pass, **promote `status` from `experimental` to `stable`**. The only remaining bespoke logic is string normalization, which is legitimately a parsing concern (every gen task has one; matharena's `extract_answer` is itself heuristic).

**Acceptance criteria:**
- [ ] Full-400 re-run confirms **false positives stay ~0** — aggressive normalization must not collapse distinct answers (e.g. piecewise `\text{ if } x \ge 2` is meaningful; don't blindly strip).
- [ ] Score improves (target ≥76%) with **no regression** against the 293 currently-correct.
- [ ] `verify_math_answer` is untouched (verbatim upstream); all new logic lives in extraction/postprocess.
- [ ] `reference_impl.notes` updated: fix the 65%/73.25% framing, document normalization as parsing, flip `status` to `stable`.

**Out of scope (needs the real fidelity fix, not parsing):** genuinely non-math-verifiable answers — prose ("All positive integers"), `\ldots` infinite sets, variable-renamed functional families with quantifiers. Closing those requires reproducing upstream's agentic function-calling submission channel (the `answer` tool) or an LLM judge; tracked separately. This RFC closes only the parsing/normalization gap.

### Feedback Period.

~1 week.

### Any Other Things.

- Does **not** block #22 merging as `experimental` — current behavior is correct and conservative (under-count only, 0 false positives verified). This is the follow-up that takes it to `stable`.
- HMMT Feb 2025 (`hmmt_feb_2025_0shot_gen`) is unaffected: it already grades with verbatim `math_verify` and is correctly `stable`.

peter-scitix:
## Type

- fix — grading correctness / upstream alignment (also promotes IMO-AnswerBench experimental → stable)

## Summary

- IMO-AnswerBench (`#22`, `status="experimental"`) scored genuinely-correct generative answers as wrong. Root cause: the bespoke `verify_answer_gen` did string-identity-only atom/set matching and never let `math_verify` judge equivalence, and `_normalize` under-normalized.
- **Move normalization to the parsing/extraction layer + grade 100% with the verbatim upstream grader:**
  - Delete bespoke `verify_answer_gen` / `_atom_equiv` / hand-rolled set matching.
  - Add `normalize_answer` (parsing only): strip `$`, `^\circ`, `\left/\right`/spacing, a leading `f(x)=` prefix, and a **trailing** `\text{… for/where/such/with …}` qualifier; rewrite `\{…\}` to a set `math_verify` parses. Conservative — inline `\text` (e.g. piecewise `\text{ if } x≥2`) is kept.
  - Task grades with `verify_math_answer` on normalized inputs, **symmetric `$`-wrapping like the HMMT sibling**, so `math_verify` does all equivalence (commutativity / factoring / set-equality).
- **Promote `experimental` → `stable`**: grading is now verbatim upstream; only *parsing* is bespoke (legitimate — every gen task has one).

## Related Issues

Closes #26. Follow-up to #22.

### Automated

- [x] Unit tests pass (`pdm run pytest`) — 2095 passed; new tests cover `normalize_answer`, the recovered math-equivalences, no-false-positive, and prose-out-of-scope.

### Manual

Re-graded the stored DeepSeek-V4-Pro run (predictions unchanged; only the grader changed):

| Grader | Score |
|---|---|
| boxed answers, no normalization (raw `verify_math_answer`) | 260/400 = 65.0% |
| #22 bespoke `verify_answer_gen` | 293/400 = 73.25% |
| **this PR (normalize + verbatim `verify_math_answer`)** | **307/400 = 76.75%** |

- **+17 genuine math-equalities recovered, 0 false positives** (verified per-sample): commutativity (`√2+1` vs `1+√2`), factoring (`12^10` vs `2^20·3^10`), evaluation (`3^25+1`), reordered sets/tuples, degree symbol, functional-form expansion, `for any constant c` qualifiers.
- **3 regressions**, all **prose** the old bespoke matching happened to accept (`"1 and odd prime numbers"` vs `"1 and all odd prime numbers"`) — now left ungraded. These are **out of scope** (prose / infinite sets / quantified functional families need the agentic answer channel or an LLM judge, per `reference_impl.notes`), never false positives.

## Acceptance criteria (from #26)

- [x] Full-400 re-run: false positives stay 0.
- [x] Score improves (76.75% ≥ 76% target); +17 recovered vs 3 prose out-of-scope.
- [x] `verify_math_answer` untouched (verbatim upstream); all new logic in normalization (parsing/postprocess).
- [x] `reference_impl.notes` updated (65%/73.25%/76.75% framing, normalization = parsing); `status` → `stable`.

## Checklist

- [x] PR title conventional format
- [x] No internal paths / credentials in committed files
- [x] AI-generated code has the `AI-Generated Code - …` docstring line
- [x] No new upper-layer deps in `core/`
- [x] Deleted code verified — `verify_answer_gen` had no other call sites

### community/ Changes
- [x] `verify_math_answer` vendored verbatim (untouched); `normalize_answer` is sieval-added, documented as a parsing helper (not upstream).


ethan-scitix (commented):
**Question: source & data provenance before `stable`**

Two things to confirm — the grader copy from `EnvCommons/answer_verification.py` itself is faithful, this is just about what it's attributed to:

1. **Attribution.** `EnvCommons/IMO-Bench` isn't the official release — that's [`google-deepmind/superhuman/imobench`](https://github.com/google-deepmind/superhuman/tree/main/imobench), which ships only data CSVs (no `answer_verification.py`) and grades with an LLM `AnswerAutoGrader`, not `math_verify`. So the vendored grader is EnvCommons's own, not DeepMind's. Could we reword `reference_impl` so it doesn't read as "reproduces the official grader" — e.g. note it's a deterministic symbolic substitute for the official LLM grader, and point the URL at the fork we actually copied?

2. **Data version.** The official [README](https://github.com/google-deepmind/superhuman/blob/main/imobench/README.md) deprecated `answerbench.csv` on 02/12/2026 for `answerbench_v2.csv`, "to fix some problems that had ambiguous problem statements or incorrect answers." We load `hf:Hwilner/imo-answerbench` (v1). Can you confirm whether the mirror is still v1, and whether we should move to v2 before freezing as `stable`? If it's v1 with known-bad golds, the 76.75% would need a re-run on v2.

peter-scitix:
Both review points addressed in `03160c2` (CI green; full `sieval eval` on v2 completed).

**#1 — "reference impl seemingly not official."** Correct, and now fixed at the attribution level. `EnvCommons/IMO-Bench` (where the grader was vendored from) is **not** a Google DeepMind artifact — it's a third-party OpenReward re-implementation (org created 2026-01-24, `fork=false`, 0 stars). The authoritative source is **`google-deepmind/superhuman` (`imobench/`) + imobench.github.io + arXiv 2511.01846**, now named in `reference_impl.source` / `url` / notes. The divergence was evaluated and is low-risk: AnswerBench is scored **solely by `math_verify`** (paper/README: *"No LLM graders"*; the LLM autograders are only for GradingBench/ProofBench, which sieval does not use), and our vendored `verify_math_answer` / `parse_answer` is **byte-identical** to EnvCommons' `answer_verification.py`, which is itself **byte-stable between our pin `66b014f1` and upstream HEAD**. So the grader content equals the official scoring; only the attribution needed correcting. Grader logic unchanged.

**#2 — "dataset was updated; we're on the old one."** Correct, and it mattered. Our pin `hf:Hwilner/imo-answerbench@0258becb` (2025-11-08) is byte-identical to the official `answerbench.csv` **v1**, which GDM **deprecated on 2026-02-12** in favor of `answerbench_v2.csv` (*"fix … ambiguous problem statements or incorrect answers"*). v1→v2 diff: **11 `Short Answer`s changed** (~7 substantive, e.g. `combinatorics-005` …765→…764, `geometry-002` 4→1, `number_theory-048` inequality ≥→≤) + 26 problem statements reworded. Repinned the dataset to official **v2** via `url:` + sha256 at an immutable commit (`google-deepmind/superhuman@96fa6c4…/imobench/answerbench_v2.csv`), CSV loader mirroring `gpqa_diamond.py`, same column mapping. There is **no HF v2 mirror** (all HF mirrors are v1/pre-v2). Two v2 rows carry upstream spreadsheet artifacts (`imo-bench-algebra-036` gold → `Algebra`; `imo-bench-geometry-004` → Excel date serial `45752`); kept **verbatim** (faithful to the checksummed official CSV) and documented in `reference_impl.notes` — they grade wrong for ~any model (≤0.5% impact).

**v2 baseline — a real run, not an offline re-grade.** Ran a full `sieval eval` on v2 through the task's own code path: **DeepSeek-V4-Pro pass@1 = 317/400 = 79.25%** (`max_tokens=131072`, **0 truncation**, 0 unrecovered failures — 15 transient scitix *"upstream stream error"* / model-serving errors recovered via `--resume`). The previously-cited **76.75%** was a v1 **offline re-grade** over stored predictions and was **not** code-path-reproducible (the real code path gave 257 over the same stitched file); it is relabeled reference-only. `verify_math_answer` untouched; grading stays normalize-in-parsing + verbatim upstream `math_verify`.

ethan-scitix (commented):
**Question (fidelity framing, gates `stable`):** `sieval/community/imo_bench.py:9-14`, `sieval/tasks/imo_answer_bench_0shot_gen.py:85-91`

The notes say *"AnswerBench is scored solely by math_verify (paper: 'No LLM graders')… so the grader is equivalent to official."* Checked against the cited paper (arXiv 2511.01846 §2.3/§5.1): the official grader is an **LLM autograder (AnswerAutoGrader, Gemini 2.5 Pro)**, and the paper explicitly rejects SymPy-style symbolic checking as too narrow. `"No LLM graders"` / `"math_verify"` appear nowhere in the paper — that phrasing is the **EnvCommons README's**, describing its own re-impl. The official `imobench/` dir ships **data only, no grader code**.

So the accurate story (and the real motivation): official provides no runnable grader → a faithful port must adopt a referenceable impl → we vendor EnvCommons's deterministic `math_verify`, which also fits sieval's reproducibility contract (an LLM grader wouldn't). This is a **deliberate deviation** from the official LLM grader and is strictly more conservative — it can't grade the prose / infinite-set / functional-family cases (the "Out of scope" list), so sieval under-counts vs official.

This matters because the `experimental → stable` promotion is justified as *"grading is now verbatim upstream,"* which isn't accurate. Please reframe: `stable` freezes a **deterministic, reproducible** grading contract on the pinned v2 data (alignment target = EnvCommons `math_verify`, not the official autograder) — not fidelity to official. With that framing, `stable` is well-supported.

**Nit:** `"byte-identical to EnvCommons"` (`imo_bench.py:13`, notes `:89`) → behaviorally identical; imports were made lazy per sieval discipline.

---

Everything else verified good (no action):
- **79.25% reproduction confirmed and code-path-reproducible** — re-grading the stored v2 run's raw outputs through this PR's code gives 317/400 with 0 flips (sample-for-sample identical to the stored feedback).
- v2 CSV checksum matches a real download; data confirmed v2 (combinatorics-005 = `1431655764`); both corrupted-gold rows (`Algebra`, `45752`) present and graded wrong, as documented.
- Grader faithfully vendors `EnvCommons/IMO-Bench@66b014f1`; `normalize_answer` is conservative (trailing `\text{…for/where…}` only, inline `\text` kept); symmetric `$`-wrap matches the HMMT sibling.
- Deleted `verify_answer_gen`/`_atom_equiv`/`_split_top_level`/`_normalize` have no remaining callers; `ensure_dataset_dict` swap correct; `check_datasets` preflight passes; 12/12 tests pass.

peter-scitix:
@ethan-scitix — you're right, and thanks for catching it; I verified against the paper directly and corrected the framing in `32b5174` (rebased onto latest `main`).

**Confirmed:** arXiv 2511.01846 (§2.3/§5.1) grades AnswerBench with an **LLM autograder — AnswerAutoGrader (Gemini 2.5 Pro)** — and explicitly argues symbolic/SymPy-style matching is *"too narrow… reduces robustness"* (autograder validated at 98.9% vs humans; e.g. it accepts `all real numbers except -4` for `(−∞,−4)∪(−4,∞)`). The `"No LLM graders / math_verify"` phrasing is **EnvCommons's README** (its lines 28/36, describing its own OpenReward re-impl), not the paper; the official `imobench/` dir ships **data only, no grader code**. My earlier "equivalent to official" wording was wrong.

**Reframed** (`sieval/community/imo_bench.py` docstring, `imo_answer_bench_0shot_gen.py` module docstring + `reference_impl.notes`):
- The official grader is an LLM autograder; official ships no runnable grader.
- We vendor EnvCommons's deterministic `verify_math_answer` (`math_verify`) as a **deliberate, reproducible substitute** — an LLM grader wouldn't satisfy sieval's reproducibility contract.
- It is **strictly more conservative**: cannot grade prose / infinite-set / functional-family answers, so sieval **under-counts vs the official autograder** (that's the "Out of scope" list).
- **`stable` reframed accordingly**: it freezes a *deterministic, reproducible grading contract on the pinned v2 data* (alignment target = EnvCommons `math_verify`, **not** fidelity to the official autograder) — not "grading is now verbatim upstream."

**Nit fixed:** `byte-identical` → `behaviorally identical` (imports made lazy per sieval discipline).

No grader code/behavior change (79.25% v2 baseline unchanged). Rebased onto `main@8c95f37`; CI re-running.

PR #29, opened 2026-07-02
feat(gsm8k): add DeepSeek-Math-aligned 0-shot chat-model task

jack-scitix-ai:
## Type

- feature — new benchmark, task, or capability

## Summary

- Add `gsm8k_0shot_gen` (`GSM8KZeroShotGenTask`): a 0-shot chat GSM8K task aligned with DeepSeek-Math's `gsm8k-test` eval path (pinned commit  `b8b0f8ce`). Prompt = question + "Please reason step by step, and put your
  final answer within \boxed{}."; the chat template is applied by the serving backend; greedy decoding `temperature=0, top_p=1.0, max_tokens=1024`.
- Vendor DeepSeek-Math's answer extraction + scoring byte-for-byte into `sieval/community/deepseek_math.py` (identical to the sieval-math repo's copy, so the two PRs share the file conflict-free). The task uses `extract_answer(exhaust=False)` (= DeepSeek `extract_last_single_answer`) and `is_correct`/`math_equal` (= `eval_last_single_answer`).
- Gold derived from `openai/gsm8k` like DeepSeek's `process_gsm8k_test` (`answer.split("####")[-1]`, commas removed).
- Reference: DeepSeek-Math evaluation (https://github.com/deepseek-ai/DeepSeek-Math/tree/b8b0f8ce093d80bf8e9a641e44142f06d092c305/evaluation); target score from the DeepSeek LLM report (arXiv:2401.02954), Table 6.

### Automated

- [x] Lint/format clean (`ruff check && ruff format --check`)
- [x] Type check clean (`ty check` or `mypy --strict`)
- [x] Unit tests pass (`pdm run pytest`) 

### Manual
  Model: DeepSeek-LLM-7B-Chat (sglang, bf16, greedy temperature=0, top_p=1.0, max_tokens=1024), full 1319-sample GSM8K test split, 0 failures.
  | Model | Expected | Actual | Diff |
  |---|---|---|---|
  | DeepSeek-LLM-7B-Chat | 63.0 (DeepSeek LLM report, Table 6) | 63.31 | +0.31 (+0.31%) |

### Required (all PRs)

- [x] PR title follows conventional format (`type(scope): description`)
- [x] No internal paths, credentials, or personal info in committed files
- [x] AI-generated code has `AI-Generated Code - <model> (<provider>)` in module docstring
- [x] No new upper-layer dependencies added to `core/`
- [x] Deleted code verified — no remaining call sites depend on it

### If: New or Modified Benchmark

[...]

PR #31, opened 2026-07-06
feat(hendrycks_math): add full MATH dataset and DeepSeek-Math-aligned base-model task

jack-scitix-ai:
## Type

## Summary

- Add `hendrycks_math_kshot_base_gen` (`HendrycksMathFewShotBaseGenTask`): full Hendrycks MATH (5,000-test) base-model task — a strict port of DeepSeek-Math's `math-cot-test` (Minerva 4-shot CoT prompt + DeepSeek's list-valued extraction + `eval_math`/`math_equal`). Ref: https://github.com/deepseek-ai/DeepSeek-Math/tree/b8b0f8ce093d80bf8e9a641e44142f06d092c305/evaluation
- Add `hendrycks_math` dataset (EleutherAI/hendrycks_math, revision-pinned).
- Extend the existing `sieval/community/deepseek_math.py` (landed in #29) with the MATH-specific helpers (`few_shot_prompt`/`format_prompt`/`STOP_WORDS`, `extract_math_answer`, `extract_math_few_shot_cot_answer`, `eval_math`) — vendored verbatim from the same pinned commit; verified 0/5000 mismatch vs upstream modules.

### Automated

### Manual

- [x] Qwen2.5-72B-Base, full Hendrycks MATH (5,000 test), greedy, deterministic seed=0, `max_gen_toks=1024`, `tp_size=2`:

  | Model | Expected| Actual | Diff |
  | --- | --- | --- | --- |
  | Qwen2.5-72B-Base | 54.4 (DeepSeek-V3 report) | 61.74 | +7.34(+7.34%) |

61.74 is the stored 1024/tp2 generations re-scored under the fixed `parse_latex` (antlr 4.11.0) — the generations are unchanged, only the scorer's last layer. The run's stored `report.json` shows 60.60 because it was scored before the antlr fix; a fresh run with the merged deps reproduces 61.74. The result is above the report's 54.4; most likely a harness difference (the report's number is DeepSeek-V3's internal eval, not the open-source `math-cot-test` this ports — verified byte-identical, 0/5000 mismatch).

Reproducibility: `deterministic: true` pins the seed but not batch-invariant kernels, so `tp_size` shifts the score ~1pp under identical decoding — fix `tp_size` when reproducing.

### Required (all PRs)

### If: community/ Changes

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "dev", "drop", "ifbench", "ifeval", "math", "t-eval", "test"]
 strategy = ["inherit_metadata"]
 lock_version = "4.5.0"
-content_hash = "sha256:d2e318510436de5f2157f0c79e01dedbbaa05f5388f4fd0867ac7da504c31612"
+content_hash = "sha256:aadbe110ed98781ee5c51ae38bd40780304ed26da3c371a0fd3f5ff1fd1256ef"
 
 [[metadata.targets]]
 requires_python = ">=3.12,<3.15"
@@ -928,7 +928,7 @@ files = [
 
 [[package]]
 name = "latex2sympy2-extended"
-version = "1.11.0"
+version = "1.10.2"
 requires_python = ">=3.10"
 summary = "Convert LaTeX math to SymPy expressions"
 groups = ["math"]
@@ -937,24 +937,24 @@ dependencies = [
     "sympy",
 ]
 files = [
-    {file = "latex2sympy2_extended-1.11.0-py3-none-any.whl", hash = "sha256:aebb77d52ce269e25028e4bea89ddb14d242ba36bcf7b636496fb5fd9728d234"},
-    {file = "latex2sympy2_extended-1.11.0.tar.gz", hash = "sha256:9695657c81b50abba2636638638618db59f4663ed2a4a12d62cef74a40e28fec"},
+    {file = "latex2sympy2_extended-1.10.2-py3-none-any.whl", hash = "sha256:f910442c5b02a466c1046f47d05cc5285181068b882399281f30102715337fb7"},
+    {file = "latex2sympy2_extended-1.10.2.tar.gz", hash = "sha256:41a517ffcc5a140e910a7d1646ce6ff440817e5f9d48fc8279d88bd0925bc389"},
 ]
 
 [[package]]
 name = "latex2sympy2-extended"
-version = "1.11.0"
+version = "1.10.2"
 extras = ["antlr4_11_0"]
 requires_python = ">=3.10"
 summary = "Convert LaTeX math to SymPy expressions"
 groups = ["math"]
 dependencies = [
     "antlr4-python3-runtime==4.11.0",
-    "latex2sympy2-extended==1.11.0",
+    "latex2sympy2-extended==1.10.2",
 ]
 files = [
-    {file = "latex2sympy2_extended-1.11.0-py3-none-any.whl", hash = "sha256:aebb77d52ce269e25028e4bea89ddb14d242ba36bcf7b636496fb5fd9728d234"},
-    {file = "latex2sympy2_extended-1.11.0.tar.gz", hash = "sha256:9695657c81b50abba2636638638618db59f4663ed2a4a12d62cef74a40e28fec"},
+    {file = "latex2sympy2_extended-1.10.2-py3-none-any.whl", hash = "sha256:f910442c5b02a466c1046f47d05cc5285181068b882399281f30102715337fb7"},
+    {file = "latex2sympy2_extended-1.10.2.tar.gz", hash = "sha256:41a517ffcc5a140e910a7d1646ce6ff440817e5f9d48fc8279d88bd0925bc389"},
 ]
 
 [[package]]
@@ -1195,32 +1195,32 @@ files = [
 
 [[package]]
 name = "math-verify"
-version = "0.9.0"
+version = "0.8.0"
 requires_python = ">=3.10"
 summary = "HuggingFace library for verifying mathematical answers"
 groups = ["math"]
 dependencies = [
-    "latex2sympy2-extended==1.11.0",
+    "latex2sympy2-extended==1.10.2",
 ]
 files = [
-    {file = "math_verify-0.9.0-py3-none-any.whl", hash = "sha256:3703e7c4885354027fa84409d762a596a2906d1fd4deb78361876bd905a76194"},
-    {file = "math_verify-0.9.0.tar.gz", hash = "sha256:45ac6c61344ba056b9e99a660a4bc8d044ed408f730aed68c60435aa5eec4645"},
+    {file = "math_verify-0.8.0-py3-none-any.whl", hash = "sha256:31ca651296d817a9bb3fd58ca1fd0d192dcea709b1e5ecf2d0a4514c16f89087"},
+    {file = "math_verify-0.8.0.tar.gz", hash = "sha256:3295e0adb94bfe553ff6e3189c44f1916a85aa24ab5d1900f2086a706e28f7c4"},
 ]
 
 [[package]]
 name = "math-verify"
-version = "0.9.0"
+version = "0.8.0"
 extras = ["antlr4-11-0"]
 requires_python = ">=3.10"
 summary = "HuggingFace library for verifying mathematical answers"
 groups = ["math"]
 dependencies = [
     "latex2sympy2-extended[antlr4_11_0]",
-    "math-verify==0.9.0",
+    "math-verify==0.8.0",
 ]
 files = [
-    {file = "math_verify-0.9.0-py3-none-any.whl", hash = "sha256:3703e7c4885354027fa84409d762a596a2906d1fd4deb78361876bd905a76194"},
-    {file = "math_verify-0.9.0.tar.gz", hash = "sha256:45ac6c61344ba056b9e99a660a4bc8d044ed408f730aed68c60435aa5eec4645"},
+    {file = "math_verify-0.8.0-py3-none-any.whl", hash = "sha256:31ca651296d817a9bb3fd58ca1fd0d192dcea709b1e5ecf2d0a4514c16f89087"},
+    {file = "math_verify-0.8.0.tar.gz", hash = "sha256:3295e0adb94bfe553ff6e3189c44f1916a85aa24ab5d1900f2086a706e28f7c4"},
 ]
 
 [[package]]
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "dev", "drop", "ifbench", "ifeval", "math", "t-eval", "test"]
 strategy = ["inherit_metadata"]
 lock_version = "4.5.0"
-content_hash = "sha256:cd1913d87a36d8c8241734a702065412185ca7c3e9949bc22206e4455f272c86"
+content_hash = "sha256:d2e318510436de5f2157f0c79e01dedbbaa05f5388f4fd0867ac7da504c31612"
 
 [[metadata.targets]]
 requires_python = ">=3.12,<3.15"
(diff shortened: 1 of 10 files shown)


ethan-scitix (commented):
Port verified byte-for-byte against pinned `b8b0f8ce` (prompt, extractors, `eval_math`, wiring, dataset pin all faithful). Not blocking — two things to confirm + one test nit.

**1. `sieval/tasks/hendrycks_math_kshot_base_gen.py:37` — `Question:`** documented repro decoding is `max_gen_toks=2048`, but the ported harness (`run_cot_eval.py`) uses `max_tokens=1024` for all cot-test tasks (HF path 512), and the GSM8K sibling documents 1024. As the only undocumented deviation in a "strict port," and since a doubled budget can only raise MATH accuracy, it plausibly feeds the `+5.28` gap. Intentional (MATH chains run longer)? If so, please add it to the Deviation list; otherwise drop to `1024`.

**2. `sieval/tasks/hendrycks_math_kshot_base_gen.py:126` — `Question:`** denominator is `len(finals)`, so fails are excluded. The other DeepSeek-Math task (`gsm8k_0shot_gen.py:143`) uses `finals + fails` and documents "a failure counts as wrong (DeepSeek full-set accuracy)." Consistent with `gsm8k_kshot_base_gen` but diverges from that sibling. No impact when `fails=0`; flagging as family-convention drift. OK to reconcile in a separate follow-up across the `_kshot_base_gen` tasks rather than here — just confirm the intent.

**3. `tests/unit/tasks/test_hendrycks_math_kshot_base_gen.py:146` — `Nit:`** `test_report_denominator_excludes_fails` passes `[]` for fails, so `len(finals)` vs `finals+fails` both give 50.0 — it can't actually verify exclusion. Pass one failed context and assert score stays 50.0 (not 33.3).

jack-scitix-ai:
Thanks for the review — all three addressed. Rebased onto latest `main`; conflicts were only the 3 auto-generated files (`meta/index.json`, `tasks/__init__.pyi`, `datasets/__init__.pyi`), resolved by regeneration.

**1. `max_gen_toks`** — aligned to DeepSeek's value rather than documenting a deviation. `run_cot_eval.py:80` uses `max_tokens=1024` (vLLM path; HF path is 512), so the docstring repro line and the run config are now `1024`.

Re-ran at 1024: **60.6%** — vs **60.64%** at 2048 on the same `tp_size`(tp_size 2), i.e. essentially unchanged (only 12/5000 samples flip). So the doubled budget was *not* feeding the gap. The ~1pp we'd seen came from **`tp_size` non-determinism**, not `max_tokens`: 2048-tp2=60.64 vs 2048-tp4=59.68 flips 324/5000 samples, every one with a *different* greedy generation (`deterministic: true` pins the seed but not batch-invariant kernels across tp). Net: the gap vs the report's 54.4 isn't `max_tokens` — most likely the report's internal harness/snapshot. (Score in the PR body updated to 60.6.)

**2. Denominator** — reconciled here rather than deferring. Since this is a DeepSeek-Math port, I matched the DeepSeek-Math sibling `gsm8k_0shot_gen`: `report()` now divides by `len(finals) + len(fails)` — a pipeline failure counts as **wrong** (DeepSeek full-set accuracy), with a comment citing the sibling. The `len(finals)` form is the lm-eval family (`gsm8k_kshot_base_gen`), which this task isn't. No behavioral change when `fails=0`.

**3. Test nit** — fixed: `test_report_denominator_excludes_fails` → `test_report_counts_fails_as_wrong`. It now passes one *failed* context and asserts the score is `100/3` (33.3), not 50.0 — so it actually distinguishes `finals + fails` from `len(finals)` (the old `[]`-fails version couldn't). Verified passing.

ethan-scitix (commented):
Re-review of the current commit — earlier 3 comments resolved (1024 tokens, `finals+fails` denominator, test fixed). Reproduced **60.60%** from the run shards. Four items left; only #1 is worth acting on before merge.

**1. `Bug:` avoidable `parse_latex` break — pin antlr to 4.11.0.**
`symbolic_equal`'s `parse_latex` raises in-env (sympy 1.14 needs antlr `4.11.x`, env resolved `4.13.2`), so the vendored DeepSeek matcher silently falls back to `parse_expr`/string. This is a version gap, not a conflict: `latex2sympy2-extended` accepts antlr `>=4.9.3,<=4.13.2` and ships an `antlr4_11_0` extra, so `4.11.0` satisfies both sympy and math_verify (verified in an isolated venv — `parse_latex` recovers `(x+1)^2 == x^2+2x+1`). Bounded impact (~1.2pp of currently-wrong samples are truly-equivalent fractions/decimals a working `parse_latex` would recover), but it's a fidelity break in a "strict port." Fix: `math-verify[antlr4-11-0]` in the `[math]` extra + re-lock → `parse_latex` works, the Deviation note can be deleted, and `status="stable"` is unambiguous. NB the current note ("conflicting with the 4.9.x pinned by math_verify") is inaccurate — math_verify pins a *range* and the env resolved 4.13.2.

**2. `Nit:` `sieval/datasets/hendrycks_math.py` — flag the data-source swap** like `gsm8k_0shot_gen` does: DeepSeek's `math-cot-test` loads bundled `datasets/math/test.jsonl`, this loads `EleutherAI/hendrycks_math`. Verified equivalent (test=5000, same 7 subjects) — one-line note for parity.

**3. `Nit:` `sieval/tasks/hendrycks_math_kshot_base_gen.py:138` — `exact_match` key** diverges from the sibling's `accuracy` and is a slight misnomer (it's `math_equal` equivalence, not literal match). Rename to `accuracy`, or drop it (aime/hmmt/imo/math_500 return only `{score, fails}`).

**4. `Nit:` `sieval/tasks/hendrycks_math_kshot_base_gen.py:114` — reference extraction has no guard**, unlike upstream `process_math_test` (`try/except: return` drops the sample); here it counts as fail-wrong. Immaterial in practice (extractors are robust) — worth a one-line note.

jack-scitix-ai:
Thanks for the review — all four addressed:

**1. `Bug` — parse_latex / antlr:** Fixed. Added `math-verify[antlr4-11-0]` to the `[math]` extra + re-locked → `antlr4-python3-runtime` pins to `4.11.0` and `parse_latex` now works (verified `(x+1)^2 == x^2+2x+1`). Re-scored the stored 1024/tp2 shards with the working matcher: **60.60% → 61.74%** (+57 samples — mostly base-N subscript forms like `3032_5 ≡ 3032_{5}` that only the LaTeX parser reconciles). Deleted the parse_latex deviation notes (task docstring, community docstring, `reference_impl.notes`). You're right the old "conflicting with 4.9.x" wording was inaccurate — the env had resolved `4.13.2`, a version gap, not a conflict. NB the extra also bumps `math-verify 0.8→0.9` and `latex2sympy2-extended 1.10→1.11`; `math_verify` re-verified working, so the shared instruct task's scorer is unaffected.

**2. `Nit` — data-source swap:** Done. Added a one-line note in `hendrycks_math.py`: DeepSeek's `math-cot-test` loads bundled `datasets/math/test.jsonl`; this loads `EleutherAI/hendrycks_math` — verified equivalent (5,000 test, same 7 subjects).

**3. `Nit` — `exact_match` key:** Done. Renamed `exact_match` → `accuracy` in `report()` to match the `gsm8k_0shot_gen` sibling (`{score, fails, accuracy}`) and drop the misnomer; tests updated.

**4. `Nit` — reference-extraction guard:** Done. Added a one-line docstring note: unlike upstream `process_math_test`'s `try/except: return` (which drops the sample), a failed extraction here counts as fail-wrong. Immaterial in practice — all 5,000 test rows carry a `\boxed` answer (0 extraction failures).

ethan-scitix (changes requested):
**Re-review of `0cf9c61b`.** The port itself is faithful — verified byte-for-byte against pinned `b8b0f8ce`: Minerva 4-shot prompt + stop words, the extractors, `eval_math`/`is_correct`, `max_tokens=1024` (vLLM path), the dataset pin + 7-subject layout, and the `math-cot-test` wiring (`MinervaMathPrompt`/`extract_math_few_shot_cot_answer`/`eval_math`). All 7 prior comments are resolved, 17 unit tests pass, preflight is clean, and `parse_latex` is confirmed working under antlr 4.11.0. One new blocking finding → **request changes**.

**Prior comments — all 7 resolved.** R1: `max_gen_toks`→1024, denominator `finals+fails` (matches the `gsm8k_0shot_gen` sibling), test now discriminating (`100/3`). R2: antlr 4.11.0 pin (see below), data-source note, `exact_match`→`accuracy`, reference-extraction deviation documented.

---

### `Bug:` the antlr pin silently bumps the shared math scorer (`math-verify 0.8→0.9`), changing sibling task scores

The PR reply states this bump leaves "the shared instruct task's scorer unaffected." Replaying real stored predictions through both versions disproves that:

| run | acc@0.8 | acc@0.9 | Δ |
|---|---|---|---|
| MATH-500 (500) | 65.80% | 66.00% | +0.20pp |
| full MATH (5000) | 60.78% | 60.94% | +0.16pp |

Root cause: `X%` parsing changed (0.8 → `X/100`; 0.9 → `X`), so e.g. `gold=20 / pred=20%` flips `False→True`. Every flip is one-directional — 0.9 is *more lenient*, accepting `20%` for a gold of `20`, which for a strict eval is a **false positive**, not an improvement. 8 tasks call `math_verify.verify` directly; `math_500_0shot_gen` / `imo_answer_bench` are affected (AIME/HMMT are integers → unaffected; `hendrycks_math` here uses the DeepSeek scorer, so its 60.6 is safe either way).

The bump is also **avoidable and unintended** — a side effect of the open `>=0.8.0` floor re-resolving to newest, **not** required by the fix. The port's `parse_latex` is sympy's, and `math-verify[antlr4-11-0]==0.8.0` already pulls `antlr4-python3-runtime==4.11.0` (verified in a clean venv), which fixes `parse_latex` with the scorer left untouched.

Please pin `math-verify[antlr4-11-0]==0.8.0` (+ re-lock) so the dependency change stays load-bearing-only. If you'd rather keep 0.9.0, it needs evidence it's holistically better (e.g. 0.9's `%` semantics is more faithful to MATH gold conventions) **plus** re-baselined reference numbers for the affected tasks — the default should be to keep the validated scorer.

### `Nit:` `sieval/community/deepseek_math.py:6` docstring is stale

"Faithful port — trimmed to exactly what the GSM8K 0-shot task consumes" no longer holds now that the module also serves the MATH task (correctly described at `:22-27`). Reword to name both consumers.

jack-scitix-ai:
**Bug (shared scorer bump):** Fixed — pinned `math-verify[antlr4-11-0]==0.8.0` + re-locked. You're right, my earlier "unaffected" claim was wrong: the `>=0.8.0` floor re-resolved to `0.9.0`. `==0.8.0` still pulls `antlr4-python3-runtime==4.11.0` (so `parse_latex` works), but keeps the validated scorer — the lock now differs from `main` **only** by `antlr 4.13.2→4.11.0` (`math-verify` back to 0.8.0, `latex2sympy2-extended` back to 1.10.2). Confirmed the 0.8.0 `%` semantics are restored: `verify(gold=20, pred=20%)` → `False`. `hendrycks_math`'s own score is unchanged (it uses the DeepSeek `math_equal`, not `math_verify`).

**Nit (stale docstring):** Fixed — reworded `deepseek_math.py:6` from "trimmed to exactly what the GSM8K 0-shot task consumes" to name both the GSM8K and MATH consumers, consistent with the per-consumer breakdown at `:22-27`.

**On the reported number (60.6 vs 61.74):** the run's stored `report.json` says **60.60** — that's what you reproduced from the shards, and it was scored *before* the antlr 4.11 fix (broken `parse_latex`). Re-scoring the same generations under the now-working `parse_latex` gives **61.74** (+57 samples, mostly base-N subscript forms like `3032_5 ≡ 3032_{5}`); the generations are identical, only the scorer's last layer changed. I've updated the PR body to 61.74 with that provenance. Note `--resume` can't refresh this — all 5,000 samples are FINAL, so resume hydrates them as-done and re-emits 60.60 rather than re-running feedback; only a fresh run (or offline re-score) reflects the fix. And this is independent of the `math-verify==0.8.0` pin — `hendrycks_math` scores via the DeepSeek `math_equal`/`parse_latex`, not `math_verify`.

ethan-scitix (approved):

PR #32, opened 2026-07-07
refactor(cmmlu): reclassify cmmlu task under clp eval mode

jack-scitix-ai:
## Type

## Summary

- The CMMLU task's mechanic is **clp** (single inference, next-token log-prob argmax over the fixed A/B/C/D option tokens via `top_logprobs`), but it was mislabeled as `_base_gen` + `EvalMode.PPL`. #23 introduced the `clp` eval mode/naming category; this reclassifies CMMLU into it.
- Rename `cmmlu_kshot_base_gen` → `cmmlu_kshot_clp` (file, task `name`, class `CMMLUFewShotBaseGenTask` → `CMMLUFewShotClpTask`), `eval_mode` `PPL` → `CLP`, `display_name` → "CMMLU (few-shot, base CLP)". `model_type` stays `gen`. Regenerated `sieval/meta/index.json` and `sieval/tasks/__init__.pyi`.
- Scoring pipeline (preprocess/infer/postprocess/report) is **byte-identical** → scores unchanged, no re-run required.
- Promote `choice_scores_from_top_logprobs` into `core/utils/ppl.py` (with test) and switch CMMLU to import it, dropping the task-local copy. This is the shared helper #33 (MMLU-clp) also uses — #33 drops its duplicate on rebase.
- No anomaly-rule change. An earlier revision added `detect_empty_infer_clp`, but it can't fire: empty `top_logprobs` raises in `postprocess` → sample FAILED, and detection is FINAL-gated (`anomaly.py:180`), so the raise is already the detector. Dropped it. The clp false-fire in `detect_truncated_output` (`max_tokens=1` → `finish_reason="length"`) is a pre-existing shared-rule issue, tracked for #33.

### Automated

### Manual

- [x] Full `python scripts/check_preflight.py` passes — `check_tasks` (naming, 24 exports), `check_meta_index_sync`, `check_imports` all green.

### Required (all PRs)

### If: Breaking Change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,38 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [0.6.0] - 2026-07-06
+
+### Added
+
+- New benchmark tasks & datasets:
+    - GSM8K — 8-shot base-model task (#1) and DeepSeek-Math-aligned 0-shot chat-model task (#29).
+    - TheoremQA — k-shot base-model task (#3).
+    - HumanEval — 0-shot base-model task (#6).
+    - CMMLU — few-shot base-model task (#10).
+    - MBPP — few-shot base-model task (#12).
+    - IFBench — few-shot base-model task (#13).
+    - LiveCodeBench — few-shot base-model code-generation task (#14).
+    - OpenBookQA — k-shot generative task (#19).
+    - AIME 2026 and HMMT Feb 2026 — MathArena-aligned (#16); HMMT Feb 2025 and IMO-AnswerBench (#22).
+    - CLP — eval mode and naming category (#23).
+- `SglangGenModel` — echoed-input logprobs via the SGLang `/generate` endpoint (#21).
+- `stratified_sample` dataset op (#7).
+- Dataset source integrity: pinned HF revisions and checksummed URL datasets, enforced in preflight (#8).
+- `--resume` now tolerates throughput-only (scheduling) config diffs (#4).
+- GitHub Actions CI pipeline and import-time dependency hardening (#2).
+
+### Fixed
+
+- IMO-AnswerBench: normalize during answer extraction, verbatim grader; promoted to stable (#28).
+- Pass the gold answer first to `math_verify.verify` (#18, #20).
+- Dataset integrity check compares on-the-wire bytes so gzipped responses are not falsely flagged (#17).
+
+### Changed
+
+- Renamed the `select` dataset op to `slice` (#7).
+- Sanitize CI check now detects hardcoded absolute paths and scans only tracked files (#5).
+
 ## [0.5.0] - 2026-05-06
 
 Initial public release.
@@ -63,7 +95,7 @@ Mainstream benchmarks registered in `sieval/meta/index.json`:
 
 ### Registries
 
-- `sieval/meta/index.json` (schema v0.1) — task / dataset registry, auto-generated via `scripts/sync_meta_index.py`.
+- `sieval/meta/index.json` (schema v1) — task / dataset registry, auto-generated via `scripts/sync_meta_index.py`.
 - `@sieval_task` / `@sieval_dataset` decorators with `TaskMeta` / `DatasetMeta` schemas.
 - AST-based lazy discovery in `sieval.tasks` / `sieval.datasets`.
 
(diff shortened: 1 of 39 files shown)


ethan-scitix (commented):
**Approve once coordinated with #33.** The PPL→CLP reclassification is a real correctness fix (the task does a single-call `top_logprobs` argmax over A/B/C/D — that's `clp`, not `ppl`); tests pass, preflight clean, reference URL pinned + 200. But #32 and #33 patch the same surface differently and must be reconciled first.

**Coordinate with #33 — same clp/anomaly surface, two designs.** Both touch `core/tasks/anomaly.py` and a `choice_scores_from_top_logprobs` helper; they'll hard-conflict. Where they differ, #32 is mostly the one to keep:

1. **Empty-check (correctness) — #32 is correct, #33 isn't.** This PR's `detect_empty_infer_clp` uses `not result.top_logprobs` (clp-tag-routed), catching both `None` and `[]`. #33 folds clp into `detect_empty_infer_ppl` gated on `top_logprobs is not None`, which misses `None` — and `GenModel` coalesces an empty top-k to `None` (`gen_model.py:261`), i.e. the actual production-empty result. Keep #32's semantics.

2. **Helper duplication — adopt #33's shared one.** #32 hand-rolls `_choice_scores_from_top_logprobs` in the task; #33 adds an identical `core/utils/ppl.choice_scores_from_top_logprobs`. Import the shared one instead of the local copy.

**Two problems neither PR addresses:**

- **`detect_truncated_output` false-fires on every clp sample** — it defaults to `all_tasks`, and clp infers with `max_tokens=1` → `finish_reason="length"` (`gen_model.py:195,226`). Mass "truncation" noise (severity=info, but drowns real signal); needs a clp/single-token exclusion.
- **`detect_empty_infer_clp` can't fire here** — empty `top_logprobs` → `postprocess` raises → sample FAILED → never reaches the FINAL-gated `detect()`. The `raise` is already the detector; either drop the rule or detect earlier than FINAL.

**Local nits:** rule `rationale` conflates empty-vs-dropped-tokens (`anomaly.py:443-446`); docstring reflow artifact (`cmmlu_kshot_clp.py:9-11`).

jack-scitix-ai:
Thanks — reworked #32. Mapping to your points:

**"Empty-check … #32 is correct" vs. "detect_empty_infer_clp can't fire here":**
these two pull opposite ways, and the second wins — a rule that can't fire, however correct its predicate, is dead code. Confirmed: empty `top_logprobs` raises in `postprocess` → sample FAILED, and `detect()` is FINAL-gated (`anomaly.py:180`), so the raise is already the detector. Dropped `detect_empty_infer_clp` + its test; `anomaly.py` / `test_anomaly.py` are back to `main`. The correct `not top_logprobs` semantics still live in the shared helper (they drive `postprocess`'s `all_present` → raise), just not as an anomaly rule — which also moots the #32-vs-#33 empty-check comparison and the `rationale` nit.

**"Helper duplication — adopt #33's shared one":** promoted `choice_scores_from_top_logprobs` into `core/utils/ppl.py` (with test), byte-identical to #33's, and switched CMMLU to import it (dropped the private copy). On rebase #33 deletes its duplicate function + test.

**Docstring reflow artifact:** fixed.

**"detect_truncated_output false-fires on every clp sample":** real, and pre-existing (fires on any `max_tokens=1` logprob task). Since it's a shared `all_tasks` rule, I'm fixing it once in #33 (`applies_to=["gen"]`) rather than re-touching `anomaly.py` here and reintroducing the conflict.

Overall: #32 no longer touches `anomaly.py`, so the conflict with #33 is gone, and the shared helper lands here for #33 to import.

ethan-scitix:
Pushed a small maintainer follow-up (`65383e20`) to close out the last two items so we can land this:

- **Test discriminating power** (`test_ppl.py`): the `choice_scores_from_top_logprobs` tests only ever placed option labels (A/B/C/D) in the top-k, so the `if label in scores` guard — the function's actual job on a real top-k, which is mostly *non*-option tokens — was never exercised. A mutant dropping that guard survives the whole suite yet breaks on real input (distractors leak into `scores` and flip `all_present` to `False`). Added `test_non_option_tokens_are_ignored`; confirmed it fails on that mutant.
- **Docstring reflow** (`cmmlu_kshot_clp.py`): dropped the stray short line left by an earlier edit. No behavior change.

`detect_truncated_output`'s clp false-fire stays deferred to #33 (`applies_to=["gen"]`), as agreed — this PR still doesn't touch `anomaly.py`.

Verified locally on a trial-merge with latest `main`: 36 tests pass, `ty` + `ruff` clean, full preflight green (`check_meta_index_sync` OK against the merged 26-task registry). Heads-up that the commit shows under my name because it went in via maintainer-edit on your fork.

LGTM — merging.

PR #30, opened 2026-07-06
feat(mmmlu): add mmmlu dataset and k-shot CLP base-model task

jack-scitix-ai:
## Type

- feature — new benchmark, task, or capability

## Summary

- Add **MMMLU** (OpenAI Multilingual MMLU) full multilingual k-shot base-model task + dataset (`hf:openai/MMMLU@325a01d`, revision-pinned).
- Scoring uses **`EvalMode.CLP`** (per #23): single-inference next-token conditional log-prob over the fixed A/B/C/D option tokens, read from the API's `top_logprobs`.
- Organization/aggregation follow lm-evaluation-harness `openai-mmmlu` (locale→category→subject, size-weighted); prompt + scoring follow Hendrycks `evaluate.py`.
- References: [OpenAI MMMLU](https://huggingface.co/datasets/openai/MMMLU), [lm-eval-harness openai-mmmlu](https://github.com/EleutherAI/lm-evaluation-harness/tree/1dd931087362abba74e0375c8c631295559f48b2/lm_eval/tasks/openai-mmmlu), [hendrycks/test](https://github.com/hendrycks/test), DeepSeek-V3 report (Qwen2.5-72B baseline).

**Differences vs lm-evaluation-harness `openai-mmmlu`**

- **Kept:** locale → category → subject organization and size-weighted (per-subject) aggregation; few-shot drawn from the test split (OpenAI MMMLU ships only translated test CSVs — no dev/train split).
- **Changed:** scoring is Hendrycks `evaluate.py` base-model completion (CLP) — one next token, `top_logprobs`, space-prefixed `" A"`/`" B"`/`" C"`/`" D"`, softmax over the four — instead of the chat/instruct (simple-evals) protocol the harness defaults to. Few-shot reserves the first `k` rows per `(Locale, Subject)` and excludes them from the scored set so no item appears in its own prompt.
- **Added (not in the reference):** optional stratified sub-sampling (`sample_fraction`/`sample_seed`/`sample_by`), applied in `setup()` *before* few-shot reservation so reservation, scoring, and aggregation are identical for full and sampled runs.

**Language selection** — a dataset concern via `MMMLUDataset.args.locales` (any subset of the 14 locales; Chinese is just `[zh_cn]` — there is no separate zh_cn task):

```yaml
datasets:
  mmmlu_zh_cn:
    class: MMMLUDataset
    path: openai/MMMLU
    args:
      locales: [zh_cn]          # one or more locales; list all 14 for full MMMLU
```

[...]

PR #33, opened 2026-07-07
feat(mmlu): add few-shot base-model clp task

jack-scitix-ai:
## Type

## Summary
- Add **MMLUFewShotCLPTask** (`mmlu_kshot_clp`): 5-shot base-model MCQ scored by the next single token's log-prob over A/B/C/D, read from `top_logprobs` in one `alogprobs(echo=False)` call (OpenCompass clp). Aligned with [hendrycks/test `evaluate_flan.py`](https://github.com/hendrycks/test/blob/master/evaluate_flan.py). Missing any option token from top-k fails the sample loudly instead of guessing (`DEFAULT_LOGPROBS=100`; vLLM needs `--max-logprobs 100`).
- **Repoint MMLUDataset** from the simple-evals CSV URL to `hf:cais/mmlu@c30699e` with the native schema (`question`/`subject`/`choices`/`answer`); loader is now load-only (no coercion/rename). **Breaking** to `MMLUDatasetSample`.
- **`mmlu_0shot_gen` migrated to the new schema** (coupled change): prompt text and gold answer are unchanged (same MMLU test set, positional A/B/C/D + int→letter), so a clean 0-fail score is identical; it previously had no unit test, now covered.
- **Unify `report()` to the full-set denominator** `len(finals)+len(fails)` in both MMLU tasks (pipeline failures scored wrong), matching the `gsm8k_0shot_gen` family; only shifts a reported number when `fails>0`.
- core: extend the empty-infer anomaly rule to `clp`/`top_logprobs`; add `choice_scores_from_top_logprobs` to `core/utils/ppl`.
- Reference: hendrycks/test `evaluate_flan.py`; target = DeepSeek-V3 report Table 6 (Qwen2.5-72B Base, MMLU 5-shot = 85.0).

### Automated

### Manual

Model: Qwen2.5-72B (base) · sglang · tp=2 · 5-shot · logprobs=100

| Model | Expected  | Actual | Diff |
|---|---|---|---|
| Qwen2.5-72B-Base(5-shot) | 85.0(DeepSeek-V3 Report) | 86.11| +1.11(-1.11%) |

### Required (all PRs)

### If: Breaking Change

### DIFF
--- a/sieval/core/tasks/anomaly.py
+++ b/sieval/core/tasks/anomaly.py
@@ -414,10 +414,10 @@ def detect_empty_infer_gen(ctx: TaskContext) -> set[int]:
 
 
 @sieval_detection_rule(
-    description="Inference result is empty for perplexity tasks (logprobs=[])",
+    description="Inference result is empty for logprob tasks (logprobs=[])",
     category="output_quality",
     rationale=("Empty logprobs indicate API failures or unsupported model features."),
-    applies_to=["ppl"],
+    applies_to=["ppl", "clp"],
     tags=["perplexity", "api_failure", "empty_output"],
 )
 def detect_empty_infer_ppl(ctx: TaskContext) -> set[int]:
@@ -426,14 +426,21 @@ def detect_empty_infer_ppl(ctx: TaskContext) -> set[int]:
     result = _unwrap_result(ctx.infer_result)
     if not isinstance(result, ModelOutput):
         return set()
-    # For PPL tasks, we need both logprobs and logprobs_tokens
+    # ppl reads logprobs/logprobs_tokens; clp reads top_logprobs. Flag whichever
+    # field the task populated when it comes back empty.
     has_logprobs = result.logprobs is not None
     has_logprobs_tokens = result.logprobs_tokens is not None
-    if has_logprobs or has_logprobs_tokens:
+    has_top_logprobs = result.top_logprobs is not None
+    if has_logprobs or has_logprobs_tokens or has_top_logprobs:
         logprobs_empty = has_logprobs and not result.logprobs
         logprobs_tokens_empty = has_logprobs_tokens and not result.logprobs_tokens
-        # Report index 0 as sentinel if any field is empty
-        return {0} if (logprobs_empty or logprobs_tokens_empty) else set()
+        top_logprobs_empty = has_top_logprobs and not result.top_logprobs
+        # Report index 0 as sentinel if any populated field is empty
+        return (
+            {0}
+            if (logprobs_empty or logprobs_tokens_empty or top_logprobs_empty)
+            else set()
+        )
     return set()
 
 
--- a/sieval/core/tasks/anomaly.py
+++ b/sieval/core/tasks/anomaly.py
@@ -414,10 +414,10 @@ def detect_empty_infer_gen(ctx: TaskContext) -> set[int]:
 
 
 @sieval_detection_rule(
-    description="Inference result is empty for logprob tasks (logprobs=[])",
+    description="Inference result is empty for perplexity tasks (logprobs=[])",
     category="output_quality",
     rationale=("Empty logprobs indicate API failures or unsupported model features."),
-    applies_to=["ppl", "clp"],
+    applies_to=["ppl"],
     tags=["perplexity", "api_failure", "empty_output"],
 )
 def detect_empty_infer_ppl(ctx: TaskContext) -> set[int]:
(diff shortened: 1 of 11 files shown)


ethan-scitix (commented):
Port verified faithful against the pinned `evaluate_flan.py` (format_subject / format_example / prompt header byte-exact, `dev[:k]`-per-subject shots, softmax→argmax equivalence, documented 2048-truncation deviation). Validation run confirmed: **Qwen2.5-72B base, 5-shot, 14042/14042, score 86.11 vs 85.0 target, 0 fails, deterministic across two runs.** Note: branch is stale — rebase onto latest `main` (advanced by #30 mmmlu, a direct CLP sibling).

**1. `truncated_output` anomaly fires on 100% of samples (`anomalies.json`: 14042/14042).** `alogprobs(max_tokens=1)` always finishes with `"length"`, so `detect_truncated_output` (default `applies_to=["all_tasks"]`) flags every clp/ppl sample — the anomaly report becomes pure noise. Pre-existing (cmmlu hits it too; mmmlu only escapes because it wraps infer output in `TaskStageOutput`). Since this PR already touches clp anomaly routing, fix it here: `sieval/core/tasks/anomaly.py:447` add `applies_to=["gen"]` to `detect_truncated_output` (truncation is only meaningful for generation). Existing routing tests use `{"gen"}` tags so they stay green; `test_all_tasks_rule_always_runs` (`test_anomaly.py:686`) pins the sentinel via this rule and would need repointing to `empty_postprocess`.

**2. `choice_scores_from_top_logprobs` duplicates cmmlu's private helper.** `sieval/core/utils/ppl.py:33` is a verbatim generalization of `cmmlu_kshot_base_gen.py:245` (`_choice_scores_from_top_logprobs`), and the fail-loud postprocess block is duplicated too (`mmlu_kshot_clp.py:201` vs `cmmlu_kshot_base_gen.py:401`). Extracting to `core/utils` is right, but migrate cmmlu onto it in this PR — otherwise the core helper is single-caller and the two copies will drift.

**3. CLP-family denominator convention.** `mmlu_kshot_clp.py:245` keeps infra fails in the denominator; the two method-siblings do the opposite (`mmmlu_kshot_clp` uses `total = len(finals)`; cmmlu "reported separately and excluded"). Zero numerical impact on the validated run (0 fails), so not blocking — but pick one convention for the base-model CLP family and note it. The docstring aligns with `mmlu_0shot_gen` (generative family); if that's intentional, say so explicitly given the CLP siblings differ.

**Nits:** (a) cite the 86.11 validated number in `reference_impl.notes` — cmmlu/mmmlu both do, and it's what justifies `status="stable"`. (b) `mmlu_kshot_clp.py:79` `zip(..., strict=False)` → `strict=True` (MMLU is fixed 4-way; sibling mmmlu uses strict). (c) `mmlu_0shot_gen.py:45` notes still say "aligned with simple-evals" while data now loads from `cais/mmlu` — annotate the data-path change (content is the same 14042-item test set).

ethan-scitix (commented):
Moved the full #32↔#33 coordination note to #32 to keep the review in one place — see https://github.com/scitix/sieval/pull/32.

One point specific to this PR: the empty-check here gates on `result.top_logprobs is not None`, so it only catches `[]`. `GenModel` coalesces an empty top-k to `None` (`gen_model.py:261`, `top_logprobs=top_token_logprobs or None`), which is the actual production-empty result — so that case is skipped. #32's `not result.top_logprobs` catches both; prefer that semantics when the two PRs are reconciled.

jack-scitix-ai:
Thanks for the thorough review — rebased onto latest `main` and worked through every point. Since #32 landed the shared helper + the anomaly coordination, a few of these fall out of the rebase itself.

**Helper duplication (#2) + empty-check (follow-up on #32):**
- After the rebase, `choice_scores_from_top_logprobs` comes solely from `core/utils/ppl.py` (#32's copy) — dropped our duplicate function and its `test_ppl.py` block, so both files are untouched here; `mmlu_kshot_clp` imports the shared one. cmmlu is already migrated on `main`.
- Per the #32 coordination note, the clp empty-infer rule is dead code (empty `top_logprobs` → `postprocess` raises → sample FAILED → never reaches the FINAL-gated `detect()`). Reverted `detect_empty_infer_ppl` and its test back to `main`; the `not top_logprobs` semantics still drive `postprocess`'s `all_present` → raise, just not as an anomaly rule.

**#1 — `detect_truncated_output` false-fires on every clp/ppl sample:**
- Fixed here as agreed on #32. Added `applies_to=["gen"]` (with a rationale noting `max_tokens=1` logprob tasks always finish `"length"`). Repointed `test_all_tasks_rule_always_runs` to `empty_postprocess`, and added `test_truncated_output_scoped_to_gen` pinning that it fires for `gen` but not `clp`.

**#3 — CLP-family denominator convention:**
- Picked `len(finals)` (infra failures reported separately in `fails`, not scored wrong), matching the method-siblings `cmmlu_kshot_clp` / `mmmlu_kshot_clp`; stated explicitly in the docstring and `reference_impl.notes`.
- Also reverted `mmlu_0shot_gen` back to its original `len(finals)` so this PR doesn't change a pre-existing task's scoring, and both MMLU tasks stay consistent — this removes the cross-family divergence you flagged. Denominator tests now pin the exclude-fails behavior.

**Nits:**
- (a) Cited the validated **86.11 (14042/14042, 0 fails, deterministic)** in `reference_impl.notes`.
- (b) `zip(CHOICES, choices, strict=True)`.
- (c) `mmlu_0shot_gen` note now reads "scoring aligned with simple-evals, data loaded from `cais/mmlu` (same 14042-item test set)".

ethan-scitix (approved):

PR #34, opened 2026-07-07
feat(hellaswag): add dataset and few-shot ppl task

jack-scitix-ai:
## Type

## Summary

- Add **HellaSwag** as a k-shot log-likelihood (PPL) task + dataset, aligned with
  EleutherAI **lm-evaluation-harness** `hellaswag` (https://github.com/EleutherAI/lm-evaluation-harness/tree/main/lm_eval/tasks/hellaswag): `multiple_choice` scoring — per-ending conditional log-prob, `acc` + `acc_norm` (headline = `acc_norm`), 0-shot..k-shot, validation split.
- Few-shot assembly is **byte-verified against lm-eval source** (delimiters `" "`/`"\n\n"`, gold *choice-text* answer via label→int→`choices[label]`, trailing delimiter, no description); only exemplar *selection* differs (our `retrieve_samples` vs lm-eval's `ContextSampler`) — documented as a deviation.
- Runs on the sglang backend via the official `SglangGenModel` (`engine: sglang`, #21); requires `--disable-radix-cache` for echo-logprob correctness.
- Continuation is isolated by char-offset (no tokenizer at this layer) and fails loud if it can't be located; report denominator counts pipeline failures (`len(finals)+len(fails)`).

### Automated

### Manual

- [x] Live run — Qwen2.5-72B-Base, 10-shot, sglang (`engine: sglang`, `--disable-radix-cache`):

  | Model  | Expected | Actual | Diff |
  |---|---|---|---|
  | Qwen2.5-72B-Base | 84.8 (DeepSeek-V3 report, Table 6) | 87.39| +2.59 (+2.59%) |

Target is from DeepSeek-V3's internal harness (protocol unspecified); this task follows lm-eval `acc_norm`, so a small different-harness gap is expected, not a repro miss.

### Required (all PRs)

### If: community/ Changes

### DIFF
--- a/sieval/community/hellaswag.py
+++ b/sieval/community/hellaswag.py
@@ -1,0 +1,29 @@
+# adapted from https://github.com/EleutherAI/lm-evaluation-harness/blob/1dd931087362abba74e0375c8c631295559f48b2/lm_eval/tasks/hellaswag/utils.py
+
+import re
+from collections.abc import Mapping
+from typing import Any, TypedDict
+
+
+class ProcessedDoc(TypedDict):
+    query: str
+    choices: list[str]
+    gold: int
+
+
+def preprocess(text: str) -> str:
+    text = text.strip()
+    # NOTE: Brackets are artifacts of the WikiHow dataset portion of HellaSwag.
+    text = text.replace(" [title]", ". ")
+    text = re.sub(r"\[.*?\]", "", text)
+    text = text.replace("  ", " ")
+    return text
+
+
+def process_doc(doc: Mapping[str, Any]) -> ProcessedDoc:
+    ctx = doc["ctx_a"] + " " + doc["ctx_b"].capitalize()
+    return {
+        "query": preprocess(doc["activity_label"] + ": " + ctx),
+        "choices": [preprocess(ending) for ending in doc["endings"]],
+        "gold": int(doc["label"]),
+    }
--- a/sieval/datasets/__init__.pyi
+++ b/sieval/datasets/__init__.pyi
@@ -29,6 +29,10 @@ from .gsm8k import (
     GSM8KDataset,
     GSM8KDatasetSample,
 )
+from .hellaswag import (
+    HellaSwagDataset,
+    HellaSwagDatasetSample,
+)
 from .hmmt_feb_2025 import (
     HMMTFeb2025Dataset,
     HMMTFeb2025DatasetSample,
@@ -109,6 +113,8 @@ __all__ = [
     "HMMTFeb2025DatasetSample",
     "HMMTFeb2026Dataset",
     "HMMTFeb2026DatasetSample",
+    "HellaSwagDataset",
+    "HellaSwagDatasetSample",
     "HumanEvalDataset",
     "HumanEvalDatasetSample",
     "IFBenchDataset",
--- a/sieval/datasets/hellaswag.py
+++ b/sieval/datasets/hellaswag.py
@@ -1,0 +1,65 @@
+"""
+HellaSwag dataset — commonsense sentence completion (4-way multiple choice).
+
+Mirrors the native ``Rowan/hellaswag`` parquet schema as-is. The lm-eval-harness
+query/choice text construction lives in the task layer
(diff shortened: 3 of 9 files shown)


ethan-scitix (commented):
Port verified byte-faithful to lm-eval `hellaswag` @ `1dd93108` (preprocess, few-shot assembly, acc/acc_norm). Reproduction confirmed on the author's run: Qwen2.5-72B 10-shot `acc_norm` **87.4** over the full 10042-sample validation split, `fails=0`, deterministic (two byte-identical runs). `status="stable"` is justified. Remaining items are comment/doc nits — none blocking.

**1. `Nit` — wrong comment on the report denominator.** `sieval/tasks/hellaswag_kshot_ppl.py:275` says the fails-in-denominator choice is "matching the sibling family and upstream." It isn't: the MCQ siblings (`openbookqa_kshot_gen`, `cmmlu_kshot_base_gen`) exclude fails, and lm-eval has no pipeline-failure concept. Behaviorally moot here (`fails=0`), but fix the wording so it doesn't propagate a wrong convention to copied tasks. (It matches `mbpp` only — say that if kept.)

**2. `Nit` — communicate the prefix-caching prerequisite at task level, as a principle.** echo scoring reads the prompt's *input* logprobs, which requires the serving backend's **prefix caching to be disabled** (cached positions aren't recomputed → truncated/skipped logprobs). Add one line to `reference_impl.notes` stating this **requirement**, not the backend flag — `--disable-radix-cache` (sglang) / `--no-enable-prefix-caching` (vLLM) is the backend layer's job (the `SglangGenModel` guard already owns that message). This is the only channel a future user has, since the PR ships no eval config. Confirmed load-bearing: the validated run set `disable_radix_cache: true` explicitly.

**3. `Nit` — fold the validated number into the docstring.** Replace the "expected to land near 84.8" hedge with the evidence (e.g. "validated: Qwen2.5-72B 10-shot `acc_norm` 87.4, full validation split, 0 fails, deterministic"), the way `cmmlu` cites its run. Provenance only.

**4. `Nit` (lowest) — license provenance.** `license="MIT"` is inferred from the upstream code repo; the `Rowan/hellaswag` HF card declares none. Fine as-is, just noting it against the "license = data repo's" convention.

jack-scitix-ai:
Thanks for the thorough review — agreed on all four; addressed in a follow-up `docs` commit (no behavior change, so # 1 is wording-only given `fails=0`).

**1. Denominator comment.** Fixed. Confirmed you're right: `openbookqa_kshot_gen` and `cmmlu_kshot_base_gen` both divide by `len(finals)` (exclude fails); only `mbpp_kshot_base_gen` uses `len(finals)+len(fails)` like this task. The comment now says it matches `mbpp`, notes the MCQ siblings exclude fails and that lm-eval has no pipeline-failure concept, and marks it a deliberate (moot-here) choice. Behavior unchanged; the regression test still locks it.

**2. Prefix-caching prerequisite.** Added to `reference_impl.notes` as a backend-agnostic *requirement*: "echo scoring reads the prompt's input logprobs, so the serving backend's prefix caching must be disabled (cached positions aren't recomputed → truncated logprobs); the backend layer owns the specific flag." Left the concrete flags out of the task — the `SglangGenModel` guard owns `--disable-radix-cache`, and vLLM's `--no-enable-prefix-caching` is its layer's concern. Agreed this is the only channel to a future user since the PR ships no eval config.

**3. Validated number in docstring.** Replaced the "expected to land near 84.8" hedge with the evidence: "Validated: Qwen2.5-72B-Base, 10-shot, `acc_norm` 87.4 over the full 10042-sample validation split, `fails=0`, deterministic (two byte-identical runs)." Kept 84.8 as a cross-check only (with the underspecified-protocol caveat) and named the reproduced method (lm-eval `hellaswag`). Provenance only, matching how `cmmlu` cites its run.

**4. License provenance.** Left as-is per your call. For the record: `license="MIT"` follows the upstream code repo (rowanz/hellaswag, MIT); the `Rowan/hellaswag` HF card declares none, so there's no data-repo license to defer to here.

PR #36, opened 2026-07-09
feat(arc): add ARC-Easy/Challenge datasets + ppl and clp tasks

jack-scitix-ai:
## Type

## Summary

- Adds the **ARC-Easy** and **ARC-Challenge** benchmark with **two eval methods each** (base models, `model_type: gen`, `engine: sglang`): `arc_*_kshot_ppl` (**separation** — score each option's full text) and `arc_*_kshot_clp` (**options** — score the answer letter).
- **Why two methods:** [arXiv 2412.17758v1](https://arxiv.org/abs/2412.17758) shows ARC has two incompatible eval regimes — *separation* (score option text) vs *options* (list choices, score the letter) — that models silently switched between (Llama 2→3, Mixtral, DeepSeek pre/post-V2) and that give very different numbers. We implement both and map them to sieval modes: **separation ⇔ `ppl`**, **options ⇔ `clp`**. Judged by the scored unit: separation scores a multi-token option *text* → `ppl` (per-candidate sequence likelihood); options scores a single *letter* token → `clp` (next-token logprob over option tokens). Confirmed empirically (see table: `ppl` lands in the separation regime, `clp` in the options regime).
- **Code reference = OpenCompass**, adapted: OpenCompass has `ARC_c/e_ppl_a450bd` (separation, full text) and `ARC_c/e_ppl_2ef631` (options, letter), both via PPLInferencer. Our `clp` instead reads the option letter's next-token logprob from `top_logprobs` in **one** inference — **equivalent** to 2ef631 for a single-letter continuation (min-perplexity-of-letter ≡ max-logprob-of-letter ≡ our argmax), but cheaper and using sieval's first-class `clp` mode (mechanism also mirrors CMMLU `qwen2.py`). Our `ppl` mirrors `a450bd` but normalizes **unconditionally** ([Brown et al. 2020](https://arxiv.org/abs/2005.14165) / [DeepSeek-V1's documented ARC choice](https://arxiv.org/abs/2401.02954)) rather than OpenCompass's length-norm perplexity; unconditional norm already corrects for length, so no length normalization is applied.
- **Prompts follow the DeepSeek reports:** `ppl`/separation = DeepSeek-V1 Appendix A.6 Table 19 (`Question:\nAnswer:` + full-text candidate); `clp`/options =  DeepSeek-V2 report Appendix G Table 13 (`Question:\nA. …\nAnswer: {letter}`, blank-line few-shot).

### Dataset note: one repo, two configs

`arc_challenge` and `arc_easy` are two **configs of the same HF repo** `allenai/ai2_arc`. `sieval dataset download` mirrors a repo file-level (`snapshot_download`, keyed by repo_id), so `sieval dataset download arc_easy` **and** `sieval dataset download arc_challenge` each stage the entire `allenai/ai2_arc/` once — i.e. downloading either name makes **both** datasets available. This is expected (shared storage, no duplication), not a bug; there is no per-config sub-download because the upstream repo isn't split that way.

### Why `ppl` does not list options (theory + experiment)

`clp` lists the options; `ppl` deliberately does not (no options-listed variant is shipped). Both a theoretical argument and an A/B experiment justify this:

- **Theory.** `ppl` scores each option's *full text* as a continuation and normalizes unconditionally (subtracting `logP(text | "Answer:")`), which assumes the options are **not** in the context. Listing them breaks two things: (1) scoring "repeat option *i*" becomes a **copying** task — copy-probability is high and similar for all listed options, so content discrimination collapses; (2) the unconditional baseline is computed **without** the options, so it no longer matches the now-options-listed conditional, mis-calibrating the normalization (systematically favouring long/rare options that are hard to generate but easy to copy).
- **Experiment** (separation vs. options-listed, full text):

  | Model / shots | Split | separation | options-listed | Δ |
  |---|---|---|---|---|
  | Qwen2.5-72B, 25-shot | ARC-C | 71.08 | 72.44 | +1.4 |
  | Qwen2.5-72B, 25-shot | ARC-E | 85.31 | 77.65 | −7.7 |
  | DeepSeek-7B, 0-shot | ARC-C | 50.09 | 32.59 | −17.5 |
  | DeepSeek-7B, 0-shot | ARC-E | 68.48 | 34.60 | −33.9 (≈ random) |

  The direction is inconsistent and mostly negative — noise on a strong/primed model, catastrophic on a weak/0-shot one. A valid method moves consistently; this confirms listing options is incoherent for full-text scoring. The coherent way to use options is `clp`. So `ppl` is separation-only.

### Automated

### Manual

Models: Qwen2.5-72B-Base, DeepSeek-LLM-7B-Base (sglang via `engine: sglang`).
**How targets were chosen.** The reports don't all state their ARC method, so we inferred it from arXiv 2412.17758v1's separation-vs-options analysis plus each report's appendix prompt template, and confirmed by score match:
- **`clp` ↔ DeepSeek-V3** (options): V3's appendix has no template, but V3 states it reuses V2's setup, and V2 Table 13 is the options format (letters, score the letter). Our `clp` reproduces it → compared to V3 Table 3.
- **`ppl` ↔ DeepSeek-V1** (separation): V1 Table 19 is the separation format (`Question:\nAnswer:`, score option text) with unconditional normalization. Our `ppl` reproduces it → compared to V1 Table 6.
- **Qwen2.5 report** does not state its ARC method; our `ppl` (71.08) matches its 72.4 while `clp` would be ~96, so we infer Qwen used a separation-style method and compare `ppl` to it.

| Task | Model | Split | Expected | Actual | Source | Diff |
|---|---|---|---|---|---|---|
| `clp` | Qwen2.5-72B-Base, 25-shot | ARC-C | 94.5 | 95.9 | DeepSeek-V3 report, Table 3 | +1.4(1.4%) |
| `clp` | Qwen2.5-72B-Base, 25-shot | ARC-E | 98.4 | 98.91 | DeepSeek-V3 report, Table 3 | +0.51(0.51%) |
| `ppl` | Qwen2.5-72B-Base, 25-shot | ARC-C | 72.4 | 71.08 | Qwen2.5 report, Table 2 | −1.32(-1.32%) |
| `ppl` | DeepSeek-LLM-7B-Base, 0-shot | ARC-C | 48.1 | 50.09 | DeepSeek-V1 report, Table 6| +1.99(1.99%) |
| `ppl` | DeepSeek-LLM-7B-Base, 0-shot | ARC-E | 67.9 | 68.48 | DeepSeek-V1 report, Table 6| +0.58(0.58%) |

### Required (all PRs)


ethan-scitix (changes requested):
### Blocking

**`Bug:` PPL comparison target is wrong** — `sieval/tasks/arc_challenge_kshot_ppl.py:23`, `arc_easy_kshot_ppl.py:23`
Docstrings cite *"DeepSeek-V3 base = 94.5 / 98.4 (Table 3)"*. Two errors:
- 94.5 / 98.4 is the **Qwen2.5-72B** column of Table 3, not DeepSeek-V3 (which is 95.3 / 98.9).
- That whole row is the **letter/options** regime. The bare-cloze PPL here empirically scores **71.08 / 85.31** — ~23 / ~13 pts below. PPL's real anchor is the separation number (Qwen2.5 report ~72.4, which the `clp` docstring already names). Repoint it — it doesn't belong to Table 3.

| Qwen2.5-72B, 25-shot | Challenge | Easy |
|---|---|---|
| CLP (letters) — this PR | 95.90 | 98.91 |
| PPL bare-cloze — this PR | 71.08 | 85.31 |
| Table 3 "V3" cell (actually Qwen col) | 94.5 | 98.4 |

**`Nit:` CLP target value is right but mislabeled** — `arc_challenge_kshot_clp.py:18`
94.5/98.4 is the correct value for a Qwen2.5-72B run (empirical 95.90/98.91 is in the band). Relabel `"DeepSeek-V3"` → `"Qwen2.5-72B-Base (DeepSeek-V3 report, Table 3)"`.

**`Nit:` Set `status="experimental"`** on all four — none is validated against a correct target, and the PPL rendering isn't settled (bare-cloze 71.08 vs the `-listed` text variant 72.44, which matches 72.4 better).

### Minor

**`Bug:` PPL sums the trailing generated token** — `arc_*_kshot_ppl.py:150-155`
`echo=True` + default `max_tokens=1` returns 897 input + **1 generated** token; `total_logprob` over the full sequence includes it, so it doesn't cancel in `cond − uncond`. Net impact is small (C 71.08→71.25, E 85.31→85.02; ~3% of predictions flip), but the docstring's *"exact for EM"* is false and the unit mock (`[None, value]`) hides it. Slice to `usage.input_tokens` before summing.

**`Nit:`** PPL echo calls should pass `logprobs=0` (top-k unused), matching the `hellaswag` sibling.

jack-scitix-ai:
Thanks for the thorough review — all five points addressed in the latest push (also synced with the latest `main` and regenerated the conflicting auto-generated registry files, `meta/index.json` + stubs).

**1. PPL target wrong (blocking).** Fixed. Repointed both `arc_challenge_kshot_ppl.py` and `arc_easy_kshot_ppl.py` (docstring + `reference_impl.notes`) to the **separation** regime — Qwen2.5-72B-Base ARC-Challenge ≈ 72.4 (Qwen2.5 report) — and removed the DeepSeek-V3 / Table 3 attribution. Agreed: 94.5/98.4 is the options/letter regime and belongs to the `clp` sibling, not PPL.

**2. CLP target mislabeled.** Fixed. Relabeled both `clp` tasks to `Qwen2.5-72B-Base ARC-Challenge/Easy 25-shot EM = 94.5 / 98.4, from the DeepSeek-V3 report Table 3 (Qwen2.5-72B-Base column)`, and noted DeepSeek-V3-Base's own entry is 95.3 / 98.9.

**3. status.** Set `status="experimental"` on all four tasks. Agreed none is validated against a settled target yet — and even the corrected PPL number is a *number* match, not a proven identical-method reproduction (the unconditional-normalization choice vs the reference harness isn't confirmed).

**4. PPL sums the trailing generated token (bug).** Fixed. Added a shared `echoed_logprob()` in `sieval/tasks/_arc.py` that slices the echoed sequence to `usage.input_tokens` before summing, so the generated token (from `max_tokens >= 1`) is dropped and the shared context cancels exactly in `cond − uncond`. Both PPL `postprocess` methods now use it, and the "exact for EM" wording is corrected. The mock no longer hides it (it now returns `usage.input_tokens` plus a trailing generated token), and `test_echoed_logprob_excludes_trailing_generated_token` fails against the old full-sum logic.

**5. logprobs param.** PPL echo calls now pass `logprobs=0` to match the `hellaswag` sibling; the mock asserts it.

**Score impact / re-run.** CLP is unchanged (95.9 / 98.9 stand — none of these touch its path). PPL shifts by <0.3 from #4 only (your measurement: C 71.08→71.25, E 85.31→85.02). Since the tasks are now `experimental` and the delta is sub-0.3, I didn't re-run — happy to re-run the PPL tasks for exact post-fix figures if you'd prefer them in the table.

ethan-scitix (approved):
Re-review: the prior `CHANGES_REQUESTED` is resolved — all 5 comments verified fixed (PPL target repointed to the separation regime, CLP relabeled, `status="experimental"` on all four, trailing-token drop via `echoed_logprob` with a discriminating test, `logprobs=0`). Approving. The four nits below are non-blocking cleanups worth folding in while you're here — the citation fix is the one I'd do before merge.

**`Nit:` Wrong citation author** — `arc_challenge_kshot_clp.py:4`, `arc_easy_kshot_clp.py:4`
The "options" format is attributed to *"Wang et al. 2024, arXiv:2412.17758"*. The arXiv ID is right (it's the paper the options/separation framing comes from), but the author is **Łukasz Borchmann** (*"ARC 'Challenge' Is Not That Challenging"*, Findings of ACL 2025). Fix the name, or drop it and keep the bare arXiv ID like the ppl docstrings do.

**`Nit:` Duplicate of an existing core helper** — `_arc.py:109` (`clp_scores_from_top_logprobs`)
Byte-identical to `choice_scores_from_top_logprobs` in `core/utils/ppl.py:33`, which both clp siblings already reuse (`cmmlu_kshot_clp.py:384`, `mmlu_kshot_clp.py:203`). This is a generic clp primitive with no ARC-specific content, so it should be shared to keep the family from drifting. Drop the copy, import the core helper, and pass `tuple(labels)` at the two call sites (`arc_challenge_kshot_clp.py:139`, `arc_easy_kshot_clp.py:138`).

**`Nit:` Silent full-sum when `usage` is absent** — `_arc.py:176-181`
If `output.usage` is `None`, the slice is skipped and the full echoed+generated sequence is summed — silently reintroducing the trailing-token bug this function exists to prevent. Unreachable today (the sglang echo path guarantees `usage`), but in an echo-scoring context prefer raising on `input_tokens is None` over scoring the wrong span.

**`Nit:` ARC-Easy ppl has no concrete anchor** — `arc_easy_kshot_ppl.py:24-28`
Only re-cites the Challenge ≈72.4; there's no Easy number. The Borchmann paper reports Easy separation figures — cite one, or state explicitly that no Easy separation anchor is used.

---

*Heads-up, not a request — your call:* once ARC reuses it, `choice_scores_from_top_logprobs` is a clp helper imported from a module named `ppl.py`. It's a pre-existing quirk (cmmlu/mmlu already live with it, and the module is documented as "log-probability extraction utilities", so it's arguably fine). If it bugs you, rename/move it in a separate `core` cleanup — not this PR, since it'd fan out to the other two clp tasks.

jack-scitix-ai:
Thanks — all four nits addressed:

1. **Citation.** Both `clp` docstrings now cite `Borchmann, ARC "Challenge" Is Not That Challenging` (Findings of ACL 2025; arXiv:2412.17758) instead of "Wang et al."
2. **Duplicate helper.** Dropped the local `clp_scores_from_top_logprobs` copy from `_arc.py`; both `clp` tasks now import `choice_scores_from_top_logprobs` from `sieval.core.utils.ppl` and pass `tuple(labels)` at the call sites.
3. **Silent full-sum in `echoed_logprob`.** It now raises when `usage` is absent instead of falling back to a full-sequence sum, so it can't silently reintroduce the trailing-token bug it exists to prevent.
4. **ARC-Easy ppl anchor.** The docstring now states explicitly that no ARC-Easy *separation* figure is used as an anchor — neither the Qwen2.5 nor the DeepSeek reports give one (their headline ARC number is Challenge ≈ 72.4, on the ppl sibling), so the score is left uncalibrated (`status="experimental"`).
Left the `choice_scores_from_top_logprobs`-living-in-`ppl.py` module-name quirk untouched as you suggested — glad to tackle that in a separate core cleanup rather than here.

ethan-scitix:
Pushed a small maintainer follow-up (`a477c956`) directly to this branch: reworded the two `*_ppl` docstrings + `reference_impl.notes` so the **separation** regime is attributed to the Qwen2.5-report style rather than "DeepSeek's setup" (the `clp` sibling is the one reproducing DeepSeek-V3's options format), and regenerated `meta/index.json` to match. Docstring-only, no behavior change; CI is green.

⚠️ Heads-up: I amended + force-pushed, so `feat/arc` was rewritten (`99e1d64f` → `a477c956`). If you have local commits on this branch, `git fetch` + `git reset --hard origin/feat/arc` (or rebase) before pushing again.

PR #41, opened 2026-07-21
chore(evaluator): vendor code-evaluator + float-tol + scicode support

jack-scitix-ai:
## Type

- chore — CI, tooling, dependencies, config

## Summary

- Convert `code-evaluator` from a git submodule (`submodules/code-evaluator` → `scitix/code-evaluator`) to a self-contained vendored copy under `vendor/code-evaluator`, byte-identical to the previously pinned commit `e4802268`. A plain `git clone` is now self-contained (no `git submodule update`). Requested in review.
- Port [`float-tol`](https://github.com/ethan-scitix/code-evaluator/tree/fix/checker-messages-float-tol) (`fix/checker-messages-float-tol`, `cfc47d8`): clearer LiveCodeBench output-comparison messages + opt-in float tolerance via `CODE_EVAL_FLOAT_TOL` (default off = exact `==`, matching official LiveCodeBench).
- Add SciCode execution support: route `source="scicode"` through the direct-run path, add `requirements/scicode.txt` (numpy/scipy/sympy/matplotlib, pinned) and `docker/Dockerfile.scicode` (`python:3.11-slim`, since scipy 1.16.x needs Python ≥ 3.11). venv install remains supported.
- Retarget ruff/ty/mypy `exclude` from `submodules/` to `vendor/` so the vendored third-party code isn't linted/type-checked.
- Provenance recorded in `vendor/code-evaluator/VENDORED.md`; `code-evaluator` is an internal scitix repo with no separate LICENSE. No AI-generated Python modules in this PR (vendored third-party + config/docs only).
- The sieval-side SciCode dataset/task lands in a follow-up PR that depends on this one.

### Automated

- [x] Lint/format clean (`ruff check && ruff format --check`)
- [x] Type check clean (`ty check` or `mypy --strict`)
- [x] Unit tests pass (`pdm run pytest`)

### Manual

- [x] `ty check sieval` clean; `ruff`/`format` clean with `vendor/` excluded; `pdm run pytest tests/unit` → 2379 passed. Vendored tree verified byte-identical to `e4802268` (+ documented float-tol/scicode patches); `app/*.py` compile; `source="scicode"` alias present.

### Required (all PRs)

- [x] PR title follows conventional format (`type(scope): description`)
- [x] No internal paths, credentials, or personal info in committed files
- [x] AI-generated code has `AI-Generated Code - <model> (<provider>)` in module docstring
- [x] No new upper-layer dependencies added to `core/`
- [x] Deleted code verified — no remaining call sites depend on it

PR #42, opened 2026-07-23
feat(scicode): add SciCode dataset + 0-shot generation task

jack-scitix-ai:
## Type

## Summary

- Add the **SciCode** research-coding benchmark ([paper](https://arxiv.org/abs/2407.13168) / [repo](https://github.com/scicode-bench/SciCode), pinned @ `69a8cfc`).
- `SciCodeDataset`: test split 65 problems / 288 tested sub-steps (+3 scientist-authored gold steps used only as context), dev 15/50; upstream jsonl revision-pinned, numeric-target `raw_ground.h5` mirrored on HF as a plain HTTPS artifact (sha256-checked).
- `SciCodeZeroShotGenTask` (`scicode_0shot_gen`): sequential per-sub-step generation in the self-dependency setting (upstream default — step *i* conditions on the model's own code for steps 1..i-1); optional scientist-background axis (`with_background=False` by default); scoring via the remote code-eval service (`source="scicode"`, vendored in #41).
- **`special_step_mode` (SCORE-RELEVANT), default `verbatim`**: `verbatim` injects the whole gold block for steps 13.6/62.1/76.3 — the fix for open upstream bugs `scicode-bench/SciCode#59` / `#49`, whose extraction silently drops the `class` wrapper and makes problems 13/62 structurally unsolvable; `extract` byte-matches that upstream behavior for public-leaderboard parity (one task-arg switch). The effective task default is therefore no-background + `verbatim`. `status="experimental"`.
- Vendored upstream helpers under `sieval/community/scicode/` with Apache-2.0 headers preserved; both prompt templates are byte-identical to the pinned upstream files. New `scicode` optional-dependency group (`h5py`/`numpy`/`scipy`) for eval-side h5 target reads (incl. `scipy.sparse`).
- Robustness: `postprocess` reads h5 targets off the event loop (`asyncio.to_thread`, one hop per problem); the `feedback` client HTTP timeout is sized above the eval subprocess timeout so a step that runs to that timeout is recorded as a failed step (the service returns `status=False`) rather than tripping a client `ReadTimeout` that fails the whole problem. A conditional legacy-compatible `scipy.integrate.simps` wrapper is injected only for the four programs affected by upstream issue `scicode-bench/SciCode#2`, avoiding a global SciPy downgrade.

## Related Issues

Refs #41 (vendored code-evaluator with scicode support). Upstream bug reports: `scicode-bench/SciCode#59`, `scicode-bench/SciCode#49`.

### Automated

- [x] Type check clean (`ty check` — scicode files clean; `community/` excluded per repo config)
- [x] Unit tests pass (`pdm run pytest tests/unit` — 2622 passed; 15 scicode tests)

### Manual

- [x] Full `scripts/check_preflight.py` PASS (registration, naming, dep-coverage, meta-index sync, version)
- [x] `sieval dataset download scicode` succeeds (jsonl + h5 staged, checksums verified)
- [x] `pdm install --check --frozen-lockfile` clean (lock adds only `h5py`; no version drift)
- [x] Qwen2-72B-Instruct calibration vs official (SciCode paper Tables 2/3), committed executable code @ `1939af0`, T=0 greedy, full test split (65 problems / 288 tested steps, `fails=0`, `import_errors=0`), self-dependency, sglang tp8 bf16:

| Background setting | `special_step_mode` | main (ours) | main (expected) | Δ main | sub (ours) | sub (expected) | Δ sub |
|---|---|---:|---:|---:|---:|---:|---:|
| without background (Table 2 = leaderboard) | `extract` (upstream parity) | 1.5 | 1.5 | **0.0** | 13.9 | 17.0 | −3.1 |
| without background | `verbatim` (**task default**) | 1.5 | N/A | N/A | 13.9 | N/A | N/A |
| with background (Table 3) | `extract` (upstream parity) | 4.6 | 4.6 | **0.0** | 24.0 | 22.2 | +1.8 |
| with background | `verbatim` | 4.6 | N/A | N/A | 25.0 | N/A | N/A |

- The expected values and diffs apply only to `extract`, because the official Qwen2-72B results were produced with the upstream extraction pipeline. No official expected values exist for the corrected `verbatim` mode. The two `extract` rows are therefore the apples-to-apples calibration: main-problem accuracy (the leaderboard's ranking metric) matches official **exactly** (Δ 0.0) under both background settings. The sub-problem differences are −3.1 points without background and +1.8 points with background, in opposite directions across the two settings, consistent with cross-inference-stack variation.
- **Why `verbatim` does not change the no-background score:** Qwen2-72B is not strong enough in this setting to produce otherwise-correct implementations for the affected class-dependent steps. Under `extract`, those steps fail because the required class was dropped. Under `verbatim`, the harness bug is fixed, but the model-generated downstream code is already incorrect and still fails on its own merits. The failure cause changes, but none of the 288 pass/fail outcomes changes, so both main and sub scores remain identical.
- **Why the modes diverge with background:** with scientist background enabled, `verbatim` rescues exactly `13.14`, `62.2`, and `62.4`, which fail with `NameError` under `extract`. The committed rerun therefore gains three passing steps (69/288 → 72/288). Main remains 3/65 because neither problem 13 nor problem 62 becomes fully solved. The same three steps flip on a stronger model (gpt-5.6: `extract` 7/19 → `verbatim` 10/19 on the two affected problems), showing that the impact of the upstream bug scales with model capability.
- All four post-fix runs report `import_errors=0`. The conditional compatibility wrapper removes the `scipy.integrate.simps` ImportErrors from `2.1` and `28.1–28.3`; those steps remain incorrect because of downstream model-generated `ValueError`, `NameError`, `AssertionError`, or `UnboundLocalError`, so the wrapper changes their failure causes but adds no Qwen passes.
- A previous paired calibration—not this committed rerun—observed a verdict flip at `14.2` despite byte-identical generated code and a byte-identical complete test program. Its implementation uses unseeded `numpy.random.normal()` in a Monte Carlo estimate checked against a fixed tolerance, providing evidence of observed execution-side ±1-problem (about ±1.5-point) main-score jitter. In the committed rerun, `14.2` passes in both background modes and contributes nothing to their difference.

### Required (all PRs)

### If: community/ Changes

### If: New Dependency


ethan-scitix (changes requested):
**Request changes.** Logic is faithful to upstream + the internal GLM-OpenEvals impl (verified: both jsonl checksums, h5 byte-identity via sha256, prompt templates, `extract_python_script`, and `previous_llm_code` handling all match). Three things to fix:

### 🔴 Blocking

- **Missing gold-step files.** `special_step_code()` (`sieval/community/scicode/prompts.py:102`) reads `data/{13.6,62.1,76.3}.txt`, but `.gitignore:156` (`data/`) excludes them — they're not committed and not on disk. Reproduced: all three raise `FileNotFoundError`, so problems 13/62/76 crash in `infer()` and pollute main-problem accuracy. `reference_impl.notes` claims these are vendored, but they aren't. Suggest inlining them as a `.py` module (like `community/livecodebench/prompts/*.py`) to avoid both the gitignore and wheel-packaging (`[tool.pdm.build]` declares no `includes` for data files).

### 🟠 Repro fidelity (blocks matching official numbers)

- **Timeout.** `timeout=300.0` (`sieval/tasks/scicode_0shot_gen.py:116`) vs the official hardcoded `1800` in upstream `test_generated_code.py`. 6× shorter → spurious timeouts on slow numerical steps. Bump to `1800` or document why.
- **Temperature.** Official generation is greedy (`temperature=0`). Correctly not hardcoded here, but please note in the docstring/`reference_impl.notes` that reproducing official numbers requires `temperature=0` in the model config. (`with_background=False` default is already correct — matches the official headline mode.)

### Nits

- Remove the unused `get_function_from_code` / `extract_function_name` exports (`sieval/community/scicode/__init__.py`) — dead code; they read as "aligning with upstream" but nothing calls them (upstream only uses them on the resume-from-disk path, which this task doesn't take).
- Flesh out `reference_impl.notes` with the real deviations: h5 targets inlined + remote eval service (vs in-subprocess), special-step verbatim embed, and pipeline-`fails`-counted-as-unsolved accounting.
- Document the sandbox dependency `sympy` (injected by `_cmp_upstream.py`; not in any sieval dep group — must exist in the code-eval service image).
- `_cmp_source()` (`sieval/community/scicode/harness.py:26`) re-reads the file per tested step — wrap in `functools.lru_cache`.

Blocking = the gold files + timeout; the rest can be quick follow-ups.

jack-scitix-ai:
Thanks for the thorough review — all six addressed and pushed. Point by point:

### 🔴 Blocking

- **Gold-step files.** Inlined the three non-generated gold steps into `sieval/community/scicode/_gold_steps.py` as a `GOLD_STEP_CODE` dict (per your `livecodebench/prompts/*.py` suggestion), and removed the `data/*.txt` dependency. `special_step_code()` now reads from that module — no filesystem read, so it's committed and wheel-packaged (no more gitignore/`[tool.pdm.build]` gap). The inlined strings are **byte-identical** to the upstream files (2098 / 978 / 1020 bytes, round-trip asserted before deleting the originals; delimiter chosen per file since 13.6 contains `"""` and 76.3 contains `'''`, all raw to preserve `\n`). Verified problems 13/62/76 no longer raise `FileNotFoundError`. `reference_impl.notes` updated to reflect the inline (it no longer claims a data-file vendor).

- **Timeout.** Bumped the task default `timeout` 300 → **1800** to match upstream `test_generated_code.py`.

### 🟠 Repro fidelity

- **Temperature.** Documented in `reference_impl.notes`: reproducing official numbers requires greedy decoding (`temperature=0`) in the model config; kept it out of the task (per-run knob). Also noted `with_background=False` default matches the official headline mode.

### Nits

- Removed the unused `get_function_from_code` / `extract_function_name` exports from `community/scicode/__init__.py` (grepped — no callers; the definitions stay in `parse.py` for upstream alignment).
- Fleshed out `reference_impl.notes` with the real deviations: h5 targets read eval-side and inlined into the sandbox program (vs in-subprocess), execution via the remote code-eval service over HTTP, special-step verbatim embed, and pipeline-`fails`-counted-as-unsolved in main-problem accuracy.
- Documented the sandbox `sympy` dependency (injected by the vendored comparison shim; provided by the code-eval service image via `requirements/scicode.txt`, not a sieval dep group) in `reference_impl.notes`.
- Wrapped `_cmp_source()` in `functools.lru_cache(maxsize=1)`.

ethan-scitix (changes requested):
**Request changes.** Logic is faithful and the prior round is fully addressed, but verifying against the runs in `outputs/` surfaced one blocking issue plus a few nits.

### 🔴 Blocking — committed code diverges from upstream, and the reported number isn't from it

Special steps are embedded **verbatim** (`sieval/tasks/scicode_0shot_gen.py:170-189`, `sieval/community/scicode/prompts.py:96-102`), but upstream (`gencode_json.py`) runs `get_function_from_code(gold, extract_function_name(header))` — which resolves to `__init__` and extracts a **bare method, dropping the `class` wrapper** for 13.6/62.1 (and `EnlargedBlock` for 62). This isn't cosmetic:

- In the `scicode-kimi` run, problem 13's steps `13.8–13.15` all fail with `NameError: Maxwell` (extraction path). The committed verbatim code makes `13.8`/`13.14` pass (confirmed in `scicode-kimi-p13`; the model never re-defines `Maxwell`). On a 10-problem slice this moves main accuracy 20%→30%.
- The PR's **without-background 37.8** comes from `scicode-kimi`, which used the **old extraction path at T=0.0** — not the committed verbatim code, and not the "T=1.0 official hyperparams" stated for the with-bg number. There is no full-set run of the committed code in headline (without-background) mode.
- `reference_impl.notes` says reproduction needs `temperature=0`, but the validated with-bg number (49.3) used T=1.0; `reference_impl.url` points at upstream (extraction) while the code does verbatim. Task is `status="stable"`.

Please: (1) re-run the full 65-problem set with the committed code in without-background mode and update the number, with a single stated temperature; (2) decide which reference the numbers target (public upstream = extraction, or the internal verbatim variant) and make code/url/notes consistent, flagging the special-step handling as a **score-relevant** deviation; (3) set `status="experimental"` until a full run of the committed code validates it.

### 🟠 Should fix

- `postprocess` (`sieval/tasks/scicode_0shot_gen.py:243`) opens the ~1 GB h5 synchronously per tested step inside `async def`, stalling the loop for concurrent samples. Read once per problem or `await asyncio.to_thread(...)`.

### ⚪ Nits

- The evaluator caps memory at 1024 MB (`Sample.memory_limit` default, not overridden here); upstream's subprocess has no cap — memory-heavy numerical steps could OOM here but pass upstream. Worth a note in `reference_impl.notes`.
- `.gitignore:192-197`: the personal `scicode-kimi*.yaml` entries will ship to OSS as noise — prefer a glob or `.git/info/exclude`.
- `MULTISTEP_TEMPLATE` (`sieval/community/scicode/prompts.py:29-52`) has a trailing `\n` upstream lacks (only affects `with_background=True`).
- Dataset description says "80 problems / 338 sub-steps" (`sieval/datasets/scicode.py:1,54`); the headline test split is 65 / 288 (80/338 is test+dev).

jack-scitix-ai:
### 🔴 Blocking: special-step divergence & score provenance → all three requested actions done

**(1) Full 65-problem re-run with the committed code (T=0, greedy)** — calibrated against Qwen2-72B-Instruct, which is on the public leaderboard (sglang, tp8, bf16):

| Setting | main (ours / official) | sub (ours / official) |
|---|---|---|
| no-bg (paper Table 2 = leaderboard) | **1.5 / 1.5 — exact match** | 13.9 / 17.0 (−3.1) |
| with-bg (paper Table 3) | **4.6 / 4.6 — exact match** | 25.0 / 22.2 (+2.8) |

Main accuracy (the leaderboard's ranking metric) reproduces exactly under both settings. The ±3 on sub comes from cross-stack generation differences (the official runs did not use sglang; greedy decoding is not token-identical across stacks). The two settings deviate in opposite directions, so it is not a systematic bias. The previously reported numbers (Kimi-K2.6: 49.3 with-bg @ T=1.0; 37.8 no-bg from the pre-toggle extraction path) predate the committed code and are superseded by this calibration — they remain in the PR description only as a historical note.

**(2) Code/URL/notes consistency + score-relevance** — added a `special_step_mode` task arg; the notes flag it as SCORE-RELEVANT. The `extract` branch is upstream's `get_function_from_code` path (the header's `def` matches before `class`, so 13.6/62.1 collapse to a bare `__init__`).

**(3) `status="experimental"`** — set.

---

### Proposal on the default: `verbatim` (the fix) as default, `extract` (upstream parity) as a one-line switch

Rationale (all empirically backed):

**a. The upstream behavior is a known open bug (`scicode-bench/SciCode#59`, `#49`) and makes problems 13/62 structurally unsolvable for every model.** After extract-injection, every dependent step fails with `NameError` regardless of what the model writes (measured: with gpt-5.6-terra under extract, 13.8–13.15 and 62.2/62.4 all NameError; a Kimi run's prompt contexts contained 0 occurrences of `class Maxwell`). That measures the harness, not the model. The fix proposed in those upstream issues is exactly verbatim.

**b. The practical cost to leaderboard comparability is ≈ 0:**
- Qwen2-72B (on the leaderboard) scores **identically** under both modes (the table above is the shared result), so the verbatim default still reproduces leaderboard main accuracy exactly;
- The impact on strong models has a hard bound. On a controlled 2×2 over only the two affected problems (13/62, 19 tested steps; gpt-5.6-terra, non-thinking, T=0), the two modes differ in exactly three class-dependent steps:

  | step | extract | verbatim |
  |---|---|---|
  | 13.14 `initialize(maxwell)` | NameError | ✅ PASS |
  | 62.2 `block_initial(...)` | NameError | ✅ PASS |
  | 62.4 `block_enlarged(block, ...)` | NameError | ✅ PASS |
  | remaining 16 steps | identical | identical |

  Scaled to the full split: +3/288 ≈ **+1.0 sub-accuracy pt, main unchanged** — smaller than the ±3 cross-stack noise above;
- When bug-for-bug parity is needed: `special_step_mode: extract` in the task args, prominently documented in the notes.

If you still prefer extract as the default, flipping it back is a one-line change plus doc swap — calibration data exists for both sides. Your call.

---

### 🟠 Should fix: synchronous h5 reads in postprocess — fixed
All numeric-target reads for a problem are now batched into a **single `asyncio.to_thread` hop per problem** (h5py serializes HDF5 access internally, so worker threads are safe). The event loop is no longer blocked.

### ⚪ Nits — all addressed
- **Memory cap**: documented in the notes (evaluator sandbox defaults to 1 GB vs upstream's uncapped subprocess; memory-heavy numeric steps could OOM here yet pass upstream).
- **.gitignore**: the five personal `scicode-kimi*.yaml` entries removed (moved to local `.git/info/exclude`).
- **Trailing `\n` in `MULTISTEP_TEMPLATE`**: removed; both templates are now **byte-identical** to the pinned upstream files (verified by script).
- **Dataset counts**: docstring/description now state the precise split sizes — test 65 problems / 288 tested steps (plus 3 gold context-only steps), dev 15/50; "80/338" is the tested-step total across both splits.

---

### Proactively disclosed additional changes in this round
1. **feedback client HTTP timeout buffer +5s → +120s.** Observed in practice: when a step legitimately runs to the eval service's own subprocess timeout (e.g. 62.6 `run_dmrg` @1800s), the service correctly returns `status=False` ("subprocess timeout") — but the +5s client buffer tripped a `ReadTimeout` first, failing the **whole problem** as a pipeline error instead of recording that one step as failed. This is a scoring-correctness fix. Note it cannot cause upstream divergence: the score-relevant sandbox timeout (1800s) is unchanged and matches upstream exactly; the client timeout is a transport-layer knob with no upstream counterpart (upstream runs subprocesses locally, no HTTP), and the pre-fix "step timeout → whole problem fails" behavior was itself the side inconsistent with upstream semantics.
2. **The `extract_function_name` / `get_function_from_code` exports removed in round 1 as dead code are restored** — the extract path now genuinely uses them.

ethan-scitix (commented):
Verified the calibration against the runs in `outputs/scicode-qwen2-72b-bugmatrix` and `outputs/scicode-gpt56-p1362` — every reported number reproduces, and the `verbatim` default is well-supported. All prior-round items are resolved.

### Verified

- **Qwen2-72B full 2×2**, 65 problems, `fails: 0`, T=0 / top_p=1.0, `deterministic: true` seed 0: no-bg 1.538 main / 13.889 sub (1/65, 40/288), bg 4.615 / 25.0 (3/65, 72/288). Matches the PR table exactly; the `extract` and `verbatim` reports are byte-identical.
- **The "identical under both modes" result is a real coincidence, not an inert toggle** — this was the claim I most wanted to check. In `extract_nobg`, Qwen's 13.8–13.14 fail `NameError: 'Maxwell'` and 62.2/62.4/62.5/62.6 fail `NameError: 'Block'`; under `verbatim` the same steps fail with `AssertionError` / `ValueError` / `subprocess timeout` — i.e. the model's own answers. Worth putting in the PR body; it's stronger evidence for the default than the score table.
- **gpt-5.6 on 13+62**: extract 7/19 vs verbatim 10/19 = **+3 steps**, main unchanged. The flipping steps are exactly 13.14 / 62.2 / 62.4, each `NameError` → PASS, and the other 16 are identical. +3/288 ≈ +1.0 sub pt, as documented.
- **Port fidelity**: gold steps and both prompt templates byte-identical to upstream @ `69a8cfc`; both jsonl sha256 recomputed from a fresh download and matching; split counts (65/288, 15/50) recomputed; `reference_impl.url` and the h5 mirror both 200; upstream #59/#49 exist, open, titles match the claim. Full preflight PASS.

### Should fix

**1. `Bug:` `tests/unit/tasks/test_scicode_0shot_gen.py:17` — all 10 tests are skipped in CI.** The module-level `pytest.importorskip("h5py")` skips the whole file, and `.github/workflows/ci.yml:15` installs `-G drop -G ifeval -G math -G test -G dev` with no `scicode`. Skipped ≠ failed, so CI stays green — including for the two guards on the SCORE-RELEVANT `special_step_mode` default. Fix: add `-G scicode` to `INSTALL_GROUPS` (h5py/numpy/scipy, no torch, consistent with the "light" comment). `sympy` for the postprocess test already comes from the `math` group, so this lifts all 10 and the module-level skip can stay as a no-op. I confirmed 9/10 pass without h5py; the 10th couldn't run here, so please eyeball the first CI run.

**2. `Question:` provenance of the with-bg numbers.** Per the reflog, both runs had HEAD at the round-1 commit with the round-2 changes uncommitted. Of the differences between that commit and `b578481`, the only score-affecting one I couldn't rule out is the `MULTISTEP_TEMPLATE` trailing-`\n` removal — which applies **only** to `with_background=True`. (`special_step_mode` was present at run time: the config records it. The `+5`→`+120` timeout change is provably inert here — `fails: 0` with 4 steps recorded as step-level `subprocess timeout`.) Since `BACKGROUND_COMMENT_TEMPLATE` is untouched, **the no-bg 1.5/13.9 — the leaderboard headline — is fully attributable to the committed code**; only the bg pair carries the gap. Either drop 4.6 from the "reproduces the leaderboard main accuracy exactly" phrasing in `reference_impl.notes`, or re-run just `verbatim_bg` + `extract_bg`.

### Nits

- `sieval/tasks/scicode_0shot_gen.py:146` — annotate `special_step_mode` as `Literal["extract", "verbatim"]`; keep the runtime check since the value arrives from YAML. Precedent: `sieval/tasks/t_eval_before_calling_0shot_gen.py:53`.
- `sieval/community/scicode/__init__.py` — missing the `AI-Generated Code - <model> (<provider>)` docstring last line; `sieval/community/ifbench/__init__.py:12` has it, as do the other new files in this PR.
- `sieval/tasks/scicode_0shot_gen.py:233` — `output.texts[0]` raises `IndexError` on an empty-choices response, failing the whole problem instead of one step; sibling `livecodebench_code_generation_0shot_gen.py:110` iterates `inf.texts`.
- `sieval/tasks/scicode_0shot_gen.py:309` — `feedback()` evaluates steps sequentially at up to 1800s each. Fine, but worth one line recording it as deliberate: `verbatim` raises the stakes, since 4 of Qwen's problem-13 steps burn the full 1800s under `verbatim` where `extract` killed them instantly with `NameError`.

jack-scitix-ai:
Round-3 items are addressed, and the branch is now rebased onto the current `main` after the v0.7.0 merges. I also squashed the four review-round commits into a single `feat(scicode)` commit while rebasing, so the earlier SHAs are no longer present.

The `pdm.lock` / `pyproject.toml` conflict was resolved by regenerating the lockfile with the union of the `ruler`, `ruler-gen`, and `scicode` groups. The resulting lockfile adds only `h5py`, with no unrelated version drift.

### Should fix

1. **SciCode tests silently skipped in CI**

   Fixed by adding `-G scicode` to `INSTALL_GROUPS` in `ci.yml`. All 15 SciCode tests are now collected and executed rather than skipped.

2. **With-background score provenance**

   Re-ran `extract_bg` and `verbatim_bg` after the final prompt-template fix, using the same full-split calibration configuration: Qwen2-72B-Instruct, T=0 greedy, self-dependency, sglang TP8/BF16, 65 problems / 288 tested steps, and zero pipeline failures.

   | Mode | Current main | Current sub | Previous main | Previous sub |
   |---|---:|---:|---:|---:|
   | `extract_bg` | 4.6 | 24.0 | 4.6 | 25.0 |
   | `verbatim_bg` | 3.1 | 24.7 | 4.6 | 25.0 |

   Official expected values apply only to `extract`, because the official Qwen2-72B results were produced with the upstream extraction pipeline. The PR table therefore reports expected values and diffs only for the `extract` rows; the corresponding `verbatim` cells are `N/A`.

   The apples-to-apples `extract` calibration still reproduces official main-problem accuracy exactly:

   - no background: 1.5 ours / 1.5 official;
   - with background: 4.6 ours / 4.6 official.

   The two new with-background runs differ on four of 288 pass/fail outcomes:

   - `13.14`: `extract` fails with `NameError`; `verbatim` passes;
   - `62.2`: `extract` fails with `NameError`; `verbatim` passes;
   - `62.4`: `extract` fails with `NameError`; `verbatim` passes;
   - `14.2`: `extract` passes; `verbatim` fails with `AssertionError`.

   The first three are the class-dependent steps directly rescued by the `verbatim` fix. They are also the same three steps that flip in the controlled gpt-5.6 comparison (`extract` 7/19 → `verbatim` 10/19 on problems 13 and 62).

   The unrelated `14.2` step has byte-identical generated code and a byte-identical complete test program across the two runs. Its Monte Carlo implementation repeatedly calls unseeded `numpy.random.normal()` and checks whether the estimated result falls within a fixed tolerance. One execution passes while the other raises `AssertionError`, so the flip is consistent with unseeded stochastic execution rather than a generation difference or `special_step_mode`. The saved artifacts do not record the RNG state, so I cannot attribute it to a specific random draw more strongly than that.

   This also explains the aggregate results:

   - `verbatim` rescues three affected sub-steps, but the unrelated `14.2` flip produces a net gain of two steps (69/288 → 71/288);
   - the rescued steps do not make every step of problems 13 or 62 pass, so they do not add a solved main problem;
   - the observed `14.2` flip changes problem 14 from solved to unsolved, producing the one-problem main-score change (4.6 → 3.1, i.e. 1/65 ≈ 1.5 points).

   I also clarified why `extract` and `verbatim` have identical scores without background. Qwen2-72B is not strong enough in that setting to produce otherwise-correct implementations for the affected class-dependent steps. Under `extract`, those steps fail because the required class was dropped; under `verbatim`, that harness bug is fixed, but the model-generated downstream code has already introduced its own errors before the fix can turn the step into a pass. The failure cause changes, but none of the 288 pass/fail outcomes changes, so the aggregate main and sub scores remain identical.

   The PR body now distinguishes:

   - `extract` as the official-pipeline calibration mode;
   - `verbatim` as the corrected default;
   - the lack of official expected values for `verbatim`;
   - the no-background tie caused by model capability;
   - the three class-dependent steps rescued with background;
   - the unrelated stochastic-execution evidence for the `14.2` flip.

### Nits

1. Annotated `special_step_mode` as `Literal["extract", "verbatim"]`; the runtime validation remains because the value can arrive from YAML.
2. Added the required AI-generated attribution line to `sieval/community/scicode/__init__.py`.
3. Guarded `output.texts[0]` in `infer()`. An empty-choices response now becomes an empty extraction instead of failing the entire problem; a regression test was added.
4. Documented that per-step evaluation in `feedback()` is deliberately sequential.

ethan-scitix (commented):
**Verified:** 15/15 tests now run (was 0 — `-G scicode` fixed it), full `tests/unit` 2677 pass, preflight PASS. Gold steps + both templates byte-identical to upstream; `parse.py` AST-identical modulo docstrings/lazy imports; split counts and both jsonl sha256 recomputed; h5 checksum confirmed via HF's LFS etag. All four PR-body numbers reproduce exactly. **Round-3's with-bg provenance question is closed** — the bg re-run is dated 2026-08-03 and 57/65 generations differ from the old run, exactly as the bg-only trailing-`\n` fix predicts.

### 🔴 Blocking

**1. `Bug:` 4 tested steps are unsolvable for every model — `scipy.integrate.simps` doesn't exist in the sandbox.** Steps `2.1`, `28.1`, `28.2`, `28.3` fail `ImportError` in all four runs. Problems 2 and 28 declare `from scipy.integrate import simps` in `required_dependencies`; `simps` was removed in SciPy 1.14, and `vendor/code-evaluator/requirements/scicode.txt` pins `scipy==1.16.3`. Both problems have *only* those tested steps, so both are structurally unsolvable — the same defect class as `#59`/`#49`, which this PR's whole argument is built on. Ceiling loss: **−4/288 sub (−1.39 pt), −2/65 main (−3.08 pt)** — against a no-bg headline of 1.54.

Known upstream: **`scicode-bench/SciCode#2`**, open since 2024-07-21, zero comments. Belongs in `reference_impl.notes` next to `#59`/`#49`.

Unlike those, there's **no parity trade-off**: the official numbers predate SciPy 1.14, so `scipy<1.14` *is* upstream parity and the current pin diverges with no upside. Either pin `scipy<1.14` (faithful, but changes the sandbox for all 25 scipy-using problems → full re-calibration) or inject a `simps = simpson` alias (surgical — only the 4 dead steps change, other 284 programs stay byte-identical, no re-calibration; note the `even`-handling difference). Also, that requirements file's "verified across all 80 problems" comment is falsified by the run data.

Qwen2's calibration still matched only because it solves 1–3 problems and was never going to solve those two; a stronger model gets systematically understated.

**2. `Bug:` the shipped default's with-background main accuracy is 3.08, not 4.62.**

```
bg-rerun/scicode_extract_bg    main 4.615 (3/65)   sub 23.958 (69/288)
bg-rerun/scicode_verbatim_bg   main 3.077 (2/65)   sub 24.653 (71/288)
```

The PR table is labelled `extract` and is correct, but `verbatim` is the default and its bg main appears nowhere. The extra difference is step **14.2**, PASS under `extract` → `AssertionError` under `verbatim` — and problem 14 has no special step, so the toggle can't reach it. Both runs' `raw_response` *and* evaluated `program` are **byte-identical** (9547 B, same sha256). The reason: `14.2`'s solution is a 4000-sample Monte-Carlo estimator asserted against a ±5% tolerance with **no seed in the test body** (`14.1`'s test body does seed, which is why it's stable). Measured floor: 815 byte-identical program pairs → exactly 1 verdict disagreement, all of it `14.2`.

Three corrections needed:

- PR body's "differs from `extract` only on the class-dependent steps of problems 13/62" is false — `14.2` differs too, in the opposite direction, and costs a whole problem.
- The notes blame "greedy sglang decoding." That source is real — over the 62 non-special problems, generations diverge on 1–2 per pairing (no-bg 62/62 identical; bg 60/62 and 61/62). But those problems (64, 73) changed **zero** verdicts, so they explain none of the swing. Keep it, add the eval-side source.
- Say that main accuracy carries an irreducible ±1-problem (±1.5 pt) jitter — 30–50% relative at these score levels. Otherwise "reproduces the leaderboard exactly" reads as reproducible.

### ⚪ Nits

- **`Nit:`** `report()` doesn't surface timeouts, and `verbatim_nobg` has **4** steps at the full 1800 s (`13.11`–`13.15`) where `extract_nobg` has **0** — ~2 extra hours invisible in `report.json`. Precedent: `livecodebench_code_generation_0shot_gen.py:197,203`. Same place could surface the 2–3 `MemoryError`s per run.
- **`Nit:`** `sieval/tasks/scicode_0shot_gen.py:338` — the scalar `timeout=self._timeout + 120` also sets httpx's **pool** timeout, and connections cap at `max_concurrency` (default 4, `:166`). Above that, a >1920 s queue wait raises `PoolTimeout` → re-raised → whole problem counts as unsolved. Latent here (`max_concurrency: 16`, `fails: 0`), but that relied on knowing to raise the knob. Fix: `httpx.Timeout(self._timeout + 120, pool=None)`.
- **`Nit:`** `:315` says "max_concurrency problems in flight" — that's the connection-pool cap; problems in flight is the run's `concurrency`.
- **`Nit:`** `harness.py:71-77` inlines the cmp source at top level, leaking `np`/`scipy`/`sympy` into program globals, so a solution using an undeclared module passes here and `NameError`s upstream (confirmed by construction). But an AST audit of all **1152** evaluated programs found **0** depending on it — real, never exercised. Harden (exec into the shim module's own `__dict__`; a plain `del` breaks the cmp helpers) or just annotate.
- **`Nit:`** nothing pins `timeout == self._timeout + 120` — no test for the round-2 scoring fix. Family-wide gap.

jack-scitix-ai:
Round-4 items are addressed. I also completed a fresh full four-way calibration after the `scipy.integrate.simps` compatibility fix.

### Blocking

1. **`scipy.integrate.simps` compatibility**

   I chose a surgical compatibility wrapper rather than downgrading SciPy globally or using a simple `simps = simpson` alias.

   The wrapper is injected only into the four affected programs (`2.1`, `28.1`, `28.2`, and `28.3`) when their dependencies contain `from scipy.integrate import simps`; the other 284 tested programs are unchanged. It preserves the pre-SciPy-1.14 positional API, including positional `x` and `even="avg" | "first" | "last"` behavior.

   The open upstream issue (`scicode-bench/SciCode#2`) and the compatibility adaptation are now documented in `reference_impl.notes` and `requirements/scicode.txt`. A regression test covers the legacy positional and `even` behavior while also verifying that unaffected programs receive no compatibility code.

   I then reran all four full configurations with the committed executable code at `1939af0`:

   | Setting | Mode | Main | Sub | Correct steps | `import_errors` |
   |---|---|---:|---:|---:|---:|
   | no background | `extract` | 1.5 | 13.9 | 40/288 | 0 |
   | no background | `verbatim` | 1.5 | 13.9 | 40/288 | 0 |
   | with background | `extract` | 4.6 | 24.0 | 69/288 | 0 |
   | with background | `verbatim` | 4.6 | 25.0 | 72/288 | 0 |

   All four runs cover 65 problems / 288 tested steps and report `fails=0`.

   The four previously blocked steps now execute past the import:

   - without background: `2.1` reaches a model-generated `ValueError`; `28.1–28.3` reach model-generated `AssertionError`s;
   - with background: `2.1` reaches a model-generated `NameError`; `28.1`/`28.3` reach model-generated `UnboundLocalError`s, and `28.2` reaches an `AssertionError`.

   Thus the environment-level ImportErrors are fixed. These steps still do not pass because Qwen's downstream implementations are incorrect, so this compatibility correction changes their failure causes but adds no Qwen passes.

2. **Default `verbatim` with-background score and stochastic jitter**

   The fresh committed rerun gives the shipped default (`verbatim`) a with-background score of **4.6 main / 25.0 sub** (3/65 problems, 72/288 steps), versus `extract` at **4.6 / 24.0** (3/65, 69/288).

   The current with-background runs differ on exactly three verdicts:

   - `13.14`: `extract` fails with `NameError`; `verbatim` passes;
   - `62.2`: `extract` fails with `NameError`; `verbatim` passes;
   - `62.4`: `extract` fails with `NameError`; `verbatim` passes.

   `verbatim` therefore gains exactly three steps (69/288 → 72/288). Main remains 3/65 because neither problem 13 nor problem 62 becomes fully solved.

   In this fresh rerun, `14.2` passes under both background modes and contributes nothing to their difference. I retained the earlier `14.2` observation only as historical jitter evidence: that prior paired run used byte-identical generated code and a byte-identical complete test program, but one execution passed and the other raised `AssertionError`. Because the implementation uses unseeded `numpy.random.normal()` in a Monte Carlo estimate, this supports an observed execution-side ±1-problem (about ±1.5-point) main-score jitter; the saved artifacts do not contain the RNG state, so I do not attribute it more strongly than that.

   The PR table and `reference_impl.notes` now distinguish the fresh committed result from that earlier observation. Official expected values and diffs remain limited to `extract`; the corrected `verbatim` rows remain `N/A`.

   The no-background equality is also preserved and documented: restoring the missing classes changes the affected steps from harness-level `NameError`s to execution of Qwen's own downstream code, but that code is still incorrect. Consequently, none of the 288 no-background verdicts changes.

### Nits

1. `report()` now surfaces `timeouts`, `memory_errors`, and `import_errors`; the fresh reports confirm `import_errors=0`.
2. The feedback request now uses `httpx.Timeout(self._timeout + 120, pool=None)`, so queued requests do not fail with `PoolTimeout`; connect/read/write deadlines remain bounded. A regression test asserts these timeout fields.
3. The concurrency comment now distinguishes task-runner sample concurrency from the HTTP connection-pool cap and records why per-step evaluation is deliberately sequential.
4. The comparison helper is executed in its own module namespace, so `numpy`/`scipy`/`sympy` no longer leak into solution globals. A regression test confirms undeclared imports fail as they would upstream.
5. `requirements/scicode.txt` no longer claims the original environment worked unchanged across all problems; it documents the conditional compatibility wrapper and its Python/SciPy target.

### Validation and provenance

The four reports identify the pre-rebase executable calibration commit as `1939af0`. The required rebase maps that implementation to `e0be6b1`: the stable patch-id is identical and every SciCode executable file is byte-identical. The only added base change is upstream #51 plus its generated metadata-index update. The documentation-only follow-up is now `bc3199c`.

ethan-scitix:
**Round 5 — verified against the runs in `outputs/scicode-qwen2-72b-post-simps-committed`.** All four reports reproduce the PR table exactly (1.538/13.889, 1.538/13.889, 4.615/23.958, 4.615/25.0; `fails=0`). Every round-4 item is resolved. Re-verified from scratch this round: both jsonl sha256s recomputed; the h5 checksum confirmed three ways (a local copy of the official `test_data.h5`, HF's LFS `x-linked-etag`, and the declared value all equal `48b0272a…41b890`, size 1,049,345,865 exact); gold steps byte-identical (2098/978/1020 B); both prompt templates byte-identical; `_cmp_upstream.py` identical modulo a trailing newline; the `#59`/`#49` extraction bug reproduced directly (13.6/62.1 lose the `class`, 76.3 unaffected); the `simps` wrapper checked against a faithful pre-1.14 reference over 285 randomized cases on the image's own SciPy 1.16.3 — 0 mismatches.

Two maintainer-edit commits are pushed to this branch (clean fast-forwards, no history rewritten): `b017d5f0` and `f93d5b01`. Details below.

### 🔴 What round 5 found: the `simps` wrapper was incomplete

The round-4 fix only covered the *declared-import* route. Problems **12** and **57** declare a module handle instead —

```
problem 12 → from scipy import integrate
problem 57 → from scipy import integrate, optimize
```

— and Qwen calls it through that handle:

```python
norm = integrate.simps(u**2, r_grid)
```

The shipped guard (`"from scipy.integrate import simps" in code_content`) does not match, so no wrapper was injected and those steps died on:

```
failed: [AttributeError] module 'scipy.integrate' has no attribute 'simps'
```

Consistent across all four runs — **13–14 steps each**, 12 of them being problem 12, which therefore scored **0/14 and was structurally unsolvable** for any model taking that route. Same defect class as `#59`/`#49` and as round-4's own finding.

**This is why `import_errors=0` did not catch it:** that route raises `AttributeError`, not `ImportError`. (`ModuleNotFoundError` was also uncounted — its name contains no "importerror" — though there are 0 occurrences in these runs, so no existing number is affected.)

### ✅ Score impact: zero, and measured rather than assumed

The stored `program` is self-contained (deps + the model's own prior-step code + inlined targets + tests), so an execution-side fix can be scored by replaying it — generation is byte-identical, only execution changes. I prepended the wrapper to all **68** newly covered failing steps across the four runs:

| run | newly covered | of which simps-dead | **new passes** |
|---|---:|---:|---:|
| extract_nobg | 17 | 14 | **0** |
| verbatim_nobg | 17 | 14 | **0** |
| extract_bg | 17 | 13 | **0** |
| verbatim_bg | 17 | 13 | **0** |

66 became Qwen's own errors (`AssertionError` 23, `ValueError` 9, `TypeError` 9, `NameError` 1); 2 were `12.14`, which still times out when re-run at full budget (1900 s, no output). **Zero additional passes**, so all four scores are unchanged and the `extract` anchors still match the official figures exactly — no re-calibration needed.

Same pattern as your `scipy.integrate.simps` fix: the failure *cause* changes, the verdicts do not. The value is the ceiling — without it a stronger model silently loses up to **13/288 sub-steps** plus all of **problem 12**.

### Commits pushed

**`b017d5f0`**
- `build_test_program()` selects the wrapper by bare name (`\bsimps\b`) instead of the exact import line — covers `from scipy.integrate import quad, simps` and attribute-style calls, while `simpson` correctly does not match.
- `report()` counts `ModuleNotFoundError` as an import error.
- `simps` docstring narrowed to the `even=` surface actually restored (SciPy 1.11–1.13's `even="simpson"` is not reproduced).
- 8 new test cases: a parametrized table pinning inject/no-inject across 8 import forms, and a `ModuleNotFoundError` row in the report test.

**`f93d5b01`** — `reference_impl.notes` corrected: the wrapper is no longer scoped to problems 2/28, the 12/57 route and the `AttributeError`-vs-`ImportError` distinction are recorded, and the replay result (0 additional passes) is stated.

Both mutation-tested. `ruff` + `ty` clean, **2692** unit tests pass, preflight PASS, `meta/index.json` regenerated, CI green.

### Nit (not blocking)

- **`Nit:`** un-blocked steps now execute instead of failing instantly, so problem 12 adds real wall clock (`12.14` alone exceeds the 1800 s step timeout). Expect bg runs' `timeouts` to go 2 → 3. Same trade you already accept for `verbatim` on problem 13 (`13.11`–`13.15`). Noted in the docstring; no action needed, just worth knowing when sizing runs.

### Withdrawn from round 4

I floated lowering the `scipy>=1.16.3` floor in the `scicode` dependency group — **withdrawn**, that pin is the repo-wide convention (`drop` and `ruler` use the identical floor). Changing it only here would have created the inconsistency.

### Verdict

No open technical items. Reproduction confirmed, fidelity confirmed, and the one new defect is fixed with its score impact measured at zero.

ethan-scitix (approved):

PR #43, opened 2026-07-23
feat(hle): add Humanity's Last Exam dataset + 0-shot LLM-judge task

jack-scitix-ai:
## Type

## Summary

- Adds **Humanity's Last Exam (HLE)** (Center for AI Safety; Phan et al., 2025): dataset loader, a 0-shot generative task, and a vendored metric kernel.
- Reference repo: https://github.com/centerforaisafety/hle/tree/26dca2e253b405105b4c3d8c2f5af06f86f90c66/hle_eval — `SYSTEM_PROMPT` / `JUDGE_PROMPT` / `calib_err` vendored byte-for-byte.
- Task: model answers under the HLE `Explanation/Answer/Confidence` prompt; a separate **LLM judge** (mandatory `grader` arg) decides correctness; metrics are accuracy, 95% Wald CI, and calibration error. Text-only subset by default.
- Dataset `load()` stays minimal; auxiliary `Image` columns are set `decode=False` to stay Pillow-free while preserving the raw `image` column.
- Deviations from upstream documented in module docstrings (judge reached via `ChatModel` text parsing instead of `beta.chat.completions.parse`; o1 system→user role swap dropped).
- Review cleanup: dropped a dead `_text_only` attribute; added byte-level pinning tests on the vendored prompts so upstream drift fails loudly.

### Automated

### Manual

Dataset loading validated end-to-end: full runs load `cais/hle` at the pinned revision and materialize 2158 text-only samples (same loader path as `sieval dataset download hle`).

0-shot, single attempt (n=1), non-tool — matching the official HLE protocol. Expected column: gpt-oss-20b HLE (text-only, no tools) from the [gpt-oss model card (arXiv:2508.10925)](https://arxiv.org/abs/2508.10925).

| --- | --- | --- | --- |
| gpt-oss-20b (reasoning=high) | 10.9 | 12.14 | +1.24(+1.24%) |
| gpt-oss-20b (reasoning=low)  | 4.2  | 3.61  | −0.59(-0.59%) |

Both within the <3 pp target; runs are clean (low: fails=1, no truncation; high: fails=0). The high-effort run flags 517/2158 (24%) `truncated_output` anomalies — long reasoning chains hitting `max_tokens` with an empty answer channel — yet still lands within target; expected reasoning-model behavior, not an eval defect.

Note on methodology: our judge is GPT-5.2 (text-only subset); the model card's HLE judge may differ, and HLE accuracy is judge-sensitive — treat this as a close, not identical-protocol, comparison.

- [x] Ran both effort settings; scores + fail/truncation counts captured.

### Required (all PRs)

### If: community/ Changes

### If: New Dependency


ethan-scitix (commented):
**Approve** — faithful port, prompts/metrics byte-exact vs upstream@26dca2e, and the gpt-oss-20b validation checks out (12.14/3.61 vs card 10.9/4.2, <3pp, like-for-like incl. truncation). `stable` is justified. A few non-blocking nits:

- `Nit:` Mirror the validation numbers into `reference_impl.notes` — PR bodies don't ship with the artifact, and `simpleqa_verified` already sets the precedent of putting its validation line in the notes (they get embedded in `index.json`). One line is enough, e.g. *"VALIDATION: gpt-oss-20b 12.14/3.61 vs card 10.9/4.2 (<3pp), judge GPT-5.2, text-only."*
- `Nit:` `sieval/tasks/hle_0shot_gen.py:241` — rename `n_judged` → `n_graded` to match the judge siblings (`browsecomp`, `simpleqa_verified`).
- `Nit:` `report()` could echo the truncated/anomaly count alongside `fails`. The 12.14 headline is a valid like-for-like number, but a reader of `report.json` alone can't see that 24% of outputs truncated — surfacing it makes the headline self-documenting for a collapse-prone benchmark.
- `Nit:` `sieval/community/hle.py:70` — `_CORRECT_RE` matches inside `"incorrect:"`; safe today (last-match + field order), but `\bcorrect` removes the alias for free.

jack-scitix-ai:
Thanks for the approval and the sharp nits — all four addressed in the follow-up commit:
- **`n_judged` → `n_graded`** — renamed the `report()` key; now matches `browsecomp` / `simpleqa_verified`. Grepped the tree — no other consumer read the old key. The two report tests were updated to assert `n_graded`.
- **Surface truncation in `report()`** — confirmed `report()`'s contexts already carry `infer_result.finish_reasons`, so no `core/` change was needed. Added a `truncated` count (per-attempt `finish_reason == "length"`, None-guarded) next to `fails`; a reader of `report.json` alone now sees e.g. 517/2158 truncated. New `test_report_counts_truncated_outputs` builds contexts with real `ModelOutput` finish reasons; `truncated == 0` is pinned in the empty/fails cases.
- **VALIDATION in `reference_impl.notes`** — added the one-liner (`gpt-oss-20b 12.14 / 3.61 vs card 10.9 / 4.2, <3pp, judge GPT-5.2, text-only`) so it ships in `index.json`, matching the `simpleqa_verified` precedent.
- **`\bcorrect` anchor** — added the word boundary so `_CORRECT_RE` can't match inside `incorrect:`. Added a discriminating test: `"...42 is incorrect: yes"` would parse as `(True, 100)` under the old pattern, and correctly defaults to `(False, 100)` with the anchor.

ethan-scitix (commented):
**Approve** — faithful port. Prompts (`SYSTEM_PROMPT`/`JUDGE_PROMPT`) byte-identical to upstream@`26dca2e`, `calib_err`/`dump_metrics` logic matches (incl. the dropped-last-bin quirk), dataset revision + MIT license verified, all 4 prior nits landed, 23 tests green. Two items before merge:

- `Nit:` `sieval/community/hle.py:9` and `:92` — the `hendrycks/outlier-exposure` link is on `master`; pin it so it can't drift. Latest commit touching the file (blob 200, still contains `calib_err`):
  ```
  https://github.com/hendrycks/outlier-exposure/blob/404c6268865e84dbab5ab4ccf855e8ae39de853f/utils/calibration_tools.py
  ```
  (The HLE `reference_impl` URL and `hf:cais/hle@5a81a4c` are already pinned — only this one needs it.)
- `Question:` The o1 `system`→`user` swap is intentionally dropped (documented). Confirming o1-as-candidate is out of scope — otherwise an o1 candidate gets a `system` message it rejects. Fine for the gpt-oss/GLM targets.

Please rebase onto latest `origin/main` before merge.

ethan-scitix (commented):
**1. `Bug:` `sieval/community/hle.py:75-89`, `sieval/tasks/hle_0shot_gen.py:255` — an unparseable judge reply is counted as graded, dropping upstream's two failure signals.**

Upstream's `extract_answer` returns `None` on failure, so the record never enters `judged_predictions`: excluded from both `correct` and `confidence`, still in `n`. Upstream also reports it twice — `run_judge_results.py:137-138` and `:143-145` (`if len(correct) != n: print(...)`). This port appends `(False, 100)` instead: accuracy-neutral, but it injects a max-confidence/wrong point that inflates `calibration_error`, and since `n_graded` counts it as graded, upstream's discrepancy signal is permanently 0 here. Note upstream's `except` is annotated `# very, very rare` (`:62`) — with schema enforcement it only fires on API errors; the text path makes format drift fire the same branch, so the trigger surface is far wider while observability went to zero.

Suggest **not counting unparsed replies as graded** — closes three gaps at once (upstream drop semantics for the calibration array, `n` vs `n_graded` restored, no new report key). `parse_judge` needs a contract change since it can't distinguish "judged no" from "couldn't parse"; a `judge_parsed: bool` on `JudgeFeedback` keeps the community change minimal and mirrors how `truncated` was done last round.

Separately, `:28` and `:82` justify the 100 default by citing upstream's *"Put 100 if there is no confidence score available"* — that line instructs the judge about the **candidate's** confidence, not upstream's judge-failure path.

**2. `Nit:` `sieval/community/hle.py:71-72` — the `\**` tolerance misses the most common bolding.** Asterisks after the colon are accepted, before it aren't:

```
(True, 90)   <- 'correct: **yes**'   (False, 100) <- '**correct**: yes'   <-- misses
(True, 90)   <- '**correct:** yes'   (False, 100) <- '*correct*: yes'     <-- misses
```

`r"\bcorrect\**\s*:\s*\**\s*(yes|no)"` (same for `_CONFIDENCE_RE`) closes it. No upstream analogue — `ExtractedAnswer.correct: Literal["yes","no"]` makes markdown impossible there — so this is net-new surface never validated upstream. Not a live bug with GPT-5.2, but a portability risk for the next judge.

**3. `Nit:` `sieval/tasks/hle_0shot_gen.py:26-30` — document the judge's token budget.** Upstream hardcodes `max_completion_tokens=4096` (`run_judge_results.py:48`); the docstring gives the *candidate's* upstream defaults but not the judge's. Risk is higher here than upstream: structured output means a compact reply, whereas a reasoning judge on the text path can burn its budget before reaching `correct:` and silently become `(False, 100)` via #1.

**4. `Nit:` `sieval/datasets/hle.py` — `cais/hle` is gated (`gated: auto`).** `sieval dataset download hle` needs an accepted gate + `HF_TOKEN`; undocumented, no precedent in-repo (GPQA is a `url:` source). One line in the loader docstring.

**5. `Question:` `sieval/tasks/hle_0shot_gen.py:129-130` — `text_only` runs *after* the session's dataset `operations:`, silently shrinking sampling budgets.** `_apply_dataset_operations` (`session.py:1278-1330`) runs at dataset-resolution time, before `task_class(...)` at `:1395`. On a 100-row fixture at HLE's ~14% image fraction: `slice: {num: 20}` → 17 samples, `stratified_sample per_group=20` → 17. Harmless on the reproduction path, surprising for smoke configs — one line noting the ordering.

---

**Approve**, nothing blocking. #1–#3 are one failure surface: truncation, bolding and API errors all funnel into `(False, 100)` with no signal — and that surface is what `status="stable"` rests on, validated once against one judge, in a run where the failure path never fired. Worth landing #1 and #2 here (#2 is two characters; #1 is ~15 lines + tests). #3–#5 are one-line docstring additions in files already in the diff.

jack-scitix-ai:
Thanks for the thorough pass — all five addressed:
**1 (bug) — unparseable judge replies are no longer graded.** `parse_judge` now returns `(correct, confidence, parsed)`; `feedback` records `judge_parsed` on `JudgeFeedback`, and `report()` builds the correct/confidence arrays only from parsed replies. Unparsed replies stay in `n` (still counted incorrect — hence no accuracy change) but are dropped from the calibration arrays, mirroring upstream's `None` on judge failure, so they no longer inflate `calibration_error`. Added a `judge_unparsed` count to the report so the discrepancy isn't silent. Also fixed the mis-cited "no confidence score" note: that default now only applies on the confidence-extraction path (the candidate's score), not the judge-failure path.
**2 (nit) — regex bold tolerance.** `_CORRECT_RE` / `_CONFIDENCE_RE` now use `\bcorrect\**\s*:\s*\**\s*…`, so `**correct:**` / `**correct**:` and bold values parse. Added tests.
**3 (nit)** — documented the judge's token budget in the docstring (upstream `max_completion_tokens=4096`, set via the `grader` config).
**4 (nit)** — noted in the dataset loader docstring that `cais/hle` is gated (accept the gate + `HF_TOKEN`).
**5 (question)** — added an ordering caveat: `text_only` runs at task construction, *after* session-level sampling, so budgets should be sized against the full set.

ethan-scitix:
Re-reviewed at `7be30674`. All 10 items from the previous three rounds are resolved, and I re-verified the core claims: `aggregate_metrics` reproduces upstream `dump_metrics` arithmetic exactly on a 2158-sample synthetic run, the judge regexes behave correctly on 16 reply shapes, 26 HLE tests + the full 2418-test unit suite pass, `check_tasks` / `check_datasets` are clean, `sync_meta_index.py` regenerates with zero drift, all three pinned URLs resolve, and the branch is 0 commits behind `main`. The unparsed-judge fix does not move the validation numbers (12.65 either way), so the 12.14 / 3.61 table stands.

Four things below; the first two I'd like fixed before merge.

**Bug: the `*` convention is stated backwards, and it ships in `index.json`.**
`sieval/tasks/hle_0shot_gen.py:11-12` and the `reference_impl.notes` at `sieval/tasks/hle_0shot_gen.py:104-105` both say the *full set* is the asterisked variant. It's the other way round: lastexam.ai's footnote reads "\*Model is not multi-modal, evaluated on text-only subset." — the full set is the unmarked default and *text-only* is what carries the asterisk. The notes text is embedded in `sieval/meta/index.json`, so this needs a regen after the fix. Note this also removes the only stated justification for `text_only: bool = True` being the default; the default may well still be right for the text-only models we actually run, but it should say so explicitly rather than lean on an inverted claim. (I checked the official leaderboard convention only — if you have a specific technical report using the opposite convention, cite it and I'll withdraw this.)

**Bug: `report()` doesn't record which subset was evaluated.**
The keys at `sieval/tasks/hle_0shot_gen.py:266-276` are `score / accuracy / confidence_interval / calibration_error / n / n_graded / fails / judge_unparsed / truncated` — nothing distinguishes a 2158-row text-only run from a 2500-row full-set run, and those two numbers aren't comparable. Same argument that motivated `truncated`: make the headline self-documenting. An `image_questions` count (or a `subset` string) would do it.

**Question: should `text_only` live on the dataset instead of the task?**
Upstream's `run_model_predictions.py` @ 26dca2e has no subset flag at all — its entire modality handling is the per-sample `if question['image']` branch that `preprocess` already mirrors. So `text_only` + `_select_text_only` is a structural divergence from upstream that isn't declared in "Deviations from upstream". Meanwhile population reshaping is what the dataset layer already does (`operations:` → slice / shuffle / stratified_sample), and `session.py` applies `datasets.*.args` at `:1204-1210` *before* `operations:` at `:1233`. Moving the flag to `HLEDataset.load(..., *, text_only=True)` would keep the task isomorphic to upstream, delete `_select_text_only` and the `_hf_dict` reconstruction, and dissolve the ordering caveat at `sieval/tasks/hle_0shot_gen.py:14-17` structurally rather than documenting around it. Costs: no existing loader takes a custom load arg, and the signature must capture it explicitly so it can't leak into `load_dataset`. Your call — this is a design preference, not a defect, but if you keep it task-side please declare the divergence in the deviations list.

**Nits:**

- `sieval/community/hle.py:75-76`: a JSON-shaped judge reply (`"correct": "yes"`) still misses. Adding `["']*` on both sides of the colon in both patterns covers it — the markdown-bold case is already handled, this is the same class.
- `sieval/tasks/hle_0shot_gen.py:167`: `type(dataset)(_hf_dict=filtered)` is the only production use of the private ctor param in the tree; core's own idiom is `_clone_with_new_dict`. They're behaviorally identical today (no dataset subclass overrides `__init__` or carries state beyond `_dataset_dict`), so a one-line comment saying why the ctor path is used is enough — moot if the flag moves to the dataset.
- `sieval/community/hle.py:164`: `calibration_error = 0.0` under the `< BETA` guard conflates "not computed" with "perfectly calibrated". The docstring explains it, but a consumer reading `report.json` can't tell. Consider `None`.
- `truncated` and `judge_unparsed` are accumulated over `finals` only while `n` spans `finals + fails` — correct as written, worth a half-sentence so nobody reads them as rates over `n`.

jack-scitix-ai:
Thanks — all four addressed:

**1 (bug) — asterisk convention.** Fixed in both the module docstring and `reference_impl.notes`: the **text-only** subset is the asterisked variant (it excludes the multimodal questions); the full text+image set is the unmarked headline. Also restored the justification for the `text_only=True` default that the flip removed: the full set requires a vision-capable candidate *and* judge, so text-only is the broadly reproducible default.

**2 (bug) — subset now recorded.** `report()` emits a `subset` field (`text_only` | `full`) so a text-only run is distinguishable from a full-set run from the report alone.

**3 (question) — kept task-side, divergence declared.** Added a bullet to "Deviations from upstream" noting `text_only` / `_select_text_only` is a sieval addition (upstream has no subset flag). I considered moving it to `HLEDataset.load(..., *, text_only=True)` — there's precedent (`mmmlu` filters by locale/subject in `load()`) — but kept it task-side for two reasons:
- A *dataset* defaulting `text_only=True` would hide the image rows by default, which is surprising for a dataset (siblings default to the full set, e.g. `mmmlu`'s `config="all"`). Text-only reads more naturally as an eval-protocol choice on the task.
- Upstream has no subset flag anywhere (task *or* dataset), so relocating it wouldn't make it match upstream — it only moves the sieval-added divergence. The current split keeps the dataset loader maximally upstream-aligned (a pure full-set load), with subset selection as the task's concern.
Happy to move it if you feel strongly — it's a small change — but that's the reasoning for leaving it.

**4 (nits).**
- Judge regex now tolerates quotes as well as bold (`\bcorrect[\s"'*]*:…`), so a JSON-shaped reply like `{"correct": "yes"}` parses. Added a test.
- `_select_text_only` now uses core's `_clone_with_new_dict` idiom instead of the private `_hf_dict` ctor param.
- `calibration_error` is `None` (not `0.0`) under the `< BETA` guard, so "not computed" isn't conflated with perfect calibration. Return types widened accordingly; tests assert `is None`.
- Added a half-sentence noting `truncated` / `judge_unparsed` are counts over `finals`, not rates over `n` (which also spans `fails`).

ethan-scitix:
Pushed three commits as a maintainer edit and rebased onto latest `main` (`96e67d16`). Your seven commits kept their content but have new SHAs, so locally you'll want `git fetch && git reset --hard origin/feat/hle` rather than a pull.

### 1. `text_only` moved to the dataset layer — I'm reversing a call I'd explicitly left to you

Last round I tagged this `Question:`, wrote *"Your call — this is a design preference, not a defect"*, and offered a fallback: keep it task-side, declare the divergence. You did exactly that. So this isn't something you failed to address — I'm reopening it, and the reason is two things I only established afterwards:

- **The cross-layer private access was a product of the task-side choice.** `dataset._clone_with_new_dict(...)` was the only place in `sieval/tasks/` or `sieval/datasets/` reaching into a core private. Moving the filter into `load()` removes it with no `core/` change.
- **The one cost I cited against moving was wrong.** I wrote "no existing loader takes a custom load arg" — `sieval/datasets/mmmlu.py:316-324` does precisely that, five explicit params ahead of `**kwargs`. You cited mmmlu as precedent and you were right; I should have checked before listing it as a cost.

For the record, of your two arguments for keeping it task-side: the one I don't think holds is *"upstream has no subset flag anywhere, so relocating wouldn't make it match upstream"* — the goal was task-to-upstream isomorphism, not dataset-to-upstream parity. The other one (a dataset hiding image rows by default is surprising) is fair, and the loader docstring now states the default and its reason explicitly.

Four things collapse into the one change: `_select_text_only` (29 lines), the `_clone_with_new_dict` access, the "Deviations from upstream" entry for subset selection, and the ordering caveat — the last dissolved structurally, since `datasets.*.args` is applied before `operations:`, so a sampling budget now covers exactly the set that gets graded. `report()` still emits `subset`, reading the new public `HLEDataset.text_only`.

### 2. `cast_column` was failing open

Not part of the refactor; found while touching the file. Given a column name absent from the table, `cast_column` doesn't raise — it silently adds an empty one:

```
before: ['id', 'question', 'image', 'answer']
after : ['id', 'question', 'image', 'answer', 'image_preview']
```

So an upstream rename of `image_preview` / `rationale_image` would drop the Pillow guard with no signal, surfacing only as an ImportError in an environment without Pillow. Both columns are now asserted present before the cast.

Worth recording *why* the cast exists at all, which the docstring now does: upstream reads the split with `load_dataset(...).to_dict()`, a bulk export that skips the per-example decoder — hence its Pillow-free `requirements.txt`. sieval materializes rows one at a time via `test_set[sample_id]`, which does decode. Without that note the cast reads as gratuitous divergence and someone will eventually delete it.

### 3. `calib_err`'s single-bin band

The `None` sentinel you added covers `< BETA`. But at `BETA..2*BETA` graded attempts, `calib_err` builds exactly one bin and its `range(len(bins) - 1)` loop never runs, so it returns a structural `0.0` for any input:

```
n_graded= 99 -> None      n_graded=150 -> 0.0
n_graded=100 -> 0.0       n_graded=200 -> 7.0
```

That's upstream's dropped-last-bin behaviour and unreachable on the full ~2.5k set, so I left the threshold alone and documented it instead. Also trimmed the module and `parse_judge` docstrings — the upstream-drop semantics were stated three times over.

### Verification

2459 unit tests pass (31 HLE, up from 27), `ty` and `ruff` clean, full preflight 19/19 including `check_imports` and `check_meta_index_sync`, `index.json` regenerated. Prompts remain byte-identical to upstream@`26dca2e` and the `aggregate_metrics` arithmetic is untouched, so the 12.14 / 3.61 validation table still stands.

Push back on any of it — especially #1, since I'm overriding a decision I had said was yours.

Issue #27, opened 2026-07-02
[RFC]: Recipe capability layer — split model-type capabilities out of hardware profiles


ethan-scitix:
## Motivation.

The inference recipe system organizes serving params as a single flat bag per `profiles[hardware][precision][framework]`, matched by **family + size only** (no base-vs-instruct awareness). Two problems follow:

1. **Capability params live in the wrong place.** `tool_call_parser`, `enable_auto_tool_choice`, `reasoning_parser` are baked into *every* hardware×precision profile. They are meaningless for a base-model logprob eval, but a base checkpoint of the same family/size resolves to the same profile and inherits them anyway.

2. **There is no channel for an eval to state what it needs from serving.** Two live instances already exist:
   - **Input logprobs / prefix cache.** PPL tasks that read the *input* prompt's logprobs (`echo=True`) require the serving framework's prefix cache to be **off**. Verified backend behavior:
     - **sglang** (radix cache, on by default): on a cache hit it truncates `input_token_logprobs` to `prompt_tokens − cached_tokens` and does not recompute cached positions — silently returning a partial, non-reproducible token set. Needs `--disable-radix-cache`.
     - **vLLM V1**: `prompt_logprobs`/echo is mutually exclusive with automatic prefix caching — the request *errors* rather than truncating. Needs `--no-enable-prefix-caching`.
     - Output-only logprob tasks (e.g. CMMLU, `echo=False`, first-output-token top-k) are unaffected and benefit from the cache. So the trigger is **"reads input logprobs" (`echo=True`)**, finer than `eval_mode`.
   - **Top-k breadth.** CMMLU needs the server launched with `--max-logprobs 100` on vLLM (default 20; sglang serves 100 out of the box) so all option tokens appear in the top-k. Today this relies on an operator reading the task docstring.

Neither problem is served by the current recipe structure; both point at the same missing dimension — **model capability**.

## Proposed Change.

Reorganize recipes around two independent param sources, kept separate by design (different owners, lifetimes, vocabularies).

**Mechanism A — model-type capability profiles (recipe-authored).** Split the flat `profiles` bag into a `hardware` layer (hw × precision × framework: dtype, memory, parallelism, context) and a `capabilities` layer (model-type × framework: parsers, tool-choice, and model-intrinsic serving recommendations such as gpt-oss's prefix-cache-off). A recipe entry serves both the base and instruct checkpoint of a family/size; `capabilities` holds both variants and resolution picks one.

```yaml
qwen2.5-7b:
[...]

Issue #24, opened 2026-07-01
[RFC]: Resume version-compatibility gate + per-record version provenance


ethan-scitix:
## Motivation.

Precise reproducibility is a product contract in SiEval, but `--resume` does **not**
currently gate on the sieval version.

Today the strict-match resume check compares persisted config bodies byte-for-byte
(`effective_config.yaml` / `infer_plans.yaml`, in `cli/leaderboard/session.py`
`_persist_yaml_with_strict_resume`). The sieval version is written to `meta.json`
(`core/tasks/saver.py` `write_run_meta`) purely as audit metadata — **nothing ever
reads it back**. The only way a version change affects resume today is indirectly:
if a new version re-resolves `infer_plans.yaml` to a different body, the config
strict-match trips. A change that does not alter the config/plan (e.g. a scoring
semantics fix, a record-format tweak, or a model-interface redesign) is invisible
to resume.

Consequences:

- A run started under version A can be resumed under version B with no check, as
  long as the config YAML is identical. Results computed under two different
  versions get silently merged into one `report.json`.
- Worse, resume across a **format-incompatible** boundary is not refused.
- There is no mechanism to safely defer interface-breaking changes: e.g. a future
  redesign of the `Model` logprobs interface has nothing to rely on to refuse
  resume across the incompatible bump.

## Proposed Change.

Two complementary mechanisms. They are independent of the sglang native-backend
work (see Non-goals).

### A. Resume version gate (go/no-go before running)

On resume, read a format-stable version anchor from `meta.json` and decide:

```
v_run = meta.json version (already written today)
v_cur = current __version__

1. v_run == v_cur (exact string)                 -> RESUME (silent)   # same build, incl. dev
2. unparseable / 0.0.0                            -> REJECT
3. either is a local/dev build                    -> REJECT           # unpinnable, non-exact
   (packaging: v.local is not None or v.is_devrelease)
4. compat_key(v_run) != compat_key(v_cur)         -> REJECT
5. else (same series, non-exact)                  -> RESUME + provenance (see B)

compat_key(v) = (v.major,)            if v.major >= 1     # >=1.0: major is the break axis
              = (v.major, v.minor)    if v.major == 0     # <1.0 : minor is the break axis
```

- EM always passes first, so resuming your own run (including today's untagged dev
  builds, `local != None`) is never blocked.
- Straddling the 1.0 boundary (e.g. `0.9.x` vs `1.0.x`) yields different keys and
  is correctly rejected.
[...]

PR #45, opened 2026-07-28
feat(models): rebuild the provider-neutral model binding plane (RFC #25)

jack-scitix-ai:
## Type

## Summary

- Implements RFC #25 Revision 2 PR 1: a keyed capability catalog, normalized `TaskRequirements`, pure deterministic `reconcile()`, and immutable `RuntimeBindingPlan`.
- Separates protocol (`Dialect`), server semantics (`Engine`), desired state (`DeploymentPlan`), realized state (`Deployment` + `ServingFacts`), and connection ownership (`ConnectionPool`).
- Makes normalized task input requirements the sole task-side authority consumed by `derive_model_type`; YAML `type:` is a checked assertion when task evidence exists.
- Runs request defaults, domain checks, capability admission, and the residual consume/reject/no-op/passthrough leaf audit before limiter acquisition or network I/O, without a condition AST or constraint solver.
- Registers the seven-dialect symbol space while PR 1 binds only `openai_chat` and `openai_completions`; later adapters remain isolated follow-up PRs, and `SglangGenModel` remains a temporary legacy facade.
- Preserves shared clients and hierarchical quotas through pool-backed `with_args()`, `with_dialect()`, and compatibility rebinding; existing task `agenerate()`/`alogprobs()` call sites remain supported.
- Fixes unknown echoed-token boundaries, extension-body routing, streamed tool-call reconstruction, singular-channel `n > 1` validation, and fabricated provider identity; echoed `top_logprobs` remain completion-only.
- Breaking migration: cross-kind `as_type()` and derived YAML kind conversion are removed. Use a separate root for another input modality, or `with_dialect()` for a compatible dialect on the same pool; unsupported leaves and explicit type mismatches now fail loudly.

## Related Issues

Refs #24, #25, #27, #47, #58, #59, #60, #61, #62

### Automated

### Required (all PRs)

### If: Breaking Change


ethan-scitix (commented):
## Verdict

**The wire extraction is correct — I verified it. The skeleton it lands on needs rework before merge.**

Splitting those two apart, because the second conclusion should not be read as doubt about the first.

### What I verified empirically

- `pdm run pytest tests/unit tests/integration` → **2509 passed**; `ty check sieval` clean; `ruff` clean apart from the pre-existing generated `sieval/_version.py`.
- Coverage on changed modules: `ir.py`/`sglang.py`/`capabilities.py` 100%, both OpenAI transports 98%, `model.py` 95% — meets the `core/CLAUDE.md` gate. (I did not re-run mutmut.)
- **A/B'd the legacy backends against the new transports on identical wire payloads.** I extracted `origin/main:sglang_gen_model.py` and `origin/main:gen_model.py` and ran both implementations over the same realistic sglang `/generate` and vLLM `/v1/completions` echo responses:

  ```
  sglang / vLLM × echo on/off — four paths:
    texts / finish_reasons / logprobs_tokens / logprobs / usage  →  bit-identical
  ```

  The PPL/CLP-critical fields survive the rewrite exactly. The only divergence is `top_logprobs` under `echo=True` (legacy returned prompt+completion positions, new returns completion-only) — no in-tree task reads it, every CLP task uses `echo=False` and every PPL task uses `logprobs=0`, and the new behaviour actually matches what `choice_scores_from_top_logprobs` documents (`"top_logprobs[0]` (the next-token distribution)"), which the legacy echo layout violated. That's a latent-bug fix, and it's annotated at `sieval/core/models/model.py:509-513`. It should be in the PR summary too, since it is user-visible for anyone building on `alogprobs(echo=True, logprobs>0)`.

So the streaming accumulation, echo split, sglang triple parsing, token-text normalisation and radix guard are all faithful transpositions. That's the expensive part, and it's done.

---

## Why the skeleton needs rework

**The capability abstraction was built, but neither decision point was migrated onto it.**

`Task.model_type` + `isinstance` (`sieval/core/tasks/task.py:79`) and the config-layer `type:` binary are both untouched, while `Task.requires` (`task.py:60`) has **zero users** in the entire repo. So the PR ships two parallel systems where the new one decides nothing. Everything below is a symptom of that, not an independent problem:

**1. The IR silently ignores four Request features — the exact failure mode this PR exists to remove.** `transports/openai_completions.py:92` and `transports/sglang.py:159` read neither `req.session_id`, `req.server_tools`, `req.suffix`, nor any `req.reasoning` axis. `openai_chat.py:110` *does* reject `session_id`, so siblings disagree on one field. Verified:

```
Request(input="p", session_id="resp_abc", server_tools=(ServerToolSpec(type="web_search"),),
        suffix="TAIL", reasoning=ReasoningParams(effort="high", budget_tokens=4096))

completions wire body: {'model': 'm', 'prompt': 'p', 'stream': False}
sglang wire body:      {'text': 'p', 'sampling_params': {}}
chat:                  CapabilityError: ...does not support stateful session_id
```

`assert_capability` only fires where a caller remembers to call it, and nothing calls it for these.

**2. The capability catalog contradicts the implementation, and rejects features that work.** `OpenAIChatTransport` lowers `reasoning_effort` (`openai_chat.py:156`) and lifts reasoning text, but declares neither `Reasoning` nor `ReasoningEffort`:

```
reasoning_effort actually sent on the wire: True high
reasoning actually lifted: ReasoningOutput(text='deep', ...)
assert_capability(Reasoning)       -> REJECTED: OpenAIChatTransport does not support: Reasoning
assert_capability(ReasoningEffort) -> REJECTED: OpenAIChatTransport does not support: ReasoningEffort
```

That's fail-loud in the wrong direction. **15 of the 24 `Capability` members are declared by no transport at all** and are unreachable from any `Request` field any transport consults.

**3. `Capability` is a `Flag` but capabilities are stored in a `frozenset` — a composite is always reported missing.** `capabilities.py:13` + `model.py:237`. Verified false positive on two *supported* caps:

```python
c.assert_capability(Capability.Chat | Capability.FunctionCalling)
# both individually supported: True True
# CapabilityError: OpenAIChatTransport does not support: Chat|FunctionCalling
```

Declaring `Flag` invites `|`, and `Task.requires: ClassVar[frozenset[Capability]]` makes `frozenset({Capability.Chat | Capability.SampledLogprobs})` a natural authoring mistake that silently rejects a working model. Either a plain `Enum`, or store a single bitmask and test `caps & required == required`.

**4. Removing cross-kind derivation dropped quota sharing with no replacement, and left a footgun.** `as_type` shared one client and one limiter across kinds; the docs now say "define a separate base model instead" (`docs/guide/configuration.md:31`), which means a second `AsyncOpenAI` client and a *separate* pool — losing the hierarchical concurrency `core/CLAUDE.md` treats as an engine invariant. `Model.__init__` already takes `transport=` (`model.py:133`), which is the seam that would preserve sharing, but `with_args` doesn't handle it:

```python
d = base.with_args(transport=OpenAICompletionsTransport(...))
# derived _transport is base transport? True
# extra_wire_params: {'transport': 'OpenAICompletionsTransport'}   → forwarded to the API
```

Please don't remove a feature before its replacement exists — `as_type` can stay until the revision provides a transport-swap that keeps the pool.

**5. `openai_completions.py:186` — when the server omits `usage`, echoed prompt tokens are labelled as sampled output.** `boundary = usage.input_tokens if usage is not None else 0`. Verified on a vLLM-shaped echo response with `usage=None`:

```
input_scoring.token_logprobs: ()
logprobs (claimed SAMPLED output): 'Q', ' is', ' it', '?\n', ' A', ' B'   # 5 are prompt
```

No shipped task is affected (the legacy bridge re-concatenates), but `arun`/`Response` is the documented forward path and a new consumer gets prompt tokens as output with no signal. Contrast the sibling: `SglangTransport._guard_radix_cache` *raises* when it can't verify the echoed length rather than scoring silently. Same stance here — refuse the split when the boundary is unknowable.

**6. Two different "what kind is this model?" implementations in one PR.** `session.py:1062` uses `Capability.Chat in base_model.capabilities`; `task.py:79` uses `isinstance`. Pick one.

---

## Two claims in the code that don't hold

Worth correcting before the revision builds on them:

- **`top_k` is not a "vLLM extension upstream OpenAI rejects" that warrants a first-class field** (`openai_chat.py:131`, `openai_completions.py:108`). `leaderboards/qwen3_proxy_5min_202606.yaml:78` already passes it through `extra_body`. So promoting `top_k_sampling` to `SamplingParams` wasn't required — the repo's own configs route it through passthrough. It's a good illustration of the promotion rule the revision needs.
- **"sglang's `/v1/completions` rejects `echo=True` together with `logprobs`"** (`sglang_gen_model.py:3`, `transports/sglang.py:3`) has **no supporting record anywhere in `docs/designs/`** — I searched. It is the sole stated justification for `SglangGenModel` existing. Meanwhile `docs/superpowers/specs/2026-07-01-recipe-capability-layer-design.md` documents a *different* constraint under a "Verified backend behavior" heading: input logprobs × prefix cache, on **both** engines (sglang truncates silently, vLLM V1 errors). The rejection claim looks like a possible mis-attribution of that to the protocol layer. That same section also states vLLM V1 **has** `prompt_logprobs`, which makes `openai_completions.py:6`'s "missing native scoring endpoint" wrong and the whole `echo` workaround (plus its boundary split) potentially unnecessary.

**Two probes gate the revision's scope, because each can delete an entire code path.** Please run these before reworking:

1. Does sglang's `/v1/completions` actually accept `echo=True` + `logprobs`? If yes, the native `/generate` frontend may be removable — along with token-text normalisation, `--skip-tokenizer-init` handling and the `max_new_tokens=0` clamp.
2. Is `prompt_logprobs` available on the deployed vLLM? If yes, the `echo` workaround and the `usage.input_tokens` split both go away.

---

## Direction for the RFC #25 revision

The IR *concept* is right — input declares capability, the frontend translates to each vendor's fields. What went wrong is the **population order**: the vocabulary was filled from vendor docs instead of growing from frontends that exist. So the revision keeps the idea and changes how a field earns its way in.

1. **Split frontend from engine.** Capability is `f(frontend, engine, instance)`. `vLLM` and `sglang` both serve `openai_completions`; `sglang` also serves native `/generate` — so `SglangGenModel` fusing "sglang engine" + "native protocol" into one class is the root of several oddities here, including the radix guard protecting only one of the two sglang paths. This is the one thing that cannot be deferred: it determines the class structure, and #47's engine-level constraints have nowhere to live without it.

2. **Two-level capability with domains.** First level = a concern that ≥2 providers express differently (`sampling`, `scoring`, `reasoning`, `tools`, `structured_output`, `caching`, `session`, `modality`). Second level = the fields inside it, each carrying a **domain**, not just a boolean — `effort` is a different enum per vendor, and top-k breadth is "supported up to N", not "supported". First-level groups should map 1:1 onto `Request` sub-records so there is one structure, not two.

3. **Derive the gate structurally, drop the central catalog.** Have each frontend declare which `Request` fields its `lower()` consumes, and assert in `Model.arun` that no field set to a non-default value went unconsumed. `Request` is a frozen dataclass with defaults, so this is computable — a ~10-line prototype catches all four silently-dropped fields across all three transports with no `Capability` enum involved. Then a missing or wrong enum member can never cause a silent drop, and the field-promotion rule becomes enforceable: promote only when ≥2 frontends lower it to *different* wire names (translation coupling), otherwise leave it in `extra_wire_params`. That's `CLAUDE.md`'s "extract on coupling, not on call count".

4. **Keep the output side closed — do not make it polymorphic.** `Response` carries `@sieval_record` and is the persisted schema coupled to RFC #24's resume gate; an open subclass hierarchy makes the on-disk shape depend on which subclass showed up, which cannot be versioned. Instead: one optional field per modality, each typed as a record (the existing `InputScoringResult` / `ReasoningOutput` / `UsageStats` pattern), additive only, never renamed or retyped. That's also how the currently-missing channels land — `embedding`, `media` — without polymorphism.

5. **`ChatModel` / `GenModel` become deprecated aliases for the OpenAI chat-completions and completions bindings.** After this PR they're 26 and 22 lines that only pick a transport, which `Model(transport=...)` already does; the `isinstance` kind check is the only thing keeping them alive. Note the sequencing constraint: if they become factory/partial aliases, `isinstance(m, ChatModel)` raises `TypeError: isinstance() arg 2 must be a type` — loudly, not silently — so the alias change and the `requires`-replaces-`model_type` migration must land together. Also drop the `Model[TModelInput]` generic and the `openai.types.chat.ChatCompletionMessageParam` leak: a provider-agnostic layer shouldn't pin its Model signature to one vendor's SDK type, and it currently disagrees with `Request.input`'s own `str | list[dict[str, Any]]`.

6. **Naming.** `Transport` collides with httpx (`tests/unit/scripts/test_check_preflight.py:1025` literally has `transport=transport`), and `backend` is already triple-booked — `sieval/infer/backends/` (launch-side translators), and this PR's own docstrings (`core/models/__init__.py:1` "Model backends", `chat_model.py` "the backend selector"). Suggest **`Frontend`**: the PR's docstrings already use "provider frontend" 10 times in prose and it has no collision. That frees `backend` to mean the engine only.

7. **Reject serving-axis capabilities in `requires` rather than accepting and ignoring them.** `scoring.topk_breadth` and `caching.prefix_cache` are enforced by #47, not this revision. Better that nobody can write the declaration than that someone writes one which silently does nothing — that's the same failure as items 1–2 above, and it would be ours.

---

## Also

- **Config-layer `type: chat|gen` should stay as-is in the revision.** It's hard-coded in 6 places (`task.py:58`, `tasks/meta.py:73,143`, `cli/validation.py:151`, `session.py:64,1056`) and feeds `sieval/meta/index.json` (35 `model_type` occurrences, `schema_version: 1`), so changing it triggers the meta-drift preflight and wants its own RFC. Just say explicitly in the revision that it's known-remaining — otherwise whoever adds the Anthropic frontend will assume the groundwork is done.
- **`docs/superpowers/specs/2026-07-01-recipe-capability-layer-design.md` is not tracked by git.** It's the only record of the verified backend behaviours, it's explicitly related to #21/#24/#25, and it designs the launch-flag half of the same vocabulary. Please commit it — it should probably be merged into the RFC #25 revision.
- Two small ones: `_SAMPLING_PARAM_MAP` in `sglang.py:47` is now mostly dead (7 of 9 keys are first-class `SamplingParams` fields popped before they can reach it, so only `min_p` and `repetition_penalty` are reachable); and `sglang.py:186` loses the legacy `max(max_tokens, 1)` clamp on the `echo=False` path, so `alogprobs(echo=False, max_tokens=0)` now forwards `max_new_tokens=0`, which sglang rejects.
- The bundled `[tool.mutmut]` fix and the two `collect_ignore` guards are fine as-is — disclosed with clear rationale. I confirmed `also_copy = [..., "scripts"]` is *not* redundant with the `tests/unit/scripts` skip, because `tests/unit/core/tasks/test_meta_pilot.py` also reaches into `scripts/`.
- Title should be `feat(models)!:` — this removes public API and invalidates previously-working configs, and `CHANGELOG.md` is generated from commits at release time.

---

## Filed alongside this review

- **#47** — RFC for the capability constraint layer (the cross-group constraints this PR's `Capability` set cannot express: `input_logprobs ⊥ prefix_cache`, top-k breadth floors, protocol-mandatory fields). Depends on items 1, 2 and a `verify_instance()` seam above.
- **#46** — `fix(scripts): enforce relative-import scope in check_layer_imports`. `CLAUDE.md` `## Import Policy` and `CONTRIBUTING.md:59` both say "same package: relative; cross-package: absolute", but nothing enforced the second half. The 7 `from ..capabilities` / `from ..ir` / `from ..exceptions` imports in `transports/*.py` are currently the only `..` imports in the tree and every other nested subpackage uses the absolute form (`task.py:12`, `session.py:27`). Once #46 lands, `python scripts/check_layer_imports.py` will flag them with the absolute form to use — worth rebasing onto it and converting them either way.

Happy to review the RFC #25 revision before implementation — the class-structure decision in item 1 is the one worth agreeing on first, since everything else follows from it.

jack-scitix-ai:
Thanks for the detailed review. I kept the wire extraction you A/B verified and rebuilt the capability/class skeleton around it according to the RFC #25 Revision 2 agreement.

Here is how the six blocking findings map to the new implementation.

### 1. Request fields can no longer be silently dropped

The canonical request is now structured into input, sampling, scoring, reasoning, tools, structured output, session, scheduling, and dialect-specific options.

Each active non-default request leaf must receive one explicit decision:

- consumed;
- rejected;
- accepted as a no-op;
- passed through to a named wire container.

`Model.arun()` checks the leaf audit before limiter acquisition or network I/O. `suffix`, session state, hosted tools, and reasoning fields therefore cannot silently disappear.

This remains a residual structural check—not a condition AST, accepted-region algebra, subset prover, or general constraint solver.

### 2. Capability declarations no longer act as a contradictory gate

The keyed capability registry is the declaration vocabulary agreed in Revision 2. Executable support comes from joining:

- task intent;
- dialect binding;
- model declaration;
- engine/deployment serving outcome.

The per-call leaf audit is independent of that catalog and prevents a declaration from saying “supported” while `lower()` ignores the field.

Provider-specific values that are not canonical request semantics remain under `DialectOptions` and must name their passthrough destination.

### 3. The `Flag` + `frozenset` bug is removed

The deprecated compatibility `Capability` namespace is now a plain `Enum`. Composite values such as:

```python
Capability.Chat | Capability.FunctionCalling
```

are no longer representable.

Canonical code uses keyed capability records and immutable reconciled plans instead of set subtraction over enum combinations.

### 4. Cross-protocol derivation has a quota-preserving replacement

`ConnectionPool` owns the connection and shared limiter. Derived models retain that pool, and `with_args()` may add one model-local sub-quota without creating another client or losing the parent cap.

`with_dialect()` rebinds against an already reconciled `RuntimeBindingPlan` while retaining the same deployment, pool, client, and limiter hierarchy. It first verifies connection family, deployment fingerprint, root deployment, and requested model identity.

`with_args(transport=...)` can no longer become an API parameter: binding-resource names such as `transport`, `dialect`, `pool`, credentials, and endpoints are explicitly rejected.

The old cross-kind `as_type()` path was removed only after this replacement existed. `as_compat_type()` remains a one-cycle shape adapter for legacy task constructors; it does not perform an unreconciled dialect conversion.

### 5. An unknown echo boundary now fails loudly

OpenAI Completions input scoring no longer falls back to a boundary of zero.

When echoed prompt scoring is requested, the dialect requires consistent usage data and validates that:

- the prompt-token boundary is positive;
- it does not exceed the returned logprob positions;
- the completion-token count agrees with the remaining positions.

Missing or contradictory usage raises an output-contract error instead of labelling prompt tokens as sampled output.

The user-visible `echo=True` behavior noted in the review is also explicit: sampled `top_logprobs` are completion-only, while prompt scoring is represented separately by `input_scoring`.

### 6. Model kind now has one task-side authority

`TaskRequirements.input` is the normalized task-side source consumed by the shared `derive_model_type()` resolver.

Legacy `Task.model_type` is mechanically projected into that requirement for schema-version-1 compatibility, but runtime resolution no longer independently reads it or uses wrapper `isinstance()` checks as a competing authority.

When task evidence exists, YAML `type:` is a checked assertion. A conflicting explicit value fails during prelaunch reconciliation instead of overriding the task requirement.

### Other points from the review

- The provider frontend is now named `Dialect`; `Engine` is reserved for serving semantics.
- `DeploymentPlan` remains desired state, while `Deployment` plus `ServingFacts` represents the realized service.
- `reconcile()` is a pure deterministic join producing immutable `RuntimeBindingPlan` values. It performs no I/O and contains no general-purpose constraint solver.
- `Response` remains a closed additive persisted record instead of a polymorphic hierarchy.
- `top_k` is routed through the SDK’s `extra_body` wire container rather than treated as an upstream OpenAI keyword.
- Engine/instance constraints such as prefix-cache compatibility remain behind the `Engine.verify_deployment()` seam for #47.
- `openai_chat` and `openai_completions` are the only executable canonical dialects in this PR. The remaining dialect IDs are reserved for the adapter sequence agreed in #25.
- `sglang_native` is not claimed as implemented here. The `/generate` implementation is isolated behind the explicit `sglang_legacy` compatibility facade, and the SGLang probes remain a prerequisite for PR 5.
- The SGLang input-scoring `max_new_tokens >= 1` clamp has been restored.

### Verification

The branch is rebased onto the latest `main`.

The previous PR body’s mutation result is intentionally not reused because it does not cover this rebuilt implementation.

Could you please re-review the rebuilt binding plane, particularly the `Dialect` / `Engine` / `Deployment` separation, `reconcile()` output, and the pre-I/O request-leaf audit?

ethan-scitix (commented):
## Verdict

**Five of the six blocking findings from the previous review are fixed, and I verified each by running it. Requesting changes on five items — one is the previous review's finding #1, still unfixed on the third transport it named.**

The `Dialect`/`Engine`/`Deployment` separation, `reconcile()`'s output shape, the pre-I/O leaf audit on both OpenAI dialects, and the echo-boundary hardening are solid. Remaining work is bounded and mostly mechanical.

**Verified:** 4840 tests pass; `ruff` + `ty` clean; coverage 98% (every module ≥96%). You correctly didn't reuse the old mutation number — I ran **7 targeted mutations** against the headline invariants (audit missing-leaf, both echo-boundary assertions, capability gate, `with_args` binding-key rejection, `derive_model_type`'s checked assertion, sglang clamp) and **all 7 were caught**. Also re-ran all four prior reproductions and trial-rebased onto current `main`.

## Prior comments

| # | Prior finding | Status |
|---|---|---|
| 1 | IR silently ignores 4 Request features | **Partial** — fixed on both OpenAI dialects, unaddressed on sglang (**B1**) |
| 2 | Catalog contradicts implementation; 15/24 caps unreachable | **Resolved** — chat declares + lowers `Reasoning`/`ReasoningEffort`; 12 keyed caps, 11 reachable (**N2**) |
| 3 | `Flag` + `frozenset` composite false positive | **Resolved** — plain `Enum`, composite raises `TypeError` |
| 4 | Quota sharing dropped with no replacement | **Resolved in code** (`with_dialect` preserves pool/connection/limiter), **docs regressed** (**B3**) |
| 5 | Unknown echo boundary labels prompt tokens as output | **Resolved** — verified and mutation-tested |
| 6 | Two model-kind authorities | **Resolved** — `TaskRequirements.input` is sole authority; `requires` now has 9 users (was zero) |

Secondary items **resolved**: `top_k` via `extra_body`; `Transport`→`Dialect` naming; `Model[TModelInput]` generic and the `openai.types.chat` leak dropped; #46 relative-import scope (zero `..` imports left); config-layer `type:` documented as a checked assertion.
Secondary items **unaddressed**: sglang `echo=False` clamp (**B4**); `_SAMPLING_PARAM_MAP` mostly dead (**N7**); untracked `recipe-capability-layer-design.md` and the missing breaking-change marker (**N8**).

**Confirmed compliant — not up for re-litigation:** the staged dialect rollout matches the #25 scope decision of 2026-08-06 (plane + two dialects = step 1 of 5; the other five are reserved symbols, `bindable=False`), so it isn't a unilateral deferral. Layer direction is clean — `infer/ → core.models` only, zero reverse imports, and no engine flag vocabulary (`disable_radix_cache`, `reasoning_parser`, …) leaked into `core/`; `reconcile()`'s `launch_patch` merges declared patches and never names a flag. `multimodal_input` is declared `Supported` and genuinely lowers.

---

## Blocking

### B1 — The leaf audit is bypassed on the sglang path; prior finding #1 still reproduces verbatim

`sglang_gen_model.py:128` overrides `Model.arun` and runs `validate_request_invariants` only:

```python
async def arun(self, req: Request) -> Response:
    validate_request_invariants(req)
    async with self._pool.acquire(self._limiter):
        return await self._legacy_transport.arun(req)
```

No `_apply_request_defaults`, `validate_runtime_binding_plan`, `RequestAudit`, `raise_rejections`, `prepare`/`finish`, `output_contract.validate`, or `_run_request_checks`. So all six of `suffix`, `session.previous_response_id`, `session.opaque_continuation`, `tools.hosted`, `reasoning.effort`, `reasoning.budget_tokens` still vanish into the previous review's exact body, `{'text': 'p', 'sampling_params': {}}`.

Not a quarantined facade: `session.py:1350-1354` routes **every** `engine: sglang` + `type: gen` root to `sglang_legacy` (`sglang_native` isn't bindable in PR 1), `session.py:2793` constructs it, and `_kwargs_to_request` builds these leaves from ordinary `args:` kwargs. End to end on a stubbed client:

```python
await m.agenerate("prompt", reasoning_effort="high", session_id="resp_abc", suffix="TAIL")
# POSTed: {'text': 'prompt', 'sampling_params': {'temperature': 0.0}}  — three fields gone, no error
```

Every sglang-served PPL/CLP eval runs this path. **Take the cheap fix now:** have `SglangGenModel.arun` hard-reject any active leaf it cannot lower. Making it a real `Dialect` is PR 5, gated on the SGLang probes — don't let B1 wait for that. A facade that accepts and discards is the failure mode this PR exists to remove.

### B2 — After rebasing onto `main`, 9 unit tests make live network calls

The rebase itself is trivial: `main` is 4 commits ahead, **one conflict, one hunk**, a docstring collision in `core/tasks/meta.py`. But the rebased tree fails **9 tests** in `tests/unit/tasks/test_inverse_ifeval_0shot_gen.py` with `openai.AuthenticationError: 401`. That file passes **20/20 on plain `main`** — it arrived via #91, after this branch's base.

Cause: this PR removes the `_agenerate_impl`/`_alogprobs_impl` seam (gone from production entirely), and that test's `_ScriptedChatModel` overrides it. The override becomes silently dead, the real `arun` runs, and the test **dials the network** — a 401 in CI, a real billed call if credentials are present.

Fix is mechanical and the template is in this PR: commit `416c2864` migrated `test_browsecomp_0shot_gen.py` the same way (`_build_default_transport()` returning a `HandlerTransport`). Two asks beyond the fix: **rebase again immediately before merge** (every task PR landing on `main` first hits the same trap), and consider a guard — e.g. `Model.__init_subclass__` raising if a subclass defines those names — so the next one fails loudly instead of dialing out.

Rest of the rebased tree is clean: preflight **20/20 PASS** (including `check_meta_index_sync`, so `requires` causes no meta-drift), suite 5019 passed / 9 failed, confined to that file.

### B3 — `as_type` should be deleted, not documented as deleted

Three records, one conclusion: #25's body (yours, 2026-07-01) line 54 says "**Drop `as_type`** … has no real consumer"; the 2026-07-28 review granted a stay of execution ("can stay **until** the revision provides a transport-swap that keeps the pool"); that condition is now met — `with_dialect()` exists and preserves pool, connection, and shared limiter (verified). And it has **zero production call sites** — only the two definitions, tests that exist solely to test it, and a doc line claiming it's gone.

Current state is wrong in both directions. The PR body says cross-kind `as_type()` was "removed"; it works fully and shares the pool (`GenModel.as_type(ChatModel)` → `openai_chat`, same pool/connection). And `configuration.md:95` tells users to "define a separate base model instead", costing them a second client and pool — the exact hierarchical-concurrency invariant finding #4 was protecting.

Please delete `as_type` from both wrappers plus its tests, and rewrite `configuration.md:31`/`:95` to document `with_dialect()` as the quota-preserving path. Runtime conversion is already vestigial: `TaskRequirements.input` + `derive_model_type()` decide the kind at bind time. **Keep `as_compat_type`** — one real caller (`session.py:2737`), and it rejects a mismatched kind rather than converting. Bonus: deleting `as_type` removes the `chat_model ⇄ gen_model` import cycle outright (both sides are function-local imports inside it).

### B4 — `alogprobs(echo=False, max_tokens=0)` on sglang forwards `max_new_tokens=0`, which sglang rejects

`transports/sglang.py:189-191` clamps only under `input_scoring`; `main` clamped unconditionally (`max(max_tokens, 1)`):

```
max_tokens=0  input_scoring=True  -> max_new_tokens=1   (fixed)
max_tokens=0  input_scoring=False -> max_new_tokens=0   (regression vs main)
```

The reply describes this as "restored", but the original finding was the `echo=False` half, which is the one still broken. No shipped task passes `max_tokens=0`, so it's latent — still a public-API regression. Hoist the clamp out of the condition.

### B5 — `ToolResultPart.is_error` is silently dropped while the audit reports the leaf accounted for

`openai_chat.py:425-438` reads `call_id` and `result` and never touches `is_error`, yet `input.modality.tool_result` is recorded `Consumed`:

```
ToolResultPart("call_1", "boom", is_error=True)
  wire: [{'role': 'tool', 'tool_call_id': 'call_1', 'content': 'boom'}]
```

Leaf granularity covers a whole record, so a non-default sub-field inside it is invisible to `active_request_leaves`. Chat Completions has no wire field for this, so dropping it is defensible — but that's exactly what `NoOp(path, reason)` is for. It also has a scheduled cost: Anthropic's `tool_result` **has** `is_error`, so step 3 of the #25 sequence must touch the IR or `openai_chat`'s lowering, which that scope decision's diff rule defines as "the plane is wrong." Either flatten these into leaf paths, or state in `dialect.py:126` that granularity stops at the record boundary.

---

## Structural — cheap now, expensive later; worth bundling

**N1 — The fingerprint layer has two defects.** (a) `model.py:355` and `sglang_gen_model.py:75` seed the binding with `uuid4().hex`, which reaches `RuntimeBindingPlan.fingerprint` and lands in `CapabilityEvidence` on `Response.provenance` (`model.py:874-875`, `:1219-1220`). Two identical constructions in one process give different `plan_fingerprint` values while `deployment_fingerprint` stays stable. This does not break `--resume` (the config path uses deterministic scopes and `Model.bind`), but `CapabilityEvidence` exists so plans can be compared across runs — a per-run-random fingerprint makes the field inert. (b) `_fingerprint` is implemented **three times** (`deployment.py`, `model.py`, `reconcile.py`) and `deployment.py`'s copy omits `allow_nan=False`, so it hashes `NaN` into non-standard JSON where the others raise. Collapse to one.

**N2 — `prefill` is the one keyed capability no `Request` leaf can activate.** The other 11 all map from a request path; `prefill` has no `Request` field and only `SglangTransport` implements it, off `dialect_options`. Both canonical dialects list it in `_CANONICAL_WIRE_KEYS`, so a caller gets "`'prefill'` must use its canonical provider-neutral field" — naming a field that doesn't exist. Like B5, this is a scheduled diff-rule violation (assistant prefill is an Anthropic feature). Give it a real `Request` owner or drop it from `CAPABILITY_KEYS` until the dialect that needs it lands.

**N3 — Duplicated private primitives.** Nine private names are defined in 2+ files; three byte-identical: `_copy_json_value` (20 lines, `capabilities.py` + `requirements.py`), `_validate_nonempty_string`, `_validate_top_logprobs`. Plus `_json_value` twice with different bodies (`ir.py`, `model.py`) and `_fingerprint` ×3. *Not* problems: 97 private functions with 43 single-caller is normal decomposition (the rule targets abstractions, not named local steps), and the 9 with zero call sites are all correctly registered callbacks. The gap is that there's no shared-internals module — suggest one `_shared.py`, and **rename** the per-dialect parsers, since two files each defining `_usage_stats`/`_choice_index` implies identical semantics they don't have.

**N4 — All three runtime import cycles come from the compat wrappers.** `model ⇄ chat_model` and `model ⇄ gen_model` via `as_compat_type` (`model.py:672-673`); `chat_model ⇄ gen_model` via `as_type`. The canonical plane is acyclic, and `dialect_registry ⇄ reconcile` is `TYPE_CHECKING`-only (clean). B3 kills the third; moving the wrapper-class → `input_kind` map into `dialect_registry` as data (it already has `input_kinds`) kills the other two → zero runtime cycles.

**N5 — `Any`/`cast` composition.** 54 `Any`, 51 `cast`, **zero** `type: ignore`/`noqa` — nothing passes by suppression. But: 21 × `dict[str, Any]`/`list[Any]`/`Mapping[str, Any]` are identical to bare generics, and worse, `PreparedRequest.body` is `Mapping[str, Any]` (`dialect.py:90`) while `ir.py` types the same data as `Mapping[str, JSONValue]` — the weak type is the plane↔dialect boundary itself, so a non-JSON object in `body` goes unnoticed. ~13 SDK-boundary `raw: Any`/`client: Any` should be `object` (the bodies already use `getattr`, and `_part_to_wire(part: object)` in the same file already does this). ~11 casts are one pre-existing root cause, not this PR's fault — `core/types.py:8` uses invariant `list["JSONValue"]` deliberately; one helper (`_json_list`) replaces them. Two genuine widenings worth explaining: `dialect_registry.py:283`, `sglang_gen_model.py:103`.

**N6 — `model.py` is 1265 lines because two lifetimes share a file.** ≈588 lines (46%) is the compat bridge (`_kwargs_to_request` alone is 212, plus `_response_to_model_output` 66, the `ModelOutput` family, `build_legacy_openai_binding` 64, `_legacy_runtime_plan` 61); canonical `Model` is ≈363, which is fine. Split for lifetime, not size — the plane evolves, the bridge dies after one compat cycle, and kept together its removal is surgery instead of a file deletion. It's nearly mechanical: `_kwargs_to_request` touches only `self.dialect_id`/`_validate_n`, `_response_to_model_output` only `self.meta` — free-function-shaped already, **no mixin needed**; only `_coerce_input` is genuinely polymorphic. Suggested: `model.py` ≈450, `_legacy_bridge.py` ≈400, `_legacy_binding.py` ≈130 (which makes N1(a) visible, since `uuid4()` lives there).

**N7 — `transports/` is named as deprecated but holds a live engine.** 383 of its 430 lines are `transports/sglang.py`, the only sglang `/generate` implementation; the other three files are 12–13-line aliases. The alias *direction* is correct, but the package name misdescribes its contents, and contamination runs backwards too — `dialects/openai_chat.py:505` carries a deprecated `capabilities` property "retained with the transport import alias." Moving the sglang wire code to `dialects/sglang.py` is the same action as B1's expensive tier and leaves `transports/` deletable in one commit. Related and still open: `_SAMPLING_PARAM_MAP` keeps 9 keys but `setdefault` runs *after* the 7 first-class fields are written, so only `min_p`/`repetition_penalty` are unconditionally live; and since `SglangTransport` has no `validate_request`, `dialect_options={"max_tokens": …}` is silently accepted as a second source, which both canonical dialects reject.

**N8 — Nothing marks this as breaking.** Title is `feat(models):` and none of the 5 commits carries `!` or a `BREAKING CHANGE:` footer (all bodies empty), so the removed generic, the canonical `Transport` name, derived-YAML kind conversion, and `with_args`'s new rejection all ship unflagged in a commit-generated `CHANGELOG`. Separately, `docs/superpowers/specs/2026-07-01-recipe-capability-layer-design.md` is still untracked — it's the only record of the verified backend behaviours, and the two SGLang probes it gates decide whether the `/generate` frontend survives, i.e. it unblocks B1's long-term fix.

---

## Follow-up — separate PR

- **Split `core/models/` by concern:** declaration/binding/reconciliation (4478 lines) vs wire protocol (2715) vs `Model` + wrappers (2107) under one package name. Suggest `core/models/` + `core/binding/`. Payoff is enforceability — today nothing stops a dialect importing `reconcile`; after the split `check_layer_imports.py` can assert it. Do this **after** B1, since moving sglang changes the boundary.
- **`DeploymentPlan` exists twice** — `infer/topology/models.py:168` and `core/models/deployment.py:70` `DeploymentPlanProjection`. Legitimate narrow-projection pattern, but a sync point nothing checks; worth a test asserting coverage.
- **`finish()`'s consumption cross-check cannot fail** — `dialect.py:87` promises "independently auditable evidence," but one closure populates both sides (`openai_chat.py:575-578`), so the mismatch branch is unreachable. The check with teeth is `missing = active - decisions`. Soften the docstring, or have `prepare` derive `consumed_paths` from the wire body it built.

## Question

`reasoning.summary == "none"` is recorded as a `NoOp` (`openai_chat.py:625-632`), but `ReasoningParams.summary` defaults to `None`, so `"none"` is a leaf the caller set explicitly. Is the intent that the two are semantically identical? If so, the `NoOp` reason describes the server rather than the equivalence, and `dialect.py:365` already treats them the same. Worth one clarifying word, since `NoOp` is the decision that asserts equivalence.

## Recommended action

**Request changes**, on B1–B5. B1 is the one that matters: this PR's headline invariant, unfixed on the default path for an entire engine, with the prior reproduction still byte-identical. B2 breaks 9 tests that pass on `main` and makes unit tests dial a real API. B3 is mostly deletion plus two doc lines; B4 and B5 are a few lines each.

N1–N8 are cleanup, not merge gates — but N1(b), N2, N3 and B5 are cheap now and expensive later, and N2/B5 are scheduled violations of the diff rule this sequence's own scope decision set for step 3. Worth bundling rather than filing.

Order: B1 (cheap tier) → B2 → B3 → B4/B5 → N1/N2/N3 → rebase again immediately before merge.

jack-scitix-ai:
Thanks for the second detailed review. I agree that B1 was the headline invariant: leaving the default path for an entire engine able to accept and discard active request fields would have undermined the purpose of the PR.

I rebased again onto current `main` and addressed B1–B5. The current head is `171f5da`.

### B1 — SGLang legacy now fails before allocation or I/O

`SglangGenModel.arun()` now calls a strict `_validate_legacy_request_leaves()` gate before acquiring its pool or invoking the native transport.

The gate:

- admits only leaves that `/generate` actually lowers;
- rejects suffix, reasoning, hosted-tool, session, and other unsupported active leaves;
- rejects unknown, wrong-dialect, null, and duplicate `dialect_options`;
- explicitly documents `scheduling.stream` as a native single-POST no-op;
- explicitly documents `sampling.seed` as a no-op on native `/generate`, because managed deterministic deployments pin `DeploymentPlan.seed` at engine startup.

The unsupported-field reproduction now raises `RequestAuditError` before pool acquisition, and the transport receives no request. Deterministic SGLang has a separate regression test proving that the injected request seed is accepted only through this documented no-op path.

This remains the cheap PR-1 compatibility gate. It does not turn `sglang_legacy` into the future bindable `sglang_native` dialect or move the PR-5 probes into this PR.

### B2 — removed hooks now fail loudly, and rebased tests use the public seam

`Model.__init_subclass__()` now raises at class definition if a subclass defines the removed `_agenerate_impl` or `_alogprobs_impl` hooks. A stale test double therefore cannot silently fall through to a real API call.

The affected tests were migrated to `_build_default_transport()` plus `HandlerTransport`, exercising canonical `Request` and `Response` values instead of the removed hooks.

After the latest rebase, the same migration was applied to the newly merged:

- AGIEval;
- GSM1K chat;
- GSM1K completion;
- GSM-Plus.

AGIEval also exposed a real composition requirement rather than only a test problem: its independent extractor is now a reconciled CHAT model role with a managed pool and lifecycle. `extractor: self` remains a sentinel until task construction, so it resolves to the candidate only after `infer_args` have been applied and therefore preserves the candidate’s pool, limiter, identity, and task-specific defaults.

### B3 — `as_type()` is removed

`ChatModel.as_type()` and `GenModel.as_type()` and their dedicated tests are removed.

The configuration guide now documents the quota-preserving path:

1. composition reconciles a target `RuntimeBindingPlan`;
2. `with_dialect()` verifies the target deployment, connection family, and identity;
3. the rebound model retains the existing client, pool, and limiter hierarchy.

`as_compat_type()` remains because the session has a real one-cycle compatibility caller. It is a non-owning shape adapter for an already-bound model and refuses to convert an incompatible input kind.

### B4 — the SGLang zero-token clamp covers both scoring paths

Native SGLang now clamps an explicit `max_new_tokens=0` to `1` whenever either input scoring or sampled-output logprobs are requested.

This covers both `echo=True` and `echo=False` and restores the legacy behavior without changing the server default when no explicit zero was supplied.

### B5 — branch-sensitive record fields are explicit leaves

`ToolResultPart.is_error=True` is now its own active leaf.

I chose `Rejected` rather than `NoOp` for OpenAI Chat: a failed tool result is not semantically equivalent to a successful one, and Chat Completions has no field that can transmit that distinction. The request therefore fails loudly instead of emitting byte-identical success-shaped wire data.

The same branch-level audit found and fixed the analogous image case:

- data-backed image `media_type` is consumed when constructing the data URL;
- URL-backed image `media_type` is explicitly rejected because Chat Completions has no corresponding wire field.

The audit documentation now states the intended boundary: record-level leaves cover fields preserved by every admitted branch, while branch-sensitive optional fields require their own leaves. Schema fixtures and regression tests cover both sub-fields.

### `reasoning.summary == "none"`

Yes, the intended meaning is semantic equivalence for the current Chat Completions dialect: it does not request a provider-side summary transformation and leaves the visible reasoning channel unchanged.

The no-op reason now states that equivalence directly rather than merely describing the server.

### Non-blocking structural findings

I also bundled the cheap structural changes that were directly adjacent to the blocking fixes:

- all model-plane hashing now uses one `_fingerprint.py` helper with canonical JSON and `allow_nan=False`;
- removing `as_type()` removes the wrapper-to-wrapper conversion cycle;
- the release-driving commit `6e25101` is marked `feat(models)!:` and carries an explicit `BREAKING CHANGE:` footer.

I have not claimed the broader N2/N3/N5/N6/N7 work as resolved in PR 1. The remaining private-helper consolidation, compatibility-bridge file split, `Any`/`cast` tightening, package-boundary split, and native-SGLang relocation remain follow-up work. `prefill` remains a legacy dialect option until a bindable adapter owns its provider-neutral representation, and `sglang_native` remains gated on the agreed PR-5 probes.

The previously untracked `docs/superpowers` artifact was not introduced as a new repository source in this PR. The accepted design and sequencing are recorded in RFC #25 Revision 2, while concrete serving constraints and probes remain owned by #47 and the native-SGLang follow-up.

ethan-scitix (commented):
## Verdict

Five of five blocking findings from the last review are fixed — verified by running each. One new blocking bug; the structural follow-ups are still open, and one prior finding is only partly done. **Commit messages and identities clean**: 9 commits, one author (`jack-scitix-ai <jack02@scitix.ai>`, same address as 10 already merged on `main`), no trailers, no paths.

**Verified on `171f5da8`**, rebased clean onto `fb6ab105`: 5157 tests pass; `ruff`/format/`ty` clean; preflight 26/26; coverage 98.43%, every module ≥96%. Three mutations on the new invariants (B1 gate, B4 clamp, B5 leaf) — all caught. B1's gate also checked in the inverse direction: all 12 leaves it admits as lowered genuinely reach the wire, covering `SamplingParams` (8 lowered + `seed` as a documented NoOp) and all 3 `ScoringParams`.

| Prior | Status |
|---|---|
| B1 sglang audit bypassed | **Resolved** — prior repro raises before pool acquire, no POST |
| B2 tests dial network | **Resolved** — `__init_subclass__` rejects both removed hooks |
| B3 delete `as_type` | **Resolved** — gone from code, tests, docs; `as_compat_type` kept |
| B4 `echo=False` clamp | **Resolved** — all 8 `alogprobs` shapes clamp to ≥1 |
| B5 `is_error` dropped | **Resolved** — own leaf, plus image `media_type` |
| N1(b) `_fingerprint` ×3, N8 breaking marker, summary Question | **Resolved** |
| **N4 import cycles** | **Partial — 1 of 3** (below) |
| N1(a) `uuid4`, N2 `prefill` | Unaddressed |
| N3/N5/N6/N7 | Deferred; reasonable for PR 1 |
| Follow-up: split `core/models/` | Deferred (was separate-PR material) |
| Follow-up: `DeploymentPlan` ×2 sync test | **Unaddressed** — `test_deployment.py` tests the projection's own validation, nothing asserts it covers the source's fields |
| Follow-up: `finish()` cross-check can't fail | **Unaddressed** — neither fix applied (below) |

## Blocking

### Bug: the `extractor` role is missing from the pre-session role registry

`session.py:1567` iterates `("grader", "extractor")`; `sieval/cli/resolution.py:484` still handles only `grader`. So `resolve_config_model_types()` hands `model_requirements_for()` no `extractor` binding and `agieval_0shot_gen.py:234` raises:

```
$ sieval run agieval-math.yaml --dry-run -o json    # extractor: {model: gpt-3.5-turbo, ...}
{"ok": false, "error": "AGIEvalZeroShotGenTask requires a 'extractor' model binding"}
```

That is the upstream-faithful config documented inside `examples/agieval-math.yaml` itself. It breaks `sieval run`, `--dry-run` and `sieval infer` (every path through `_prepare_launch_batch`), while `sieval eval` accepts the identical config (`validate_eval_config` → `errors: []`). An inline `grader` works on both, isolating it to the role list. It shipped green because `tests/unit/cli/test_resolution.py` has 22 tests and no non-candidate role — while the same scenario *is* covered on the other path (`test_session.py:2785`).

Fix plus mirror test ≈10 lines; I patched it and 466 existing tests still pass. Please have both sites read one shared role constant — two hardcoded lists that must agree is the "extract on coupling" case, and the next role added hits this again.

## Worth bundling — cheap now, expensive later

- **N4: 2 of 3 cycles remain.** `as_type`'s removal killed `chat_model ⇄ gen_model`, but `model ⇄ chat_model` and `model ⇄ gen_model` are still live, from exactly where the last review predicted — `as_compat_type`'s function-local imports at `model.py:651-652`, which exist only to map a wrapper class to `"chat"`/`"completion"`. Moving that map into `dialect_registry` (it already has `input_kinds`) still takes it to zero.
- **`ir.py:18` imports `EngineSource` from `deployment`** — the plane's own types depend on the binding module, so `ir` can't be lifted without dragging 507 lines with it. Relocating one alias fixes it.
- **Land the boundary rule now, before the split.** `check_layer_imports.py:69` already has `FORBIDDEN_SUBPACKAGE` (dotted sub-package keys, one rule today, with a dead-rule guard). Measured inside `core/models/` — 20 modules, 53 runtime edges, 9544 lines — nothing imports `reconcile` at runtime except `model`; `dialect_registry → reconcile` is `TYPE_CHECKING`-only. So the boundary the split would enforce is *currently intact*, and one config line pins it while that's still true.
- **Bug: `prefill`'s rejection names a field that doesn't exist** — `'prefill' must use its canonical provider-neutral field`. Deferring the capability is fine; the accurate reason already sits at `openai_chat.py:155`. Second time raised.
- **Nit: the gate's two duplicate branches are untested** — `sglang_gen_model.py:113`/`:116`, the only uncovered lines there, and two of the four rejections the PR description advertises. Both reachable; I hit each by hand.
- **Nit: `dialect_options={"max_tokens": …}` alone is accepted on sglang**, unconditionally rejected by both OpenAI dialects. Duplicate-only → unconditional closes the drift.
- **Nit:** `task.py:217` — "a `'extractor'`" → "an".

## Follow-up — issues, not merge gates

- **Decide whether `capabilities` is vocabulary or logic, then split.** `capabilities.py` (922 lines) holds both the declaration vocabulary a dialect legitimately needs (`CAPABILITY_KEYS`, 12 `*Options`, `CapabilitySpec`) and binder-only logic (`normalize_capability_declarations`, `aggregate_capability_intents`, `CapabilityIntent`, `RequestDefaults`). That single module straddles the line `core/models/` + `core/binding/` would draw, which is why all 5 wire→binding edges are dialects reaching for the vocabulary half and dragging the logic half along (the other 2 are `dialects/* → deployment` for one `BINDING_RESOURCE_KEYS` constant). Settling this first makes the split obvious instead of arbitrary.
- **`finish()`'s cross-check still cannot fail.** `consume()` at `openai_chat.py:598-601` does `audit.consumed(path)` and `consumed.add(path)` in the same two lines, and that set becomes `consumed_paths` at `:695` — so `dialect.py:221` compares a set with itself. Worth at least softening `dialect.py:87`'s "independently auditable evidence": a check that claims a defence it doesn't provide is worse than no check.
- **`uuid4()` still randomizes `plan_fingerprint`** (`model.py:344`, `sglang_gen_model.py:163`), and it reaches disk — `model.py:1199` writes `provenance` into `model_calls[*].model.provenance`, `loader.py:61` rehydrates it, and `build_legacy_openai_binding` stays live via `ChatModel(**grader)` in three tasks. Two identical runs therefore produce records differing in a provenance field. `--resume` is unaffected (that gate compares `infer_plans.yaml`/`effective_config.yaml`) — promote to blocking only if byte-level artifact diffing counts as a supported reproducibility check.
- **Question: grader migration is 4 of 7.** browsecomp/aa_lcr/hle/simpleqa_verified get reconciled role models; advanced_if/complex_constraints/inverse_ifeval have no `model_requirements_for` and still build a private client in `Task.__init__`. Not a regression (main was 7/7 private), but the same `grader:` key now means "pooled role model" in four tasks and "hidden private connection" in three — the latter being the quota invariant prior finding #4 protected. Intentional for PR 1?
- **Nit:** `configuration.md:19` pairs `name: "gpt-4o"` with `type: "gen"`. The flip is required (the example's task is base-completion, and `type: chat` now correctly fails), but gpt-4o has no completions endpoint — rename the example model rather than the type.

## Recommended action

**Request changes** on the `extractor` bug alone. The bundled items are a few lines each, and of them the `FORBIDDEN_SUBPACKAGE` rule has the best return: it stops the dependency structure degrading further without touching the architecture. Everything the last review blocked on is genuinely fixed and the sglang gate is sound in both directions. Rebase again immediately before merge.

jack-scitix-ai:
Thanks for the follow-up review. I fixed the `extractor` blocker and used the same pass to close the adjacent role-composition gaps it exposed.

### Blocking role-registry bug

`resolution.py` and `EvalSession` now read the same `TASK_MODEL_ROLES` registry, so `grader` and `extractor` cannot diverge between pre-session resolution and execution.

The pre-session path now handles both supported extractor forms:

- an inline extractor binding is normalized and presented to `model_requirements_for()`;
- `extractor: self` remains a task sentinel and is resolved against the final candidate after `infer_args`.

Mirror tests cover inline extractor resolution, `extractor: self`, malformed role sources, and agreement between dry-run/prelaunch and session setup.

### Role composition was completed, not left 4/7

The remaining `advanced_if`, `complex_constraints`, and `inverse_ifeval` grader paths were not safe to leave as private clients: after PR 1 introduced `models_by_role` injection for bound roles, those constructors could fail at runtime.

They now declare a CHAT grader requirement, accept the reconciled `models_by_role`, reject competing grader sources, and preserve the existing direct-construction compatibility path.

Role validation now fails in both prelaunch entry points when:

- configuration supplies a role that the task constructor does not own;
- a requirement hook omits a normalized configured role;
- an auxiliary-role hook cannot receive `models_by_role`;
- a `Model` is hidden under an unregistered ordinary task argument.

### Bundled structural items

I also applied the cheap items from the review:

- removed the two remaining runtime wrapper cycles by registering exact compatibility wrapper kinds in `dialect_registry`;
- moved `EngineSource` into a neutral private sibling module;
- added the `dialects -> reconcile` prohibition to `check_layer_imports.py`, with positive and negative tests;
- made `prefill` / `prefix` rejection messages describe the actual missing wire operation rather than a nonexistent canonical field;
- tested the reachable SGLang legacy rejection branches and made canonical-owned dialect options such as `max_tokens` unconditionally invalid;
- changed `PreparedRequest` from “independently auditable evidence” to “explicit audit evidence”;
- fixed the `extractor` article and changed the completion example model to `gpt-3.5-turbo-instruct`.

### Additional prelaunch hardening

While tracing the complete role matrix, I also made the two composition entry points share inline-binding normalization, resource validation, credential stripping, and fingerprints.

Composition-owned task constructor keys are rejected before client creation. External roles retain the exact caller-owned model source for that `(task, role)` instead of recovering an arbitrary model with the same binding ID. Inline roles also reject the named-model-only `sglang_legacy` bypass during resolution instead of failing later during binding.

The broader capability/package split, source/projection synchronization check, and legacy `uuid4()` artifact-determinism question remain follow-up work as classified in the review; this patch does not broaden PR 1 into those refactors.

ethan-scitix:
@jack-scitix-ai — everything blocking is resolved; I pushed three small commits to this branch directly (maintainer edit) rather than asking for another round. Heads-up so they don't surprise you on your next pull, and one question about what comes after this merges.

## What I pushed

`629367ae..f5211fa3`, fast-forward, all nine checks green and `mergeStateStatus: CLEAN`.

| commit | what | why |
|---|---|---|
| `00905f9e` | `PreparedRequest.body` → `Mapping[str, JSONValue]` | it was `Mapping[str, Any]` while `ir.py` types the same wire payload as `JSONValue`, so the plane↔dialect boundary was the one place a non-JSON object passed unnoticed. One line; no dialect needed changing, and `ty` now rejects it at the construction site |
| `59615739` | `validate_eval_config` runs the model-role check | `_validate_tasks` called `validate_task_config_args` without `task_class`, which early-returns before the role check. `sieval eval` accepted `grader:` on a task whose constructor has no such parameter while `run --dry-run` rejected it — the same two-entry-point divergence that let the missing `extractor` role ship. Resolution is fail-soft; the import failure is still reported by the class-import check |
| `f5211fa3` | `DeploymentPlan` fingerprint coverage derived from the dataclass | `test_fingerprint_covers_all_plan_fields` hand-listed its variants, so it kept passing after the source grew a field — which is exactly how `DeploymentPlanProjection` could silently fall behind. Now cross-checked against `dataclasses.fields` |

Verification: 5237 tests pass (was 5224), `ruff`/`ruff format`/`ty` clean, preflight 25/25. The ten new inline-normalizer guard tests were mutation-checked — disabling each of the four guards in `normalize_inline_model_binding` fails the suite. The `DeploymentPlan` test was checked the same way: adding a field makes it fail with an instruction to decide whether the projection must carry it.

One side effect worth knowing: the branch now has three distinct authors, so GitHub will append a `Co-authored-by:` trailer per author when this is squashed. All three are legitimate addresses.

## The ask: please drive steps 2–5

Revision 2 fixes the order — `openai_responses` (2), `anthropic_messages` (3), `google_genai` (4), `sglang_native` (5), `vllm_native` deferred — and states that **#25 is complete only after PR 5**. But there is nothing tracking them: no issues, no drafts, and this is the only open PR in the repo. Once it merges, the sequence has no artifact at all.

Could you open a tracking issue (or four) for steps 2–5? Two things are worth capturing while they are fresh:

- **PR 5 has a hard prerequisite you wrote yourself**: *"The probe must run and its evidence must be recorded before PR 5 implementation."* Neither SGLang probe has run, and the design document that gates them is still untracked. It should not live only in a working tree — but note it belongs in the internal repo, not here, now that this one is public.
- **Steps 2–4 are the falsification steps.** `openai_responses` tests the output envelope, `anthropic_messages` forces capability *absence* to be first-class, `google_genai` forces real message-shape lowering. If one of them needs a shared-plane change, that is the signal the diff rule exists to catch — worth stating in the tracking issue so it is not read as a nuisance.

## Timing question, so we don't collide

Three refactors are queued behind this merge, all of them things this PR's review surfaced:

1. `_resolve_grader` is duplicated verbatim across seven tasks (plus agieval's extractor variant) — the declaration half is already shared via `Task._bind_role_requirement`, only the resolution half is copied. This one has a deadline: the next grader-bearing benchmark adds an eighth copy.
2. Six private primitives are defined twice each — `_copy_json_value`, `_validate_nonempty_string`, `_validate_top_logprobs`, `_json_value`, `_usage_stats`, `_choice_index`. (`_usage_stats`/`_choice_index` want renaming rather than merging: same name, different semantics per dialect.)
3. `model.py` split by lifetime — the compat bridge is ~37% of it and dies after one deprecation cycle; isolating it makes that a file deletion instead of surgery. Measured: `_kwargs_to_request` is 212 lines and touches only `self._validate_n` and `self.dialect_id`, so it is still free-function-shaped. Splitting out `_legacy_binding.py` also makes the `uuid4()` binding fingerprint visible and removable.

By your own §3 diff rule, adapter PRs are confined to a new dialect package plus one registry entry — so plane refactors can only land *between* adapter PRs, and right now is the largest such gap. (2) and (3) are plane surgery; (1) touches only task files and collides with nothing.

**When do you plan to start PR 2?** If you are already writing it locally, I will hold (2) and (3) until it merges rather than make you rebase. If it is not started, I would rather take the gap now.

Issue #58, opened 2026-07-30
[RFC]: Observed-serving facts must flow back into dataset packing — the RULER case


ethan-scitix:
### Motivation.

RFC #27's Mechanism B and RFC #47 both model a **one-way** channel: the eval declares what it needs, the launcher turns it into flags. RULER (#11, landed 2026-07-30) is the first case where the dependency runs **both ways**, and today it is resolved by a hand-written knob plus a fallback heuristic that its own docstring calls wrong.

The fact in question is *the real served context window, and how it was reached* (native window vs YaRN-scaled, and to what size). Three places need it and none of them share it:

**1. The dataset needs it, and says so.** `sieval/datasets/ruler/_shared.py:135-139`:

> *"Which case applies is not derivable from `context_length` alone — it depends on serving topology (native window vs YaRN-scaled, and to what size), which this function doesn't have visibility into. Callers must say so explicitly via `reserve_think_budget`."*

The fallback when the caller doesn't say so is `context_length == 131072`, documented in the same file (`_shared.py:157-162`) as *"wrong for e.g. a length served natively at that exact size (no YaRN, no larger native window)"*. The decision changes how many tokens each sample is packed with — a wrong answer does not error, it silently changes the sample set.

**2. The launcher already has it, in a different vocabulary.** `examples/ruler-qwen3-8b-nonthinking.yaml:81-85` expresses the serving side as hand-written user overrides, duplicated per model entry:

```yaml
overrides:
  context_length: 131072
  disable_cuda_graph: true
  json_model_override_args: '{"rope_scaling": {"rope_type": "yarn", "factor": 4,
    "original_max_position_embeddings": 32768}}'
```

Every shipped recipe caps `max_model_len` / `context_length` at **<= 40960** (qwen3: 8192/16384/32768/40960; qwen2.5: 16384/32768). So RULER's 64k and 128k cells are only reachable by overriding the recipe, and nothing connects that override to the dataset's packing decision.

**3. The task reads a per-sample stamp of the assumption.** `ruler.py:302,328` stamps `think_budget_reserved` onto every sample and `ruler_0shot_gen.py:181-188` sizes `max_tokens` from it. The assumption is **already serialized per sample** — nothing verifies it against the server that actually answers.

[...]

Issue #47, opened 2026-07-28
[RFC]: Capability constraint layer — declare and enforce the silent cross-group constraints


ethan-scitix:
### Motivation.

Some eval requirements are not request parameters — they are **relationships** between capabilities, and violating them produces **silently wrong scores** rather than an error. Today those relationships live only in task docstrings.

Concretely, in `sieval/tasks/`:

```
files carrying --disable-radix-cache / --max-logprobs prose:  9
total mentions:                                              28
lines of code that read or set them:                          0
```

`_arc.py`, `arc_{easy,challenge}_kshot_{ppl,clp}.py`, `c_eval_kshot_clp.py`, `cmmlu_kshot_clp.py`, `mmlu_kshot_clp.py`, `mmmlu_kshot_clp.py` all instruct an operator to launch the server a particular way. Nothing verifies they did. Two known failure modes:

1. **PPL tasks × prefix cache.** On a cache hit sglang truncates `input_token_logprobs` to `prompt_tokens − cached_tokens` and does **not** report it; vLLM V1 errors instead. Requires `--disable-radix-cache` / `--no-enable-prefix-caching`. `SglangTransport._guard_radix_cache` catches the sglang symptom at request time — but only on the native `/generate` path, and only after a deployment has already been spun up.
2. **CMMLU / C-Eval / MMLU top-k breadth.** vLLM defaults to `--max-logprobs 20`; A/B/C/D are not guaranteed to appear. Needs `--max-logprobs 100`. Currently, per `docs/superpowers/specs/2026-07-01-recipe-capability-layer-design.md`, this *"relies on an operator reading the task docstring."*

RFC #25 introduces a capability model, but its `Capability` set can only express *"this feature is supported"*. It cannot express *"these two must not be on together"*, *"this field needs a value ≥ N"*, or *"this one is mandatory"* — so neither failure above is declarable, let alone enforced. This RFC proposes the missing layer.

This is deliberately **separate from and downstream of** RFC #25. The constraint table has to be filled from probe results against real servers, and those probes have not run yet; three of the six constraint claims currently encoded in comments are unverified (see "Any Other Things"). Writing the table now would repeat the mistake RFC #25 made: modelling from vendor docs instead of from observed behaviour.

#### 1. Three axes — where a constraint is declared

Capability is `f(frontend, engine, instance)`, not a property of one object. `vLLM` and `sglang` both serve the `openai_completions` wire protocol; `sglang` additionally serves its native `/generate`. So:

[...]

PR #59, opened 2026-07-30
feat(recipes): split recipe params into hardware and capability layers

ethan-scitix:
## Type

## Summary

Implements **Mechanism A** of RFC #27. The RFC's scope was split first (see [its comment thread](https://github.com/scitix/sieval/issues/27#issuecomment-5133471958)): Mechanism B is superseded by #47, and the two-way case neither can express is now #58. This PR is only the recipe schema half.

- **The problem is live, but it is hygiene, not a fault.** `profiles[hardware][precision][framework]` mixed perf params with behavior params and matched on family + size only, so a base checkpoint resolved to the instruct profile and inherited its parsers. `qwen2.5-72b` (`size_range: [60, 80]`) carries `enable_auto_tool_choice: true` + `tool_call_parser: hermes` on **every** leaf — and Qwen2.5-72B-Base is the named comparison target of the ARC and HellaSwag ppl/clp tasks, which resolve to that same bucket. Those flags are **inert on a base checkpoint: accepted and unused, not a startup failure** (see Manual below). So this is a correctness-of-record fix — the params we persist for such a run currently overstate what the engine used — not a crash fix.
- **Two layers now:** `hardware[gpu][precision][framework]` (dtype, memory, parallelism, context) and `capabilities[instruct|base][framework]` (parsers, tool choice). `resolve_profile` → `resolve_hardware_profile` + `resolve_capability_profile`, merged capabilities-last. Collapsing the duplicated capability params takes them from **352 lines to 78**, and the three recipe files from 1328 to 1203.
- **The model type is derived, not read.** No config in the repo sets `type:` — it is left to inference from the tasks pointing at the model. Reading the explicit field alone (as the original plan did) would have resolved every base checkpoint to instruct capabilities and made this change a no-op on real configs. The derivation the eval session already performed is promoted to a shared `derive_model_type`, since both sites must agree for one model.
- **gpt-oss `no_enable_prefix_caching` deliberately stays on the hardware layer.** It is model-intrinsic rather than hardware, but prefix caching is being modelled as an engine-level constraint in #47; moving it now would migrate it twice. Noted in the YAML.

### Review follow-ups (3rd commit)

- **The golden fixture pinned a reimplementation of the merge, not the production one.** `test_golden.py` called the two layer helpers itself, so swapping the `merge_params` argument order in *either* `_resolve_recipe_params` or `auto_resolve_plan` left the entire suite green while the key order in `infer_plans.yaml` moved — precisely what `--resume` compares byte-for-byte. Both production paths now assert hardware-before-capabilities on a **resolved plan**; both mutations fail as they should.
- **`auto_resolve_plan` merged the layers with a dict literal**, which `sieval/infer/params.py` calls out as the anti-pattern `merge_params` exists to replace (it skips key normalization and the dash/underscore collision check). It also collapsed a present-but-empty hardware profile to `None`; presence is now preserved.
- **`capability_model_type` mapped everything that was not `"gen"` onto `"instruct"`.** This change makes `base` a first-class recipe word while the config keyword stays `gen`, so `type: base` was a likely mistake that silently selected the *opposite* layer — and the eval session only rejects an unknown type after the engine is launched. Unknown types now raise while the typo is still cheap.
- **`derive_model_type` is reached from the infer layer before an `EvalSession` exists**, and full config validation only runs under `--dry-run`, so a list-shaped `tasks:` surfaced as an `AttributeError` from inside recipe resolution. The session's shape check is now shared; both entry points report the same error.
- Recorded rationale for three things a reader would otherwise have to re-derive: why `base:` is declared explicitly empty rather than omitted, why `derive_model_type` stays in `session.py` (the inference step *is* task-class resolution), and why `sieval infer start` on a bare checkpoint keeps the instruct default (no task context exists there; deriving from an absent `chat_template` would close it, but a dead flag does not justify widening introspection beyond `config.json`). `_capture_golden.py` is pinned to the pre-split ref, since `origin/main` stops carrying `profiles` once this merges.

## Related Issues

Refs #27. Related: #47 (supersedes Mechanism B), #58 (the two-way case).

### Automated

- [x] Lint/format clean (`ruff check && ruff format --check`) — the one remaining hit is `sieval/_version.py`, build-generated and gitignored
- [x] Type check clean (`ty check`) — "All checks passed!"
- [x] Unit tests pass — 3460 passed. Also integration (54), acceptance (10), `check_layer_imports.py` (exit 0), and `check_preflight.py` (all PASS)
- [x] Rebased onto latest `main` (`1c15c00c`)

New coverage:
- `tests/unit/infer/recipes/test_golden.py` — characterization against a committed fixture of all 168 pre-split triples: content preserved, order preserved, base path omits capability params
- `tests/unit/cli/infer/test_resolve.py` + `tests/unit/infer/topology/test_resolver.py` — **key order asserted on a resolved plan** for both production merge sites, so a reordered `merge_params` call fails instead of silently changing `infer_plans.yaml`
- `tests/unit/infer/test_recipes.py` — legacy-key rejection, `capability_model_type` mapping **and its rejection of `instruct`/`base`/typos**, capability resolution, and a sweep asserting no shipped hardware leaf still carries a parser and every entry declares both model types
- `tests/unit/cli/infer/test_resolve.py` — end-to-end through `resolve_infer_config`, including the load-bearing case: **no `type:` in config + a gen task ⇒ base capabilities**
- `tests/unit/cli/leaderboard/test_session.py` — `derive_model_type` directly (since `sieval run` calls the function rather than the session method), plus malformed-`tasks:` shape rejection

### Manual

- [x] No GPU-backed launch performed. Resolution is exercised through the mocked-GPU path used by the existing suite; nothing here touches deployment.
- [x] **Resolved — this sets the change's priority.** A base checkpoint launched with `enable_auto_tool_choice: true` **starts normally**; the flag is accepted and unused, not a startup failure. RFC #27's "meaningless" is literal. So this ships as **hygiene / correctness-of-record**, not a bug fix: the cost of the old behavior is that `infer_plans.yaml` records serving params the engine never applied. It also means `sieval infer start <checkpoint>` keeping the instruct default is an acceptable known gap rather than a live defect — documented at the call site.

### Required (all PRs)

- [x] PR title follows conventional format
- [x] No new upper-layer dependencies added to `core/` — `core/` untouched; `infer/` still imports nothing above it
- [x] Deleted code verified — `resolve_profile` had exactly two production call sites (`cli/infer/recipe.py`, `infer/topology/resolver.py`), both migrated; no references remain

### If: Breaking Change

**What breaks.** The `profiles` key is **rejected** with a migration error naming both replacement layers, rather than read for compatibility. Recipe files ship in-repo so there is no legacy on-disk state to support, and silently accepting `profiles` would keep handing base checkpoints the params this split exists to withhold. User customization via `overrides` is unchanged. `resolve_profile` is removed from the `sieval.infer.recipes` public surface. Additionally, an unknown model `type:` now raises during infer resolution rather than at model construction — same rejection, earlier and cheaper.

**This must ship as a MINOR bump (0.8.0), not a patch.** Instruct resolution is content-identical for all 168 triples, but key **order** changes for the two gpt-oss recipes, whose capability params sat *between* hardware params pre-split. `infer_plans.yaml` is dumped `sort_keys=False` and compared **byte-for-byte** under `--resume`, so a 0.7.x run resumed on this code would abort. That alone is the reason for the minor: a byte-compared on-disk artifact changed shape.

**Correction (review):** an earlier version of this section claimed the resume version gate would turn that abort into a clear version message as 0.8.0, and a confusing plan diff as 0.7.1. That is wrong — the gate cannot reach it. `EvalSession.arun()` calls `_persist_effective_config()` then `_persist_infer_plans()` *before* `_prepare_execution()`, while `gate_resume_version` fires in `TaskRunner.__init__` (built inside `_init_runner()`). So the plan-mismatch `RuntimeError` wins at **either** version. Worse, `_brief_diff` re-parses both bodies with `yaml.safe_load` before diffing, and the parse discards ordering, so this exact change reported itself as `Diff: (whitespace / formatting only)` while aborting. Reproduced, not inferred. The version bump stays 0.8.0 (correct semver); the message is fixed separately on `resume-abort-message-clarity`, which names the reordered key path and appends the producing version. Moving the gate is deliberately not done there: it reads a *per-task* `<result_dir>/<task>/meta.json` and is fail-closed on a missing one, which a not-yet-started task legitimately has. The golden test pins the reordered set to exactly those two recipes, and the two new plan-level order tests pin the merge that produces it, so no further reordering can slip in unnoticed. Staged in the pending-CHANGELOG notes with this reasoning.

## Note for reviewers

Two things I did **not** do, deliberately:

1. **Did not commit the design doc.** #47 asks for `docs/superpowers/specs/2026-07-01-recipe-capability-layer-design.md` to be committed, and I initially agreed in the #27 comment. That is wrong for this repo: `docs/designs/` is in `.gitignore` (line 185) and only `docs/guide/` is tracked — in the public **and** internal repo. Design specs are local-only by convention here, so tracking one would need an explicit convention change, which is not this PR's call. The doc now carries a scope-reconciliation header locally. The underlying concern is still valid but narrower than #47 implies: the sglang truncation evidence is already in `sglang_gen_model.py:302-328` and the `--max-logprobs` evidence in 9 task docstrings; what is genuinely unrecorded anywhere tracked is the **vLLM V1 `prompt_logprobs` × prefix-caching mutual exclusion**.
2. **Did not touch the 9 task files' prose.** #47 states its Milestone 1 must land first — converting executable prose into a declaration nothing acts on is a net loss.


ethan-scitix:
## Review follow-ups (4th commit) + a correction to this PR's own description

`6a62772f` addresses the review. Two of the items are worth calling out because they change what this PR claims about itself.

### The description was wrong about the resume abort — corrected in place

It said: *"as 0.8.0 the gate rejects that resume with a clear version message; as 0.7.1 the user would instead get a confusing plan diff."*

The gate cannot reach it. `EvalSession.arun()` calls `_persist_effective_config()` then `_persist_infer_plans()` **before** `_prepare_execution()`, while `gate_resume_version` fires in `TaskRunner.__init__` (built inside `_init_runner()`). So the plan-mismatch `RuntimeError` wins at *either* version. Worse, `_brief_diff` re-parses both bodies with `yaml.safe_load` before diffing, and the parse discards ordering — so the exact change this PR makes reported itself as:

```text
Resume aborted: .../infer_plans.yaml does not match current invocation.
Diff: (whitespace / formatting only)
```

Reproduced, not inferred. **The 0.8.0 bump stays** — a byte-compared on-disk artifact changed shape, which is the semver argument on its own — but the promised message was not real. Fixed separately in #87, which names the reordered key path and appends the producing version. Moving the gate is deliberately not done there: it reads a *per-task* `<result_dir>/<task>/meta.json` and is fail-closed on a missing one, which a not-yet-started task legitimately has.

### `resolve_capability_profile` accepted the wrong vocabulary silently

It returned `{}` for any unrecognized key, and `auto_resolve_plan` exposed that key as a public `model_type` kwarg with no validation. So passing a *config* type resolved to **no** capability params — base-like serving, the inverse of this PR's purpose:

```text
capability='instruct' -> parsers={'reasoning_parser': 'qwen3', 'tool_call_parser': 'qwen'}
capability='chat'     -> parsers={}
capability='gen'      -> parsers={}
capability='typo'     -> parsers={}
```

It now raises for anything outside `CAPABILITY_MODEL_TYPES`, and the recipe-side parameter is renamed `model_type` → **`capability`** throughout — one name was carrying two vocabularies one call apart, which is what made the above look legal at the call site.

**`sieval run`'s path-only leg had no test asserting the translation.** Every existing test patches `auto_resolve_plan` wholesale, so removing the `capability_model_type` call there left all 1101 cli+infer tests green while an instruct model lost its parsers. Three tests now pin it; that same mutation fails 3 of them.

### Also in this commit

- `sieval infer start --model-type chat|gen` — closes the documented gap without widening checkpoint introspection, since `auto_resolve_plan` already took the key. Rejected in YAML mode rather than silently overriding the config.
- The `merge_params` comment in `auto_resolve_plan` claimed it reports a dash/underscore collision *across* the two layers. It does not — detection is within a single source; across sources it normalizes and the later one wins (`merge_params({'max-model-len': 1}, {'max_model_len': 2})` → `{'max_model_len': 2}`, no raise). Restated as what it does buy.
- `RoleAssignment`'s docstring still described `engine_params` as coming from "recipe profiles", the key this PR removes.
- `_capture_golden.py` pins the full pre-split SHA and fails with the ref named rather than writing an empty fixture when it is unreachable.
- `docs/guide/infer.md` documents the capability layer and the new flag.

Verified: 3471 unit tests pass, `ruff`/`ty` clean, `check_layer_imports.py` and `check_preflight.py` pass. Content parity re-checked independently of the golden fixture — 0 differences across all 168 triples, exactly 4 reordered leaves (gpt-oss-20b/120b × H100/H200, vllm only).

The architecture finding from the review — infer CLI importing the eval session for `derive_model_type` — is split into #86 rather than bundled here, so this PR stays on the recipe schema.

PR #62, opened 2026-08-04
feat(tasks): finish the stage-output protocol migration (33 tasks, 40/40)

### LINKED ISSUE #61: [Feature]: finish the stage-output protocol migration — 32 remaining tasks

#60 landed the stage-output protocol (`PromptRecord` / `PredictionRecord` / `JudgementRecord` in `sieval/core/tasks/records.py`) and piloted it on 7 tasks. Adoption is deliberately per-task and legacy return shapes keep working — which means that, left alone, the repo carries **two on-disk shapes indefinitely**.

That is the problem this issue tracks: the protocol's entire value is that a consumer can read any sample's prompt, prediction, ground truth and verdict *without knowing which task produced it*. At **7 of 39 tasks migrated**, every consumer still needs the per-task special case for the other 32, so the benefit is not yet realized anywhere. The tail is where the payoff is, not the pilot.

Three concrete debts are already sitting in the tree, not hypothetical:

1. **The math family is split mid-way.** `aime_2026` / `hmmt_feb_2026` migrated while their near-identical siblings did not, so the shared test `tests/unit/tasks/test_math_pass_at_k_family.py::PROTOCOL_TASKS` carries a fork that builds *two different feedback shapes* depending on set membership. It should shrink to empty and the fork should go.
2. **#51's documented limit stays open for three tasks.** `hle` / `browsecomp` / `aa_lcr` keep #51's flat `grader_reply` (response text only), so a reasoning autorater that spends its whole budget thinking still records an empty reply indistinguishable from an empty API response. #60 closed this for `simpleqa_verified` only, by persisting the grader's whole `ModelOutput` as `extra.grader_output`. It closes for the other three when they adopt the same shape — HLE is the case #51 flagged as most reachable, since it normally runs a reasoning judge.
3. **Anomaly semantics differ by shape.** `detect_empty_postprocess` defers on protocol records and `detect_extraction_failure` only reads them, so the same empty-output condition is reported under different rules depending on whether a task has migrated. Uniform coverage arrives only when the shapes are uniform.

### Inventory — 32 remaining, grouped by the pilot that is already its template

Each group has a landed precedent from #60, so these are ports rather than designs.

| group | precedent from #60 | tasks | n |
| --- | --- | --- | --- |
| LLM-judge | `simpleqa_verified_0shot_gen` | `hle_0shot_gen`, `browsecomp_0shot_gen`, `aa_lcr_0shot_gen` | 3 |
| math / pass@k | `aime_2026_0shot_gen`, `hmmt_feb_2026_0shot_gen` | `aime_2024_0shot_gen`, `aime_2025_0shot_gen`, `hmmt_feb_2025_0shot_gen`, `hmmt_nov_2025_0shot_gen`, `math_500_0shot_gen`, `imo_answer_bench_0shot_gen` | 6 |
| code execution | `livecodebench_code_generation_0shot_gen` | `human_eval_0shot_gen`, `human_eval_0shot_base_gen`, `mbpp_kshot_base_gen`, `livecodebench_code_generation_kshot_base_gen` | 4 |
| instruction following | `ifeval_0shot_gen` | `ifbench_0shot_gen` | 1 |
| single-rollout generative | `mmlu_pro_0shot_gen`, `gpqa_diamond_0shot_gen` | `mmlu_0shot_gen`, `drop_kshot_gen`, `openbookqa_kshot_gen`, `gsm8k_0shot_gen`, `gsm8k_kshot_base_gen`, `hendrycks_math_kshot_base_gen`, `theoremqa_kshot_base_gen`, `ruler_0shot_gen`, `t_eval_before_calling_0shot_gen` | 9 |
| conditional log-prob (`clp`) | `mmlu_pro_0shot_gen` — **shape-identical** | `arc_challenge_kshot_clp`, `arc_easy_kshot_clp`, `c_eval_kshot_clp`, `cmmlu_kshot_clp`, `mmlu_kshot_clp`, `mmmlu_kshot_clp` | 6 |
| perplexity (`ppl`) | `mmlu_pro_0shot_gen` for the record shape | `arc_challenge_kshot_ppl`, `arc_easy_kshot_ppl`, `hellaswag_kshot_ppl` | 3 |

Two groups are hybrids worth flagging: `hmmt_nov_2025` and `imo_answer_bench` are pass@k tasks that *also* use an LLM grader, so they take the `aime_2026` envelope with `simpleqa_verified`'s grader detail under `extra`. The four code-execution tasks are the ones that currently persist the evaluator payload verbatim as `metrics` behind a `ResourceMetrics` TypedDict — #60 already had to widen that TypedDict when the evaluator gained `n_cases`/`n_passed`, which is exactly the coupling the protocol removes.

### `ppl` / `clp` are **not** blocked — correcting the earlier framing

An earlier version of this issue (and #60's own Follow-ups) described these 9 as blocked on a design call, on the grounds that they "fan out over candidate answers rather than rollouts". Checked against the code, that diagnosis does not hold:

- **`postprocess` returns a single scalar in all 9** — an argmax index (`int`) or a winning option label (`str`). Every one of these tasks produces exactly **one prediction per sample**. The candidates are *evidence for* that prediction, not rollouts of it, so a `rollouts[]` list of length 1 is the honest representation, identical to `mmlu_pro` / `gpqa`. Per-candidate scores belong in `metrics` / `extra`.
- **The fan-out is in `infer`, which the protocol explicitly does not cover.** The 6 `clp` tasks return a single `ModelOutput` from `infer` anyway; only the 3 `ppl` tasks return `list[ModelOutput]`. The stage table deliberately leaves `infer` unwrapped, so that non-uniformity is orthogonal to adopting the three record types.

Revised assessment: **6 `clp` tasks need no design work at all** (they are shape-identical to already-migrated `mmlu_pro`: `str` prompt → one `ModelOutput` → scalar prediction → correct/category verdict), **2 `ppl` tasks (`arc_*`) are clean ports**, and only **`hellaswag_kshot_ppl`** had an open question — see below, now resolved.

One detail not to trip on: `mmmlu_kshot_clp` is the single task that boxes its infer result as `TaskStageOutput[OfficialScores]`. That does not conflict with the protocol's "records are returned bare" rule, because `infer` is not a record stage — do not remove the box while migrating it.

### Co-equal metrics — resolved in #60, but the remaining tasks must use it

The real modelling gap was not candidate fan-out, it was tasks with **two or more co-equal metrics**. The verdict record had exactly one binary slot (`correct`) and one continuous one (`score`), so a task with two metrics of the *same* kind had to park one in untyped `extra`, where only that task can find it — silently defeating the protocol. `ifeval` demonstrated this as a migrated task: `loose` went entirely into `extra`, and its own `report()` had to read `feedback_result["extra"][grade]` to get it back.

#60 now carries the fix: `RolloutJudgement` / `JudgementRecord` gained an optional **`metrics: dict[str, bool | float]`** holding every metric a verdict measured, named, including the ones the headline points at. `correct` remains the single cross-task-comparable axis (`n_correct` derives from it) — forcing one headline is deliberate — but the alternatives are now enumerable by a generic consumer. `ifeval` is the pilot: strict *and* loose both sit in `metrics`, and the headline is **derived from** that mapping so it cannot drift.

The remaining multi-metric tasks must adopt it rather than repeat the demotion:

| task | co-equal metrics | note |
| --- | --- | --- |
| `hellaswag_kshot_ppl` | `acc` / `acc_norm` (two argmax rules over one inference set) | this was the only genuine `ppl` design question; `metrics` answers it |
| `gsm8k_kshot_base_gen` | `correct` / `flexible_correct` (strict vs flexible extraction) | |
| `t_eval_before_calling_0shot_gen` | the T-Eval evaluator's whole metric dict (N axes) | pick the headline deliberately; the rest go in `metrics` |
| `drop_kshot_gen` | `em` / `f1` | **fits natively** — `em` → `correct`, `f1` → `score`; no `metrics` needed |

Two traps `metrics` enforces, worth knowing before writing the next one: a `None` or non-numeric metric value is **rejected by the builders**, because `obj_to_dict` drops `None`-valued keys and an unrecordable metric would read as "never measured"; and a per-sample *rate* in `metrics` does not let `report()` reconstruct a **pooled** rate, so the raw counts behind it belong in `extra` (IFEval's `follow_instruction_list` is exactly that case).

### Suggested sequencing

One PR per family, each carrying report-parity evidence the way #60 did (real-run replay where a run exists, randomized differential trials against the pre-migration `report()` otherwise). Reasons not to batch further:

- Each family migration is **another on-disk schema break for those tasks**, so each should ride a minor bump — the `--resume` version gate (`sieval/core/runners/resume_gate.py`, break axis `(major, minor)` below 1.0) is what turns that into a version message rather than a confusing shape error. Bundling unrelated families into one release makes the blast radius harder to state in the changelog.
- Per-family PRs keep the parity evidence reviewable. #60 needed per-task verification for seven tasks; thirty-two in one diff would not be reviewable.
- **The migration unit is the family module, not the task.** `sieval/tasks/_arc.py` holds `ARCFeedback` + `arc_report` shared by all four ARC tasks (challenge/easy × ppl/clp), and `arc_report` reads `feedback_result["correct"]` directly — so the four must move together, or the shared helper gains a shape fork of exactly the kind `PROTOCOL_TASKS` already is.

## Alternatives

**Leave adoption permanently opt-in (status quo).** Two shapes on disk forever, and every downstream consumer keeps a per-task branch. This does not just delay the benefit, it forecloses it — the protocol only pays off when a reader can stop asking which task wrote the row.

**One big-bang PR for all 32.** Not reviewable. The value of #60 was largely in its parity evidence per task, and that does not compress.

**A read-side adapter that normalizes legacy shapes instead of migrating tasks.** Tempting, since it would centralize the per-task special case in one place instead of 32. But it cannot work for the thing that matters most: **a reader-side shim cannot invent data that was never written.** Two of #60's seven tasks had no ground truth on disk at all before migration, because `raw_sample` is never serialized and their old feedback shape did not include it. No adapter recovers that from existing records — only re-running with a task that persists it does. An adapter would also have to be maintained forever, against 32 shapes that each predate the contract.

## Additional context

- Protocol landed in #60, along with the `metrics` tier and its `ifeval` pilot. Types and builders: `sieval/core/tasks/records.py`; vocabulary and conventions for task authors: `sieval/tasks/CLAUDE.md` § "Stage-Output Protocol (opt-in)".
- Related: #49 (run-level `meta.json` task identity — the run-level output-file envelope stays open and unprejudiced by this), #51 (`grader_reply` on the LLM-judged family, whose documented limit item 2 above closes).
- Correcting the record: #60's description says "Other ~80 tasks untouched". The actual figure is **32 of 39 registered tasks**, which is what this issue tracks.

### Adjacent follow-ups from #60 that are *not* task migrations

Recorded here so they are not lost, but none of them block the migration:

- **Route grader `model_calls` into the profiler (core-side).** `simpleqa_verified` carries the grader's call metadata inside the record because the profiler only reads task-supplied `TaskStageMeta["model_calls"]`, so grader spend is on disk but absent from `profile.json`. Worth fixing before the judge family migrates in bulk, since all three will inherit the same workaround.
- **Failure taxonomy belongs in the code evaluator, not the client.** #60 initially derived a `failure` category in the task by substring-matching the evaluator's free-text `msg`, and removed it: the evaluator is separately deployed, its message wording has already drifted once, and a client-side classifier decays silently. If a category is wanted it belongs on the response next to `n_cases`/`n_passed`. Needs the same upstream round-trip as the case counts.
- **Land the vendored evaluator patch upstream.** `vendor/code-evaluator` carries the `n_cases`/`n_passed` patch locally (`VENDORED.md`); still to land in `scitix/code-evaluator` and be re-vendored.
- **Full per-test-case pass rate.** `n_passed` is currently cases passed *before the first failure*, since the evaluator short-circuits. A true rate means dropping the short-circuit, which makes every wrong submission run all N cases and can newly trip the shared timeout budget — ship as an opt-in flag if wanted.

- [x] Make sure you already searched for relevant issues and documentation. (Searched all open + closed issues for protocol / record / schema / migration / stage / output; the only adjacent hits are #49 and #24, neither of which covers this.)

ethan-scitix:
## Type

## Summary

Finishes the migration #60 started: **all 40 registered tasks** now return `PromptRecord` / `PredictionRecord` / `JudgementRecord`, so a consumer can read any sample's prompt, prediction, ground truth and verdict without knowing which task wrote the row. At 7/40 the protocol paid off nowhere — every consumer still branched per task for the other 33.

- **33 tasks migrated**, one commit per family: math pass@k (6), clp/ppl (9), LLM-judge (3), code-execution (4), single-rollout generative (9) + ifbench, and `scicode` last in its own commit.
- **#51's documented limit closes.** `hle`/`browsecomp`/`aa_lcr` adopt `extra.grader_output` (the grader's full `ModelOutput`), so `finish_reasons` separates a reasoning judge that burned its budget thinking from an empty API response — the case #51 flagged as most reachable on HLE.
- **The `PROTOCOL_TASKS` fork is gone.** `test_math_pass_at_k_family.py` no longer builds two feedback shapes by set membership.
- **`ResourceMetrics` is retired** from all four code-exec tasks — the coupling #61 named, where an evaluator gaining `n_cases`/`n_passed` forced a client-side TypedDict widening.
- **Grader spend now reaches `profile.json`** (core-side; #60's accepted workaround, resolved — see below).
- **`ruler` gains a per-sample verdict it never had**; **`ifbench` grades in `feedback()`** rather than `report()`, mirroring #60's ifeval relocation.
- **`scicode` is the one non-mechanical port.** A sample is a chain of *dependent* sub-steps, each generating and executing code against the previous step's output — which is not what `rollouts[]` means (independent attempts at the same thing). So it maps to **one rollout per problem**: the whole step sequence is the attempt, per-step detail lives in `rollouts[0].extra.steps`, and `metrics` carries both axes with the headline derived from them. #61's inventory of 32 missed it because #42 landed just before #60.

> **On PR shape:** #61 asks for one PR per family, and it's right that 33 tasks in one diff is a lot. **The commits are exactly that split** — one per family, plus the core profiler change and scicode's two — each independently green with its own parity evidence, so this can be landed as-is or peeled apart without rework; the call is the reviewer's. Reviewing commit-by-commit reads the way the issue intended.
>
> **Independent of #63 / #64.** This branch touches neither `pdm.lock` nor `pyproject.toml`, so it does not conflict with the lock revert (#63) and #64's new drift gate does not fire on it. Its suite passes on **both** dependency sets — see the `openai` note in the review round.

## Related Issues

Closes #61. Refs #60, #51 (its documented `grader_reply` limit closes here), #49.

### Automated

- [x] Lint/format clean (`ruff check` → all passed; `ruff format --check` → 368 files already formatted)
- [x] Type check clean (`ty check` → **All checks passed**)
- [x] Unit tests pass — **2890** collected across `tests/unit` + `tests/integration` + `tests/acceptance` (was 2827 on `main`; +63). CI reports it as `2834 passed, 55 skipped, 1 deselected`; a local venv with more optional groups installed skips fewer.
- [x] `scripts/check_preflight.py` all-PASS; `sync_meta_index.py --check` and `sync_package_stubs.py --check` clean

### Manual — report parity

`report.json` metrics are unchanged for every task, verified against the **pre-migration implementations** (loaded from `d622448c` with the task decorator stripped), not only against new tests. **8,700 randomized differential trials, 300 per task, 8,700 identical.** Each family's fixtures deliberately exercise its edge paths:

| family | tasks | trials | edge paths deliberately covered |
| --- | --- | --- | --- |
| math pass@k | 6 | 1800 | `k ∈ {1,2,4}`, `n ≥ k`, rollouts `< k` so the `short` warning path runs |
| clp/ppl | 6 reports (9 tasks; 4 ARC share `arc_report`) | 1800 | cmmlu's `raw_sample=None` skip; hellaswag `acc` + `acc_norm` |
| LLM-judge | 3 | 900 | `judge_parsed=False` so the unparsed-drop stays out of the calibration arrays |
| code-execution | 4 | 1200 | `timeouts` substring counter in both directions, incl. the near-miss `"timed out"` |
| single-rollout gen | 9 | 2700 | theoremqa `empty`; ruler across both metric families with partial matches; t_eval all 6 axes |
| scicode | 1 | 300 | 167 solved problems / 4,529 graded steps; a deliberately sabotaged variant *does* mismatch, so the differential discriminates |

- ifbench is covered instead by its unit test, whose **expected report numbers are unchanged** across the grading relocation — that equality is the parity check.
- Grader-spend routing verified **end-to-end through the real `TaskRunner` + profiler**, not just the helpers: a bare-record judge task's grader tokens land in the `feedback` stage's usage (20 in / 1 out per sample), the candidate's own spend stays attributed to `infer`, and the recorded call names the judge model.

## Corrections to #61

- It groups `hmmt_nov_2025` / `imo_answer_bench` as pass@k tasks that "also use an LLM grader". **They do not** — both grade deterministically (`math_verify` / the vendored `verify_math_answer`), and imo's own notes say an LLM grader would break the reproducibility contract. No grader detail is recorded because there is no grader.
- It counts **39** registered tasks; there are **40**. `scicode_0shot_gen` (#42) landed just before #60 and is in no group — it is migrated here too, in its own commit (see Summary).

## Deliberate non-changes (so no denominator moves)

- **cmmlu's report still groups by `ctx.raw_sample`**, not the record's `extra.subject`. The two agree whenever both exist, but a final whose `raw_sample` is absent has always been dropped from the macro-average; switching the source would silently change the denominator. The record carries `subject` regardless, so a row on disk is self-describing.
- **`ruler`'s report still calls the vendored `string_match_*`** on the whole cell. The per-sample term recorded in `feedback` is the exact decomposition, but re-deriving the cell mean in the task would fork upstream's scoring — a vendored metric that no longer runs is a reproduction that drifts silently.
- **`mmmlu_kshot_clp` and `scicode_0shot_gen` keep their `TaskStageOutput` box on `infer`.** `infer` is not a record stage, so boxing it does not violate the bare-record rule. Both need the box for the same reason: one sample is *several* model calls, so the box carries the stage value while `meta` reports every call's usage to the profiler.
- The ppl tasks' "no option could be picked" sentinel `-1` becomes `None` (the protocol's spelling). Equivalent downstream: both mismatch the gold and both map to `""` through `choice_text`.

## The one deliberate denominator change (`90981721`)

`scicode`'s two accuracies disagreed about what a pipeline failure means: `main_problem_accuracy` counted a failed problem as unsolved, while `sub_problem_accuracy` dropped its steps from the denominator entirely. One metric was diluted by a failure while the other was silently *inflated* by its removal — and a full test split fell below the fixed 288-step denominator the official sub-problem figures are computed over, making the number incomparable to the leaderboard exactly when a run was unhealthy. A failed context still carries its `raw_sample`, so its tested steps are recoverable; they now stay in the denominator scoring zero, and a new `unevaluated_steps` field says how many of `total_steps` never ran.

**This is the one place `report.json` moves,** and only when `fails > 0` — a clean run is identical. It is a separate commit *ahead* of the migration so the parity trials above measure the migration alone.

## Notes on the `metrics` tier

`metrics` carries every multi-metric task, and in each the headline is **derived from** the mapping so the two cannot drift: hellaswag (`acc`/`acc_norm`), gsm8k-base (`exact_match`/`flexible_exact_match`), drop (`em`/`f1`), ifbench (strict/loose), scicode (`main_problem_solved`/`sub_problem_pass_rate`), t_eval (the co-equal axes its configuration scores).

t_eval needed a judgement call: `correct` is "every axis the evaluator scored came out at 1.0", deliberately **not** `parse_rate`. A task whose `correct` meant "the output parsed" would look near-perfect next to every other task on the one axis that is supposed to be comparable across them. *"The axes it scored"* is load-bearing rather than decorative — see the review round below.

## Review round — fixes in the latest push

Two of these were live bugs in the migration; the rest are latent or type-level.

- **`t_eval`'s `correct` was structurally unreachable.** `feedback` recorded every axis `_evaluate` returns, including ones the configuration never scores — and `correct` is derived as "every recorded axis is 1.0", so under the default `eval_thought=False` a pre-seeded `thought: 0.0` pinned `correct` to `False` on **every sample of every run**, emptying the one axis the protocol exists to make comparable across tasks. `metrics` now carries only the scored axes, from a `_metric_keys()` shared with the macro-average so the recorded metrics, the derived `correct` and the report cannot disagree. `report.json` does not move: the same axes were already being macro-averaged, and the `*_parsed` variants now read defensively so the `str` modes that score no args still report `0.0`, as before. Fixes a latent `NameError` on the way through — `metric_keys` was left unbound for any `prompt_type` outside json/str, surfacing deep in the macro-average rather than at config time.
- **`ruler`'s report crashed on the resume path.** `report()` indexed `rollouts[0]["prediction"]`, but `postprocess` normalizes a blank response to `None` and serialization **omits `None`-valued keys** — so on the resume-report path (every sample complete, `report.json` missing) the key is *absent*, not null, and an already-paid-for long-context run died at the last step with `KeyError`. Now `.get(...) or ""`. The test helper round-trips both records through `obj_to_dict`, so every report test that builds a final now asserts against the shape `report()` actually has to read, not the more forgiving in-memory one.
- **Four judge tasks spelled `grader_output` as a literal** instead of `records.GRADER_OUTPUT_KEY`. The name is what the runner reads back to route grader spend into `profile.json`, so a divergence is silent — tokens on disk, missing from the profile.
- **The five reports that count timeouts read `r["extra"]["msg"].lower()` unguarded.** `msg` is passed straight through from the evaluator response, so a null makes a *fresh* run raise `AttributeError` — not only a resumed one. Now `(r["extra"].get("msg") or "")`.
- **The record predicates were annotated `Any`.** `Any` passes silently in both directions — a caller may pass anything, and the body may do anything with it. `object` is the accurate spelling, and switching to it surfaced that `is_judgement_record`'s narrowing never reached its caller: only `Any` made `iter_grader_outputs`' `value.get("rollouts")` type-check. Both sniffers now return `TypeGuard[Mapping]` — deliberately not `TypeIs`, since a mapping that is *not* a record still returns `False`, which `detect_empty_postprocess` relies on when it tests `isinstance(post_result, dict)` after a `False`. Two of the three signatures predate this PR. Also replaces a non-discriminating assertion: `iter_grader_outputs({"rollouts": "not a list"})` read as covering the per-rollout `isinstance` guard, but without `n_rollouts` the value is not a judgement at all, so it returned at the *first* guard and deleting the one under test failed no test.
- **One new `t_eval` test could not construct its instance in CI.** `eval_thought=True` builds the embedding client in `__init__` with `api_key=os.getenv("SIEVAL_EMBED_API_KEY", "")`, and an explicit `""` is not `None` — it skips the OpenAI client's own `OPENAI_API_KEY` fallback, so `openai` raises `Missing credentials` at *construction*. The test now sets the env var: still a real instance, still covering the `eval_thought=True` branch, no embedding call. Reproduced rather than inferred from the traceback — `api_key=""` gives the CI error exactly, and it does so on both `openai` 2.45.0 (locked when this branch was cut) and 2.9.0 (restored by #63, now on `main`), so the fix is not version-specific. The whole suite was re-run against the 2.9.0 set before #63 landed, so this branch is verified on the dependency set it will actually merge into.
- **…and the message named a variable that would not help.** Follow-on from the above, since the wart is what made that test fail in the first place. `Missing credentials` names `OPENAI_API_KEY`, which is useless here: the endpoint is our own embedding service (`SIEVAL_EMBED_API`), not OpenAI's. The failure also lands at *construction*, before any run output, so that message is all an operator gets. `eval_thought=True` now checks the key itself and raises a `ValueError` naming **`SIEVAL_EMBED_API_KEY`**, pointing at `SIEVAL_EMBED_API` for a non-default endpoint, and saying the axis is optional (leave `eval_thought` unset). Both `None` and `""` take that path — `""` is the reachable case, an exported-but-blank variable. No new exception class: sieval has no config-error base (`ResultDirExistsError` / `ResumeVersionError` both subclass builtins), and tasks must not change `core/`. The new test discriminates — pre-fix the constructor raised `OpenAIError`, which is **not** a `ValueError` subclass, and its message contained `OPENAI_API_KEY`. Default `eval_thought=False` is untouched: it builds no client, and nothing in-tree passes `eval_thought=True`.
- **Four migrated tasks still *declared* the shapes they no longer return.** `gsm8k_0shot_gen`, `gsm8k_kshot_base_gen`, `hendrycks_math_kshot_base_gen` and `theoremqa_kshot_base_gen` had bodies fully on the builders but `Task[...]` slots still reading `list[ChatCompletionUserMessageParam]` / `str` for the prompt, `str` / `list[str]` / a per-task `Prediction` TypedDict for the prediction, and a per-task `Feedback` TypedDict for the judgement. Nothing caught it: the stage methods carry no return annotations, so the generic slots are never cross-checked against the bodies — and the five dead TypedDicts stayed *reachable* as those slots' arguments, which is exactly why the deletion sweep in the checklist below read clean. On-disk output was already correct, so no report or shard moves; what changes is that the declared type now matches it, the five unused TypedDicts are gone, and `gsm8k_0shot_gen` no longer imports `openai` at all.

### Required (all PRs)

- [x] No new upper-layer dependencies added to `core/` — the core change touches only `core/tasks/records.py`, `core/utils/meta.py`, `core/runners/runner.py`
- [x] Deleted code verified — grepped to **0 remaining references** for every removed name: `ARCFeedback`, `JudgeFeedback`, `GradeFeedback`, `ResourceMetrics`, `RulerFeedback`, `StepFeedback`, `ifbench._get_report`, and the per-task `Feedback`/`Preprocessed`/`Prediction` TypedDicts. Remaining `grader_reply` hits are prose explaining what replaced it. The per-task TypedDicts held out longest — five of them survived in four files as arguments to stale `Task[...]` slots, so a reference grep could not see them; removed in `67bd03bd` (see the review round above), and the sweep now reads 0 for real.

### If: Breaking Change

**On-disk schema, for the 33 newly-migrated tasks.** Result dirs written by 0.7.x hold the old feedback shape and cannot be re-reported by the new `report()`. Migration path: **start fresh.** This rides the **0.8.0** minor bump already staged for #60 and the infer-recipe split, so the `--resume` version gate rejects a 0.7.x → 0.8.x resume with a version message rather than a confusing shape error. `report.json` shapes are unchanged for every task except scicode's added `unevaluated_steps` (above), so leaderboards, alignment cards and `resolve_model_name` are unaffected.

**Module-level names removed from `sieval/tasks/*`** are importable, so their removal is user-visible even though nothing in-tree referenced them (verified above).

## Follow-ups (not in this PR)

Three of #61's adjacent items concern the code evaluator. **None is blocked upstream** — `code-evaluator` is vendored in this repo (`vendor/code-evaluator/`, formerly a submodule), so each is a local patch plus a `VENDORED.md` entry, exactly how the `n_cases`/`n_passed` patch already in that copy was done. What actually gates two of them is a **redeploy**: tasks reach the evaluator over HTTP (`SIEVAL_CODE_EVAL_API`), so a new response field only appears once the image is rebuilt from `vendor/code-evaluator/docker/`. The sieval side reads these with `.get()`, so an unpatched evaluator keeps working with the field absent.

- Land the vendored `n_cases`/`n_passed` patch **upstream** and re-vendor. Pure upstream hygiene — it keeps the fork delta from growing. It gates nothing: the patch is in the vendored copy, and all five tasks that call the evaluator for test-case counts already read both fields (the four migrated here plus `livecodebench_code_generation_0shot_gen` from #60).
- Failure taxonomy belongs on the evaluator response next to `n_cases`/`n_passed` — a client-side classifier over free-text `msg` decays silently, which is why #60 removed one. Doable in the vendored copy; needs the redeploy.
- Full per-test-case pass rate needs dropping the evaluator's short-circuit. Also local, but it carries a real runtime cost (every case runs, which can newly trip the shared timeout budget), so it wants an opt-in env flag — precedent: `CODE_EVAL_FLOAT_TOL` in the existing patch set.

Not code-evaluator, and **not** deferred any more: the `eval_thought` credential error is fixed in this PR (see the review round).

Still open, unrelated to the above: **task-level service env vars are undocumented.** `SIEVAL_EMBED_API_KEY`, `SIEVAL_EMBED_API` and `SIEVAL_CODE_EVAL_API` appear only in code — the README and FAQ cover `SIEVAL_DATA_DIR` alone. Pre-existing and broader than this PR; worth its own small docs change.


ethan-scitix:
### Seventh commit: a scicode metric fix, not part of the 32-task migration

`e4d3cecb` rides on this branch but is a **different concern** from the six per-family migration commits the description covers. Flagging it so it isn't read as more of the same.

`scicode_0shot_gen` reports two accuracies, and they disagreed about what a pipeline failure means:

| | before | after |
| --- | --- | --- |
| `main_problem_accuracy` | failed problem = unsolved | unchanged |
| `sub_problem_accuracy` | failed problem's steps **dropped from the denominator** | steps stay in, scoring zero |

So a failure diluted one metric and silently *inflated* the other, and a full test split fell below the fixed **288-step** denominator the official sub-problem figures are computed over — the number became incomparable to the leaderboard exactly when a run was unhealthy.

The comment justifying the drop ("their step counts are unknown") was false — a failed `TaskContext` still carries its `raw_sample`, so its tested steps are recoverable. They are now recovered, with the three scientist-authored steps (13.6 / 62.1 / 76.3) excluded just as on the success path. A context that failed *before* its sample was loaded has nothing to recover and is skipped; it still dilutes main-problem accuracy.

New `report.json` field **`unevaluated_steps`** — how many of `total_steps` never ran. Lets a reader separate "the model failed these steps" from "the pipeline never reached them"; read it alongside `fails`.

**No user-facing break:** scicode (#42) is unreleased, so this pins the definition down before it ships rather than changing it under anyone.

<details>
<summary>Audited against a prior in-house SciCode port — and where I deliberately did not follow it</summary>

The rest of the port **cleared**: identical sub-problem formula, byte-identical code extraction, identical prompt templates, and equivalent special-step gold injection under the default `verbatim` mode.

Not adopted from it:

- **Filtering failures out of *both* metrics.** That hides pipeline failures and makes the denominator depend on run health. Keeping both denominators complete is both self-consistent and faithful to upstream's fixed 65 / 288.
- **An empty-`step_background` guard.** Fail-loud is correct here: a silently-empty background yields numbers that aren't comparable to either setting, the field is required in the sample `TypedDict`, and upstream crashes too.
- **Decoupling the multi-step template from `with_background`.** That is an extension over upstream, and would add an uncalibrated score-relevant knob.

</details>

**Why it is on this branch:** it is the prerequisite for scicode's own protocol migration. Putting `sub_problem_accuracy` into the `metrics` tier means welding its definition into the shared contract, so the denominator semantics had to be settled first. The remaining open call is scicode's rollout shape (dependent sub-steps as rollouts vs. one rollout per problem with step detail in `extra`) — untouched here.

Tests: 25 in `tests/unit/tasks/test_scicode_0shot_gen.py` (was 22); full suite 2931 passed; `ruff` / `ty` / `check_preflight.py` / both sync scripts clean.

ethan-scitix:
### Eighth commit: scicode migrates too — adoption is 40 of 40

`c46185e9` closes the one gap I flagged in the previous comment. That comment ended with "the remaining open call is scicode's rollout shape … — untouched here"; **it is no longer open**, so read this as superseding that sentence.

`scicode_0shot_gen` is the one registered task #61's inventory of 32 misses (#42 landed it just before #60). It is not a mechanical port — a scicode sample is a chain of **dependent** sub-steps, each generating *and* executing code, which is not what `rollouts[]` means: independent attempts at the same thing.

**One rollout per problem.** The whole step sequence is the attempt.

| | where it goes | why |
| --- | --- | --- |
| `main_problem_solved`, `sub_problem_pass_rate` | `metrics` (both tiers, headline derived) | co-equal — neither is a footnote of the other |
| per-step verdicts | `rollouts[0].extra.steps` | mechanism detail |
| `correct_steps`, `total_steps` | record `extra` | `report()` **pools** them |

That last row is the load-bearing one, and it is the same call ifeval's instruction-level accuracy already rests on: a per-problem rate cannot reconstruct a pooled one, so the raw counts have to survive to report time. Averaging the `sub_problem_pass_rate` values across problems would produce a different — and wrong — `sub_problem_accuracy`.

Steps-as-rollouts was the alternative and it loses on semantics, not on convenience: step *i*'s prompt embeds the model's own code from steps 1..*i*−1, so the steps are neither independent nor attempts at the same thing. Recording them as rollouts would make `n_correct / n_rollouts` read as a pass-rate over retries for every generic consumer.

<details>
<summary>Two details worth a second look: what <code>prompt</code> holds, and when <code>prediction</code> is <code>None</code></summary>

**`prompt` is the first *generated* step's prompt**, not the whole problem and not all N step prompts.

It is knowable at `preprocess` because every step preceding it is a special step whose gold code is static (13.6 / 62.1 / 76.3 are never generated). `infer` then **reuses** that recorded prompt for the step rather than rebuilding it, via a shared `_gold_code()` — so the record structurally cannot disagree with what the model actually saw. Only steps 2..N are rebuilt, since their prompts embed the model's own prior output and cannot exist before inference.

`extra` carries `first_generated_step` / `n_generated_steps` / `previous_code`, so `prompt` is not misreadable as the whole model input. The two rejected alternatives: persisting all N per-step prompts is quadratic text (step N's prompt contains every earlier step's), and putting the sample in duplicates `raw_sample`.

**`prediction is None` iff *no* tested step produced extractable code.** A partial extraction miss is still a scoreable answer and keeps its prediction. This is deliberate anomaly-rule hygiene: `detect_extraction_failure` reports rollout indices where `extracted` is falsy, and a multi-step task where any one empty step nulled the prediction would fire that rule on most problems, making it noise. New `StepCode.extracted_code` holds the model's per-step answer so the prediction doesn't re-run the extractor.

`reference` is `None` — the ground truth here is a *procedure* (this problem's per-step test suites), same as human_eval.

</details>

**`report.json` unchanged.** 300 randomized differential trials against the pre-migration `report()` loaded from `e4d3cecb`, **0 mismatches**, covering 167 solved problems and 4,529 graded steps — with new-shape finals produced by running the **real** `feedback()` against a stubbed evaluator, not hand-built records. A deliberately sabotaged variant (perturbing `correct_problems`) *did* mismatch, so the differential discriminates rather than comparing two copies of the same bug.

Tests: **34** in `tests/unit/tasks/test_scicode_0shot_gen.py` (was 25), including one that drives the real `feedback()` twice over scripted verdicts and feeds both records into `report()` — so report is asserted against what feedback actually writes, not against a fixture that could drift from it. Full suite 2940 passed; `ruff` / `ty` / `check_preflight.py` / both sync scripts clean.

Issue #74, opened 2026-08-06
[RFC]: Sampled-answer metric family for math and MCQ — avg@k / pass@k / pass^k, consistency, and health signals


ethan-scitix:
## Motivation.

Every generative task in SiEval reports a single headline accuracy. For the
sampled-answer families — math and MCQ — one number per (model, task) is not
enough to do what SiEval exists to do: decide whether a converted / quantized /
re-served model has *degraded*. A model whose mean accuracy is unchanged but
whose answer distribution has widened is a real delivery defect, and today it is
invisible.

Three concrete gaps in the current code:

**1. `pass@1` is already `avg@n`, but nothing says so.** 13 tasks carry a
private `_pass_at_k` (identical copy in each, e.g. `tasks/aime_2025_0shot_gen.py`).
`_pass_at_k(n, c, 1)` reduces to `c / n` — i.e. the mean over `n` rollouts, which
is what MathArena publishes as `avg@k` and what simple-evals / Codex publish as
`pass@1`. The formula is right; what is missing is that **`n` appears nowhere in
`report.json`**. A run at `n=4` and a paper number at `n=16` land in the same
column with no way to tell them apart, which is exactly where published-number
alignment goes wrong.

**2. `pass@k` alone is the wrong direction for delivery verification.** It
measures the *best-of-k* upper bound. A model with higher sampling variance can
score a *higher* `pass@k` while being worse to ship. The reliability direction
(`pass^k`: all `k` rollouts correct) is the one that catches the failure mode we
actually care about, and we do not compute it.

**3. Health signals are logged, not measured.** `PredictionRecord` carries a
per-rollout `extracted` flag, and the `n_rollouts < k` case (a model or a
mid-stream drop returning fewer than the requested `n`) only produces a
`logger.warning` — neither reaches `report.json`. So a report cannot distinguish
"the model got it wrong" from "our extractor broke" or "the stream died and the
sample was scored 0".

Scope note: this RFC deliberately targets **math and MCQ only**. Code tasks
(`human_eval` ×2, `mbpp`, `livecodebench` ×2, `scicode`) keep their existing
`pass@k` unchanged — majority voting over programs has no principled definition
short of AlphaCode-style behavioral clustering, which is a separate project.

### A. One estimator family, one module

Add `sieval/core/tasks/metrics.py` and route every task through it, replacing the
13 copies of `_pass_at_k`. This is extraction on **coupling**, not on call count:
the estimator must be identical everywhere or cross-task numbers stop being
[...]

Issue #71, opened 2026-08-06
[Feature]: give JudgementRecord.reference a durable signal for why it is absent


ethan-scitix:
### 🚀 The feature, motivation and pitch

`JudgementRecord.reference` has no durable signal for **why** it is absent, so a stored judgement cannot tell "this task's ground truth is a procedure" apart from "this sample's gold went missing".

`build_judgement_record` writes `reference` unconditionally, and `obj_to_dict` drops `None`-valued keys, so `reference=None` is present in memory and **gone** on disk:

```
procedural (human_eval):  {"rollouts":[...], "n_rollouts":1, "n_correct":1, "extra":{"entry_point":"..."}}
value      (gpqa):        {"reference":"B", "rollouts":[...], "n_rollouts":1, "n_correct":1}
```

The only way to perceive "no value-form ground truth" is to notice a missing key — and key absence collapses two causes that want different responses:

| | cause | cardinality | wanted response |
|---|---|---|---|
| (a) | the task's truth is a *procedure* (test suite, rubric) | constant for the whole task | declared, visible before reading any shard |
| (b) | this sample's gold is missing or malformed | per sample | an anomaly, reported |

`prediction` does not have this problem, and the contrast is the pitch. Its `None` has exactly one cause (extraction failed), so a single derived boolean carries it — and `records.py` calls `extracted` "The durable signal" precisely because `None` does not survive serialization. Crucially, `extracted` is not valuable on its own: it is valuable because **`detect_extraction_failure` reads it**. `reference` has neither half.

This is the same hazard `_checked_metrics` already guards for metrics, in its own words — "a `None` metric would be absent on disk, turning 'not measured' into 'never existed'". That argument was never applied to `reference`.

Proposal, in this order (the order is a dependency, not a preference):

[...]

PR #78, opened 2026-08-07
feat(tasks): add ComplexConstraints (dataset + 0-shot rubric-judged task)

ethan-scitix:
## Type

## Summary

- Adds **ComplexConstraints** (Surge AI, [arXiv:2606.09118](https://arxiv.org/abs/2606.09118),
  [`surgeai/ComplexConstraints`](https://huggingface.co/datasets/surgeai/ComplexConstraints), CC-BY-4.0):
  75 multi-constraint instruction-following prompts, each carrying 10–40 atomic rubric criteria
  (1,559 total), graded by an LLM judge rather than exact match. Two new registry entries
  (`complex_constraints` dataset, `complex_constraints_0shot_gen` task), no new dependencies.
- **Headline metric = task pass rate** (response satisfies *every* criterion) — the metric the
  paper's public 75-prompt leaderboard reports (its Table 1). The mean per-criterion pass rate is
  reported in both its published **macro** form (Table 3 caption) and the pooled **micro** form;
  the two genuinely differ because rubric sizes vary 10–40 per prompt.
- **`status="experimental"`, and the reason is score-affecting:** upstream publishes *no evaluation
  code and no judge prompt*. The paper names GPT-5-mini as the judge and defines the metrics,
  nothing more, and the dataset card adds nothing — so the rubric prompt and verdict parsing are
  **authored by this port**. Same posture as `aa_lcr`, one notch weaker (aa_lcr's templates are at
  least the upstream card's own snippets).
- **Leaderboard alignment (run 2026-08-10): the ordering reproduces exactly, the magnitudes sit
  ~13 points high.** Table and method in the Manual test plan below; `reference_impl.notes` now
  carries the measurement rather than a target. The residual is **not** judge choice among capable
  judges — three of them across two vendors agree with each other about as well as each agrees with
  itself, and all three land *above* Table 1 — so this port's authored rubric grades more leniently
  than whatever produced the board. That gap is open, and it is a second reason to ship
  `experimental`.
- Loader detail: the Hub repo ships one wide CSV whose 40 sparse `criterion_{i}` columns are
  collapsed into a single `criteria` list (a reshape, not a rename-for-uniformity — a 40-key
  mostly-absent TypedDict is unusable as a sample type). The card's `configs.data_files.path`
  names `ComplexConstraints_Benchmark_Set.csv`, which **does not exist** (casing), so
  `load_dataset("surgeai/ComplexConstraints")` cannot resolve the file at all; the loader reads the
  staged snapshot by its real name.

### Two design calls worth reviewing

**One judge call per rollout, not one per criterion.** Upstream never states its call structure, and
"atomic criteria" would argue for per-criterion calls. The framework decides it: `iter_grader_outputs`
collects `extra[GRADER_OUTPUT_KEY]` only when it is a `Mapping`, so a list of N grader outputs is
skipped silently and that spend vanishes from `profile.json` — and tasks may not modify `core/`.
Batching keeps a rollout's whole verdict set in the single `ModelOutput` the profiler expects. The
verdicts are emitted as an *indexed* PASS/FAIL list so misalignment is detectable rather than
silently shifting neighbours.

**Unreadable verdicts are counted, not just scored.** A criterion the judge returns no readable
verdict for scores not-satisfied (an unreadable verdict must never inflate a score) but is counted
separately as `n_unparsed`, per rollout and pooled in the report — so judge format drift stays
distinguishable from a model that genuinely failed the rubric. Empty/whitespace responses satisfy
zero criteria *without* invoking the judge (`grader_output` absent there, no call made).

### Automated

- [x] Unit tests pass — 48 new tests across the three modules; full `tests/unit` suite 3219 passed
- [x] `scripts/check_preflight.py` — 22 PASS / 0 FAIL (incl. `check_meta_index_sync`, dataset
      revision pinning, task naming/tags, import policy)
- [x] `sync_meta_index.py --check` and `sync_package_stubs.py --check` both clean

### Manual

- [x] Loader run against the **real** CSV: 75 rows, 1,559 criteria total — matches the paper's
      stated counts exactly; every collapsed criterion a non-empty string.
- [x] `sieval dataset download complex_constraints` succeeds; `sieval task show
      complex_constraints_0shot_gen` then reports `Ready: yes`.
- [x] End-to-end `preprocess → infer → postprocess → feedback → report` over real samples with a
      scripted judge, on the smallest (10), largest (40) and a middle (19) rubric: verdicts
      index-aligned with **0 unparsed**, and macro (68.64) ≠ micro (68.12) as expected on uneven
      rubrics.
- [x] **Score comparison table vs the published leaderboard — run 2026-08-10.** Three candidate
      response sets (one full `sieval run` each), then re-graded by other judges offline through
      the task's own `build_grader_prompt` → `parse_verdicts` → `aggregate_metrics` — **25
      gradings in total**, the GPT-5.5 response set alone graded 13 times across four judges and
      two vendors. Every cell is a **mean over 2–4 repeats with the observed range**, never a
      single grading run: the gateway used forces `temperature: 1` on its whole `gpt-5.x` family
      and the port makes one judge call per rollout (`n` repeats the *candidate*, not the judge),
      so one grading is a draw rather than a measurement — the same judge on the same responses
      spanned 20.0–29.3 over four runs.

  | Model | Expected (Table 1, task pass %) | Actual — capable judge | Actual — mini-tier judge | Diff (capable) |
  | --- | --- | --- | --- | --- |
  | GPT-5.5 | 38.7 | **51.56** (50.7–53.3, n=3) | 25.67 (20.0–29.3, n=4) | +12.9 |
  | DeepSeek-V4-Pro | 26.7 | **18.33** (18.3–18.3, n=3) | 7.78 (3.3–13.3, n=3) | −8.4 |
  | GPT-5.4 | 4.9 | **5.78** (5.3–6.7, n=3) | 13.33 (13.3–13.3, n=3) | +0.9 |
  | *ordering vs Table 1* | | **exact — adjacent gaps 32.3 / 11.7 even at worst case** | not reproduced | |

  Two of the three models originally listed here could not be run as *candidates* at all: this
  gateway refuses the OpenAI protocol for every `claude-*` and `gemini-*` model, so Claude Opus 4.8
  and Gemini 3.1 Pro are unreachable through the task's model layer (Gemini appears above only as a
  *judge*, via its native endpoint). The substitutes span the same range of the board (4.9 → 38.7),
  which is what tests the construct. DeepSeek-V4-Pro is scored on its untruncated 60-prompt subset:
  it is a reasoning model and spent the full 16k budget on reasoning for 15 of 75 prompts, which
  `n_unextracted=15` and 15 `truncated_output` anomalies both flagged.

  **What this establishes, and what stays open.** Ordering across the full range of the board
  reproduces exactly under a capable judge and survives the noise floor — adjacent gaps of 32.3 and
  11.7 points at worst case (best draw of the lower model against the worst draw of the higher),
  against per-cell spreads ≤ 2.7 — and GPT-5.4 lands 0.9 from its published value. Magnitudes do
  not match, and the residual is **not** explained by judge choice among capable judges: a frontier
  model and two Gemini tiers land on 51.6–53.3 and agree with *each other* on 98.8–99.0% of the
  1,559 criteria, about as well as any one of them agrees with itself on a repeat (99.5–99.9%), so
  38.7 sits *below* all three. The mini-tier judge is the lone outlier, ~27 points low and one-way
  (~141 criteria it alone fails against ~35 the other way, where run-to-run noise is symmetric),
  concentrated on criteria 1.8× longer than average that enumerate an exclusion list to cross-check
  — this template tells the judge to fail what it cannot verify, so a judge below the capability
  floor produces one-way false FAILs by construction. **The grader is a capability floor, not a
  name or a price:** a flash-tier judge matched a pro-tier one to 1.3 points while a mini-tier one
  was 27 off. Two judges agreeing on *how many* criteria pass (micro 76.72 vs 76.78) still differed
  2.3× on the headline because they disagreed on *which* — an all-or-nothing metric converts
  reshuffled disagreement into a large swing. Per-judge tables in `reference_impl.notes`.

  These numbers were taken before `aca7e5a` rewrote the verdict lookahead, so they were re-derived
  against it: re-parsing all **213 persisted judge replies / 4,452 criteria** with both the pre- and
  post-`aca7e5a` parser changes **no verdict and no metric** (the tightened shape does not occur in
  a real reply from any judge here, and `n_unparsed` was 0 in all 25 gradings). The table stands
  against the branch head.

  So prefer **rank agreement** over a single-point score match when reporting alignment here, and
  note two unpinned degrees of freedom that remain on the *target* side: the leaderboard snapshot
  (2026-06-03) states no repeat count, and the live board has since moved (GPT-5.5 38.7 → 44.4) and
  now splits rows by reasoning effort — worth ~4.5 points — which the paper's Table 1 rows do not
  state.

### Required (all PRs)

- [x] No new upper-layer dependencies added to `core/` (`core/` untouched)
- [x] Deleted code verified — no code deleted

- [x] Score comparison table included — populated in Manual above: ordering reproduced exactly,
      magnitudes ~13 points high under every capable judge, residual named and open
- [x] Dataset loading tested (`sieval dataset download complex_constraints` succeeds)
- [x] Task registered — auto-discovered by the lazy registry; generated `__init__.pyi` stubs and
      `sieval/meta/index.json` regenerated and committed

### If: community/ Changes

- [x] Upstream diff documented — `sieval/community/complex_constraints.py` states in its docstring
      that upstream ships no eval code and no judge prompt, so the template and parser are authored
      by this port (contrasted explicitly with `sieval.community.aa_lcr`)
- [x] License attribution preserved — dataset is CC-BY-4.0, recorded on `@sieval_dataset(license=…)`;
      no upstream code was vendored, so there is none to attribute


ethan-scitix:
Force-pushed `8d83d918` → `91dbe6a4` (review fixes + rebase). Since a force-push
hides the delta from anyone who already read the branch, here is what moved — one
of these changes the score, so it is worth not skimming.

### Blocking fix: an echoed judge instruction scored a full rubric

`GRADER_TEMPLATE` asked for `1: <PASS|FAIL>`, and `_VERDICT_RE`'s leading
`[^\w\n]*` swallowed the `<`, so that line parsed as **PASS**. A judge restating
the format it was just handed — or cut off right after restating it — therefore
scored every criterion satisfied.

Reproduced on the real 75-prompt set with a candidate answering literally `"no."`
to the first 5 prompts (98 criteria):

| | `task_pass_rate` | `n_unparsed` |
| --- | --- | --- |
| `8d83d918` | **100.0** | 0 |
| `91dbe6a4` | 0.0 | 98 |

Both properties this port documents were violated: *"an unreadable verdict must
not inflate a score"*, and format drift staying visible. And nothing downstream
would have caught it — the `truncated_output` anomaly rule reads the **candidate**
model's finish reasons, never the grader's, so `n_unparsed == 0` was the only
signal and it read clean.

Two independent guards, because the reply is not the only place an alternation can
appear: the format example now reads `<verdict>`, and the parser rejects a verdict
followed by `|`. Verified by reverting each half separately — 4 mutants, all
killed; `test_template_does_not_self_parse` fails on the pre-fix code. Tests
48 → 56.

### Also in this push

- **Grader prompt** now declares the RESPONSE block to be material, never
  instruction. The candidate is free-form text on an instruction-following
  benchmark, so it can contain its own verdict block or `BEGIN`/`END` markers.
- **`reference_impl.notes` no longer claims the paper names GPT-5-mini as *the*
  judge.** It names it for *training* only ("produced by GPT-5-mini during
  ComplexConstraints training and CoreCraft training") and never says what graded
  Table 1. That caveat was in this PR's description but not in the metadata a
  future re-runner actually reads — so the judge is an unpinned degree of freedom
  in the comparison itself, not merely an unpinned version of a known one.
- **Table 1's leaderboard URL** (`https://surgehq.ai/benchmarks/complex-constraints`,
  from its caption) is now recorded beside the 2026-06-03 snapshot, with the note
  to align against the snapshot rather than the live board.
- **Grader `max_tokens` guidance**: room for up to 40 verdict lines *after*
  reasoning. A truncated grader biases the score down via `n_unparsed`.
- **`report()` gains `n_unextracted`** (`health_metrics`), so an empty response is
  distinguishable from a rubric the judge failed outright — `n_graded` counts
  both. Only `health_metrics`, which its own docstring places outside the sampling
  gate; RFC #74 defers `pass@k` / `maj@k` for the LLM-judged family, and
  `score_key` / `denominator_policy` are left to that sweep rather than added
  piecemeal. Noted on #74 and #71, both of which this task makes stale.

### Unchanged

Rebased onto `main` (was 12 commits behind, base #73) — no conflicts, no
behavioural change from the rebase. The **score comparison table is still
unpopulated**: it needs a real judge plus a real candidate model, and remains the
reason this ships `experimental`.

Verification: `ruff` + `ruff format` clean, `ty` clean, `check_preflight.py`
24 PASS / 0 FAIL (incl. `check_meta_index_sync` — `meta/index.json` regenerated
and committed), `tests/unit` 3985 passed, CI 9/9 green.

Reviewer note: the grader template and the parser *are* the scoring function, so
any further change to either invalidates an alignment run. If more hardening is
wanted, it belongs before the first alignment, not after.

PR #89, opened 2026-08-09
feat(metrics): roll pass@k / avg@n / maj@k out to the remaining 18 tasks

ethan-scitix:
## Type

## Summary

- PR #88 shipped the shared estimators but deliberately deferred two things: the
  **18 tasks still emitting `pass@1` / `pass@<k>` from their own `_metrics`**, and
  the leaderboard column rename that moving them implies. This does both.
- 13 math tasks (AIME ×3, HMMT ×3, Apex ×2, BRUMO, CMIMC, SMT, MATH-500,
  IMO-AnswerBench) and 5 code tasks (HumanEval ×2, MBPP, LiveCodeBench ×2) now
  route through `sieval.core.tasks.metrics`. Their `report` bodies were
  **byte-identical across all 13**, and near-identical across the 5.
- **Additive at `n > 1`:** `avg@n`, `maj@k` (math only), `pass@k`, `pass^k`,
  `self_consistency`, `n`, `k`, `n_short`. At `n = 1` — the repo default — a run
  gains three keys: `score_key`, `denominator_policy` and `n_unextracted`.
- Two new helpers in `metrics.py` (`rollout_view`, `budget_metrics`) plus
  `zero_metrics` and `health_metrics`; PlatinumBench and UGMathBench are folded
  onto them rather than left as a third and fourth copy of the block #88 set out
  to remove. **No number moves in either.**

## Related Issues

Refs #74, #88

### Automated

- [x] Unit tests pass — 3975 passed over `tests/unit tests/integration tests/acceptance`
      with CI's `-m "not stress and not benchmark"`
- [x] `scripts/check_preflight.py` — all checks PASS, including `check_task_shot_knobs`,
      `check_record_key_access` and `check_meta_index_sync`
- [x] `scripts/check_layer_imports.py` clean; `sync_meta_index.py --check` and
      `sync_package_stubs.py --check` both clean (no metadata drift)
- [x] 100% line **and branch** coverage on `sieval/core/tasks/metrics.py`
      (126 statements, 0 missed; 54 branches, 0 partial). No mutation score quoted —
      `sieval/core/CLAUDE.md` records that `mutmut` is currently unobtainable, so the
      hand-written battery below stands in for it.

### Manual

- [x] **`pass@1` does not move.** The old body summed `pass_at_k(n_rollouts, n_correct, 1)`
      over finals and divided by `total`; the new one aggregates
      `rollout_metrics(...)["pass@1"]` over the same denominator.
      `build_judgement_record` derives `n_rollouts = len(rollouts)` and
      `n_correct = sum(correct)`, which is exactly what `rollout_view` reconstructs —
      same estimator, same inputs, same denominator.
- [x] **The new assertions have teeth.** Seven mutants, each reverting one real call
      site to its pre-change form, **all killed**: dropping `normalize=normalize_vote`;
      passing `None` for answers; keying the empty fallback off `total` instead of
      `per_problem`; removing `normalize_vote`'s guard; returning the misaligned
      answer list instead of `None`; hand-listing `zero_metrics`' keys; dropping
      `n_short`. Review added four more (both `maj@k` gate halves, `avg@n` varying
      with `k`, `health_metrics` zeroed) and closed two that had **survived** — see
      the review-follow-up section below. **11 mutants, 0 survivors.**
- [x] **`strip_string` really does raise on model output** — verified directly:
      `\frac`, `\sqrt`, `x\frac`, `2\sqrt` all raise `IndexError` (it indexes into
      the string it is repairing). Pinned in `tests/unit/tasks/test__math_verify.py`.
- [x] **A live run backs the new columns.** Was the last open box; closed by the
      6-cell matrix in *Live run* below — 570 rollouts against a served model.
      Every column reproduces from the persisted rollouts under an independently
      written estimator, and `pass@1` moves by exactly `0`.

## Live run

Everything above this section is offline, and a metrics change can be green on
every unit test while still being wrong about which columns a real run emits:
presence and absence are a product of `votes` × `k == n` × `n > 1`, and that
product is only exercised end to end. So: **6 cells, 570 rollouts, ~2 min**, on
Qwen3-4B-Instruct-2507 served by sglang (tp1 × dp2). Stock harness, nothing
patched.

Each cell differs from another in exactly one thing this PR changed.

| cell | n | k | `score` | `score_key` | `pass@1` | `avg@n` | `pass@k` | `pass^k` | `maj@k` | `self_cons` | `n_short` | `n_unext` |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| MATH-500 | 4 | 4 | 78.33 | pass@1 | 78.33 | 78.33 | 86.67 | 66.67 | 80.00 | 83.33 | 0 | 9 |
| MATH-500 | 4 | 2 | 80.00 | pass@1 | 80.00 | 80.00 | 86.11 | 73.89 | **absent** | 83.33 | 0 | 10 |
| MATH-500 | 1 | 1 | 80.00 | pass@1 | 80.00 | absent | absent | absent | absent | absent | absent | 3 |
| HumanEval | 4 | 4 | 88.33 | pass@1 | 88.33 | 88.33 | 90.00 | 86.67 | **absent** | **absent** | 0 | 0 |
| GSM8K | 4 | 4 | 93.33 | **accuracy** | 90.83 | 90.83 | 93.33 | 83.33 | 93.33 | 97.50 | 0 | 0 |
| MMLU | — | — | 70.00 | score | — | — | — | — | — | — | — | 0 |

Every presence and every absence in that table is the documented contract.

**Temperature 0.7, not greedy 0** — deliberately. At temperature 0 a draw of four
is four identical rollouts: `self_consistency` pins at 100 and `maj@k` restates
`pass@1`, so the run passes while measuring nothing.

- **`pass@1` does not move — measured, delta `0.00e+00`.** The claim above was
  argued from the code; here it is replayed against real model output. The
  pre-#89 body summed `pass_at_k(n_rollouts, n_correct, 1)` over finals and
  divided by total, and `build_judgement_record` still writes both fields, so the
  old formula runs against the same records: `78.333333333333` vs
  `78.333333333333`, and the same to twelve places in all five cells. Checked
  beside it, since the equivalence means nothing if the inputs differ: the stored
  `n_rollouts` / `n_correct` agree with the rollout list on every record.
- **The single-draw headline really is separate.** GSM8K's `accuracy` (93.33,
  first rollout) and `pass@1` (90.83, `c/n` over the draw) sit 2.5 points apart on
  real output. At `n = 1` they coincide, which is what makes adopting a budget
  non-breaking for a stored score.
- **`k > n` is rejected, and it was rejecting something real.** HumanEval at
  `k=2, n=1` now exits non-zero at construction with no `report.json` written. The
  same config on pre-#89 `main` writes
  `{"score":100.0,"pass@1":100.0,"pass@2":0.0}` — the confident wrong number this
  PR describes, reproduced rather than asserted.
- **`n_short = 0` is a live counter, not a dead one.** It read 0 in every cell,
  which is indistinguishable from a counter that never fires — and a short draw
  cannot be provoked from the serving side, since sglang always returns the `n`
  choices it is asked for. So the control replays the *real* records through the
  *real* `report()` with one sample truncated 4 → 3. The untouched replay
  reproduces the on-disk report exactly; the truncated one gives `n_short` 1, the
  budget warning, `pass@k` 86.67 → 83.33 — and **`maj@k` survives**, which is
  exactly the review follow-up's new gate (`k == requested`, not
  `k == requested == n`). Pre-follow-up, one short sample in 30 would have
  removed the column for the other 29.
- **`n_unextracted` is diagnostic.** MATH-500 reported 9 of 120. Cross-referencing
  `finish_reasons` on the same records: all 120 split `{'stop': 111, 'length': 9}`
  and the 9 unextracted are `{'length': 9}` — exactly the rollouts that hit
  `max_tokens`, no false positives or negatives. It caught the run's token budget,
  which is the parser-vs-model separation working.
- **The MCQ path**, with `n: 2` on the *model* and a task that takes none: the
  warning fires once, names the 30 unscored rollouts, and names the **model**
  config as the remedy; `n_unextracted` is reported by a task with no budget at
  all; `denominator_policy` comes back `judged` against the math family's
  `requested`, so the declared-not-resolved split is visible in real artifacts.
- **`sieval leaderboard report` renders all five cells** off the new report shape
  — the migration claim (`scanner.py` reads `report["score"]` and nothing else)
  confirmed end to end rather than by inspection.

The checker does **not** import `sieval.core.tasks.metrics`: its estimators are
written from the closed forms via `math.comb`, where the code under test uses
running products, so a shared bug would have to be a bug in the algebra. It
borrows exactly one thing, `normalize_vote` — a canonicalizer, not an estimator;
reimplementing `strip_string` would be copying it rather than checking it. Every
vote-based column is therefore computed twice, once through that normalizer and
once through the identity, asserting the normalizer only ever *merges* clusters
(83.33 vs 82.50 identity on the first cell). 91 checks, 0 failures.

**Three things the run found.** The first is the only one this PR introduced, and
is fixed here; the other two are pre-existing and belong in their own issues.

1. `docs/guide/metrics.md` said a non-zero `n_short` "makes **every** figure above
   a lower bound". Only `pass@k` and `pass^k` are floored, by the `n < k` guard —
   `pass@1`, `avg@n`, `maj@k` and `self_consistency` are computed over whatever
   arrived, so a short draw moves them in *either* direction. The control above is
   the counterexample: truncating one draw of four raised `pass@1` from 78.33 to
   78.89, because the rollout that went missing happened to be a wrong one.
   **Narrowed in `3d12319`**, with a tightening pass over the page in `e448ec2`
   (words, not claims — every statement the page made it still makes).
2. `score_key` is self-referential for `mmlu` / `mmlu_pro`, which set it to
   `"score"` where the doc defines it as naming *another* key — so
   `report[report["score_key"]]` is a tautology and a consumer walking the
   indirection lands back where it started. Two of 28 report sites. Pre-existing.
3. `sieval_versions` does not identify the code that ran: this branch's reports
   carry `0.5.1.dev13+ga7f31db` where `git describe` says `v0.7.0-33-gd37da0d`,
   because `sieval/_version.py` is generated and untracked, so a fresh worktree
   falls back to stale installed metadata. It means a live-run artifact cannot
   self-identify as the output of a given branch — here the provenance argument
   had to be structural, that `avg@n` / `maj@k` / `score_key` do not exist on
   `main` at all. Pre-existing, and worth its own issue.

### Required (all PRs)

- [x] No new upper-layer dependencies added to `core/` — the two new helpers read
      `TaskContext` / record shapes, both of which live in `core.tasks` already
- [x] Deleted code verified — the 18 private `_metrics` methods and the 13
      `_pass_at_k`-style report loops have no remaining call sites (AST-checked);
      `count_short` keeps its public callers via `budget_metrics`

### If: Breaking Change

**What breaks — two renames, same class.** A run at `k=4` used to write a `pass@4`
key into `report.json`; it now writes `pass@k`, with the budget reported once as the
`n` / `k` fields. A stored leaderboard row or dashboard keyed on the literal `pass@4`
no longer matches. And `avg@k` is now **`avg@n`**: the metric takes no `k` and does
not vary with one (at `n=4, k=2` it averages four verdicts where `pass@k` estimates
over two), so two keys were wearing the same `@k` suffix for different meanings. Same
value, new name.

**Migration:** read `pass@k` and take the budget from the sibling `n` / `k` fields.
`cli/leaderboard/scanner.py` reads `report["score"]` and nothing else, so stored
reports stay readable; `score` and `pass@1` keep both their name and their value.
This is the rename #88's "Notes for review" flagged as deserving its own decision —
the point of the literal `k` is that a column keeps its identity when the budget
changes, which is what a `pass@4` / `pass@16` split cannot do.

**Version:** needs a **MINOR** bump (0.8.0), not a patch — an on-disk artifact key
changed. Staged in the pending-changelog notes.

## Notes for review

Three judgement calls worth a second opinion:

- **`maj@k` is a lower bound, not an estimate.** Votes cluster on *strings* via a new
  `sieval/tasks/_math_verify.py::normalize_vote` — `community.math.strip_string`, the
  canonicalizer this family already applies to its golds — where the grader compares
  *symbolically*. So `\dfrac{1}{2}` and `1/2` become one vote, but `0.5` and
  `\frac{1}{2}` still split. The residual bias is downward (a real majority can be
  missed; a false one cannot be manufactured), and it is stated in the docstring and
  pinned by a test. RFC #74 D.3 wanted clustering through the verifier itself; that
  needs O(n²) `math_verify` calls per problem inside `report()` and is not attempted
  here.
- **`normalize_vote` swallows exceptions on purpose.** `strip_string` is written for
  curated references and indexes into the string it is repairing. It is applied to
  raw model answers at the end of a scored run, so an un-normalized vote key costs
  one cluster while an exception costs the whole report.
- **The sampling block is gated on `n > 1`**, following PlatinumBench's precedent
  from #88. At `n = 1` there is no draw to describe: `avg@n` restates `pass@1`,
  `maj@k` restates the one verdict, `n_short` is 0 by construction. `n_unextracted`
  is deliberately **outside** that gate — see below.

And one thing the migration surfaced rather than introduced — **four code tasks had
no `k <= n` guard.** `human_eval_0shot_gen`, `human_eval_0shot_base_gen` and both
LiveCodeBench tasks accepted `k=2, n=1` and reported a confident `pass@2 = 0.0` (the
`n < k` guard in `pass_at_k`) next to a real `pass@1`. They now raise, matching MBPP,
PlatinumBench and the math family. Fixing it was not optional: the `n > 1` gate would
otherwise have silently *dropped* their `pass@k` instead of reporting a wrong one.
A second latent defect went with it — three of the five returned only
`{score, fails}` on an empty or all-failed run, so a consumer reading `pass@1` or
`timeouts` hit a `KeyError` on exactly the runs least able to afford it.

## Review follow-up (`fa54dce6`)

Six findings from review, settled in one commit rather than deferred.

- **`maj@k` no longer loses the column because one draw came back short.** The gate
  was `k == requested == n`, bundling a configuration question ("does `k` cover the
  budget" — a real ambiguity, since at `k=2, n=4` there is no answer to *which two*
  without a seed) with a run-health one ("did the draw arrive whole"). Because
  `aggregate` drops a key missing for **any** problem, one truncated sample in 501
  removed `maj@k` for the other 500 — measured — while a run where *every* sample
  failed still reported `maj@k: 0.0` via `zero_metrics`. The protection was landing
  on the rare case and not the common one. A long draw tripped it too, with
  `n_short` at 0, so the report carried no evidence of why the column vanished.
  **Final gate: `k == requested`, nothing else.** A short draw now votes and biases
  `maj@k` the way it already biases every other key, annotated by `n_short`. The
  decisive argument: `self_consistency` clusters the *same* answers with the *same*
  normalizer and had the opposite policy, so one report was giving two answers to
  whether a single draw is fit to cluster. `k < n` is refused exactly as before, so
  the truncated-to-exactly-`k` inversion stays fixed.
- **`warn_unscored_rollouts` named a remedy that raises.** It told the reader to
  "set `n` per task (`tasks.mmlu_0shot_gen.args`)"; all four tasks that emit it
  publish a single-draw number, take no `n`, and raise `TypeError` on one. It now
  names the **model** config. Parameter `knob=` → `task=`.
- **`avg@k` → `avg@n`** (see *What breaks*). Kept, not dropped: `pass@1` *estimates*
  a single draw's success rate, `avg@n` *measures* the mean of the draw that was
  paid for, and they separate as soon as a verdict stops being a bool. Now said in
  the docstring rather than left to be inferred from two equal numbers.
- **`n_unextracted` moves out of the `n > 1` gate** into its own `health_metrics()`,
  reported at all 28 report sites including the four MCQ tasks that have no `n`. It
  measures the parser, not the draw; gating it behind a sampling budget withheld it
  from exactly the configuration where a silently-stopped extractor survives longest.
- **Two test gaps, both found by mutation.** Swapping `pass_pow_k` for `pass_at_k` at
  the `rollout_metrics` call site passed the whole suite — `pass^k` was covered as a
  *function*, but the call site only asserted the key's presence, so a regression
  collapsing the pair the docs tell readers to compare would have shipped green. And
  disabling `warn_unscored_rollouts`' `logger.warning` also passed: every caller
  discards the return value, so the log line *is* the behaviour and it was untested.
- **Nits.** `mmlu`/`mmlu_pro` guarded `feedback_result` being `None` and then
  subscripted it two lines later. Their `Task[...]` report slot still said
  `dict[str, float]` while returning `score_key` / `denominator_policy` strings —
  invisible to `ty`, since generic slots are not cross-checked against stage bodies.
  Per-rollout grading in gsm8k-0shot / hendrycks / theoremqa ran sequentially, making
  a sample's worst case `n × GRADE_TIMEOUT` once those tasks gained a budget; now
  `asyncio.gather`. The short-draw warning fired at `n=1` for half the family only,
  where "short" just means zero rollouts — gated on `n > 1`.

`pass@1` still does not move: 20 randomized trials against the pre-#89 body, worst
delta 1.4e-14.


ethan-scitix:
## Update: RFC #74 wave 2 landed in the same branch (37e2a53d)

Following review discussion, two decisions were taken and one more wave shipped.

**MCQ (wave 3) stays out**, and not only because it needs item E first: those four
tasks report `accuracy` and have no `pass@1` at all, so there is nothing to "roll
out" to them. What E actually guards there is narrower than the RFC implies —
`ChatModel._agenerate_impl` merges `{**self._kwargs, **kwargs}`, so a model-level
`n` DOES reach a task that passes none of its own; `mmlu_0shot_gen.postprocess`
then does `inf.texts[0]` (`# n=1, only one choice`). The score stays a valid
single-draw accuracy; the cost is paying for n generations to score one. Whether
those benchmarks should sample at all is an open product decision, and if the
answer is no the cheaper fix is a guard, not a `k`/`n` knob.

**`maj@k` for the code family stays out** — confirmed as the right call.

**Wave 2 shipped**: `gsm8k_0shot_gen`, `gsm8k_kshot_base_gen`,
`hendrycks_math_kshot_base_gen`, `theoremqa_kshot_base_gen`. All four had the same
single-draw shape (`inf.texts[0]` + `rollouts[0]`, no knob). Their headline
deliberately keeps its **first-rollout** definition — these publish greedy
single-draw numbers — with `pass@1` = c/n beside it. At n=1 the two coincide, so
no stored score moves and the report gains exactly `score_key`.

Two supporting changes:

- `sampling_report` in `core/tasks/metrics.py` is now the whole `n > 1` block; the
  18 tasks from the first commit plus PlatinumBench fold onto it (24 sites, one
  shape). Denominator stays a per-task parameter (RFC #74 F).
- `build_prediction_record` gains `extras=`, one dict per rollout.
  `RolloutPrediction.extra` was in the declared schema but unreachable from the
  builder, so GSM8K's few-shot port kept its flexible-match answer and both
  extraction methods in the **sample-level** slot — facts about one response, which
  under `n > 1` silently meant "rollout 0's". On-disk shape change for that task's
  prediction records.

### Verification

3924 tests pass; `ruff`, `ty check`, `check_preflight.py` (all PASS),
`check_layer_imports.py`, and both sync scripts clean. `core/tasks/metrics.py` is
at 100% line **and** branch coverage. Ten more mutants, each reverting one real
call site, all killed — the four `texts[0]` caps, dropping `n=` from `agenerate`,
turning the first-rollout headline into `pass@1`, leaking the sampling block into
an `n=1` report, moving the flexible match back to the sample-level slot, and
silently zipping misaligned `extras`.

Still no live run behind the new columns; everything is offline.

### What remains in RFC #74 after this

| Item | State |
| --- | --- |
| A — one estimator module | done; `pass^k` / the generalized `G-Pass@k_tau` form is **not** implemented |
| B — name the budget | done as `n` / `k` / `n_short`; no per-sample `n_rollouts` distribution. `avg@k` is shipped alongside `pass@1`, deliberately against B's wording (see #88) |
| C — health metrics | `n_short` done; **`n_unextracted` not implemented** |
| D — consistency then majority | `maj@k` done (math only); **`self_consistency` not implemented**, and D.3's verifier-based clustering is string-based instead, documented as a lower bound |
| E — MCQ stop hard-coding rollout 0 | done for the four wave-2 math tasks; **not done for MCQ** (deferred above) |
| F — declare the denominator | passed explicitly per task and documented in code; the **policy is not recorded in the report or `TaskMeta`** |
| G — offline recompute over shards | **not implemented.** `sampling_report` takes live `TaskContext`s; `sieval leaderboard report` aggregates stored `report.json` files rather than recomputing from shards |

None of these blocks anything shipped here.

ethan-scitix:
## Update 2: four more RFC #74 items (b9d8432a, 2cab6d34)

- **`pass^k`** — `C(c,k)/C(n,k)`, all k of a random k-subset correct, beside `pass@k` whenever k > 1. The pair is the point: `pass@k` is an upper bound and RISES with sampling variance, so a model that got less reliable can score better on it. This is the key that falls instead.
- **`n_unextracted`** — rollouts whose answer could not be recovered, counted per rollout. A run whose extractor stopped matching is otherwise indistinguishable, in every other key, from one whose model got worse.
- **`denominator_policy`** — `"requested"` (finals + fails) or `"judged"` (finals only), beside `score_key` at all 25 report sites. RFC #74 F's "declare it, do not unify it" is now machine-readable instead of a code comment. Unifying would move `score` for eight tasks.
- **`self_consistency`** — modal-cluster share. The dispersion metric `maj@k` cannot be: 4/4 agreeing and 3/4 agreeing both score `maj@k` 1.0, and a model whose answer distribution widened without its mean changing moves only this key. Correctness-blind by construction. No `k == n` gate — it describes the draw that arrived, not the budget a majority is over.
- **The four MCQ tasks stop discarding rollouts.** Still **no** `k`/`n` knob — they publish a single-draw number and wave 3 is a separate decision — but `inf.texts[0]` / `rollouts[0]` is gone. A model-level `n` reaches a task that passes none of its own, so those draws were generated, billed and dropped before anything recorded them. Now every draw is extracted, graded and written to disk, the headline still scores the first alone, and `report()` warns once with the count it did not score. Raising would be wrong: one model config legitimately serves a sampling math task and a single-draw MCQ task in the same run. **No score moves.**

Two items are decided *against* rather than deferred, and the reasoning belongs in the record:

- **`G-Pass@k_tau` (the tunable-threshold sweep) will not be added.** Both endpoints ship; the sweep between them needs a `tau` configuration story that has no obviously right answer, and `ceil(tau*k)` makes many `tau` values collapse onto the same metric (k=4: 0.6 and 0.75 are both 3).
- **Offline recompute (G) will not be added.** A shipped `recompute` command makes `report.json` mutable by later code — a stored report would stop being a function of the run that produced it, which cuts against this repo's reproducibility posture (`--resume` strict match, no `--force-*`, no bypass env vars). The one real need, backfilling a specific expensive historical run, is served by a one-off script that leaves no permanent API.

### Verification

3956 tests pass; `ruff`, `ty check`, `check_preflight.py` (23/23 PASS), `check_layer_imports.py`, and both sync scripts clean. `core/tasks/metrics.py` is at 100% line **and** branch coverage (124 statements, 54 branches). Fifteen more mutants across the two commits, all killed.

### Known sharp edge, not fixed

`rollout_metrics` gates `maj@k` on `k == len(correct)`, and `len(correct)` is the **observed** rollout count, not the requested `n`. So a sample whose draw truncated to exactly `k` gets a `maj@k` over a truncated (order-biased) subsample, while a clean draw at the same `k < n` config correctly does not. Two things blunt it: `n_short` reports the truncation, and `aggregate` drops `maj@k` entirely when it is missing for *any* problem — so a mixed run self-heals and only an all-samples-truncate-to-exactly-k run is affected. Fix is to thread the requested `n` into `rollout_metrics`; happy to do it if wanted.

ethan-scitix:
## Update 3: the maj@k boundary is fixed, and `report.json` now has a doc (c736f348)

**The sharp edge from Update 2 is closed.** `rollout_metrics` gated `maj@k` on
`k == len(correct)`, and `len(correct)` is the *observed* rollout count — so at
`n=4, k=2` a sample whose draw truncated to 2 got a `maj@k` while a clean
4-rollout draw at the same config correctly did not. Inverted, and in the worse
direction: a truncated draw is whatever finished first, not a random 2 of 4.

It now takes `n_requested` and gates on `k == n_requested == observed`. Both
halves are load-bearing — one rejects a majority over a sub-sample of the budget,
the other over a draw that came back short. `sampling_report` threads the budget
through; the default falls back to the observed count for callers that already
know the draw is complete (`zero_metrics` synthesizes one).

Only the buggy case changes. **No shipped number is known to be affected**: it
needs *every* sample to truncate to exactly `k`, because a mixed run already
self-healed — with one clean sample present, `aggregate` drops a key that is
missing for any problem.

**`docs/guide/metrics.md`**, linked from the README index. `report.json` had no
documentation at all — the keys lived only in code comments, and this branch
added eight more. Beyond a reference table it carries the part a table cannot:
the pairs that have to be read together.

- `pass@k` **with** `pass^k` — the upper bound RISES with sampling variance, so
  alone it flatters a model that got less reliable. Worked example in the page:
  halving reliability moves `pass@2` by 17 points and `pass^2` by 83.
- `maj@k` **with** `self_consistency` — thresholded vs continuous; only the
  second moves when an answer distribution widens without its mean changing.
- `self_consistency` **with** `n_unextracted` — model instability vs parser
  failure, which the shared denominator otherwise conflates.
- `n_short` **with everything** — a non-zero value makes every sampling metric a
  lower bound rather than a measurement.

Plus the `pass@<k>` → `pass@k` migration note, the `denominator_policy` split
table, why `maj@k` can be missing (three conditions), and why it is a lower bound
rather than an estimate.

### Verification

3959 tests pass; `ruff`, `ty check`, `check_preflight.py` (23/23 PASS, including
`check_links` over the new doc), `markdownlint`, and both sync scripts clean.
`core/tasks/metrics.py` stays at 100% line **and** branch coverage. Three more
mutants, all killed: reverting the gate to the observed count, dropping its
completeness half, and unthreading the budget from `sampling_report`.

### Still outstanding

**No live run backs any of the new columns.** Everything on this branch is
offline — `avg@k` / `pass@k` / `pass^k` / `maj@k` / `self_consistency` /
`n_unextracted` have not been measured against a served model, so no score
deltas are claimed. `score` and `pass@1` keep their name *and* value by
construction (same estimator, same inputs, same denominator), which is the part
that does not need a run to assert.

This branch is 5 commits / 51 files and is still a draft; a review pass before
marking it ready seems warranted, particularly on the 25 mechanically-edited
report sites.

PR #99, opened 2026-08-13
refactor(models): split model.py by lifetime, not by size

ethan-scitix:
## Type

  **One scoped exception**, called out under Manual: `response_to_model_output` now copies the caller's `model_meta` before attaching provenance. Both current call sites pass a fresh `meta()`, so nothing on disk was ever wrong — the guarantee is restored before a caller can get it wrong.

> **Rebased onto `main` after #100 merged.** #98 has merged too, so this is no longer a stacked PR. #100 edited two of the definitions this PR relocates, so its work is carried into the new homes and *verified* rather than assumed — see Manual.

## Summary

- `model.py` held two things with **different lifetimes**: the canonical `Model` composition root, which the plane keeps evolving, and the one-cycle compatibility surface, which is scheduled to die. Kept in one file, removing the compat layer is surgery on a live module; split out, it is a file deletion.

  | file | lines | holds |
  |---|---|---|
  | `model.py` | 1260 → **664** | canonical `Model` |
  | `_legacy_bridge.py` | **454** | task kwargs ↔ `Request`/`Response`, the `ModelOutput` family |
  | `_legacy_binding.py` | **170** | `RuntimeBindingPlan` fabricated for the bare `ChatModel`/`GenModel` constructors |

- The split works because the bridge was already free-function shaped, as the review measured: `_kwargs_to_request` (212 lines) touched only `self.dialect_id` and the static `self._validate_n`; `_response_to_model_output` only `self.meta()`. Those three become parameters. Nothing else changed.
- **Direction is now one-way**, which is the property that makes the eventual delete cheap: `_legacy_bridge` imports `deployment` + `ir` and *nothing* from `model`. `_legacy_binding` is reachable only from `chat_model`/`gen_model`, so it leaves when the wrappers do.
- `uuid4()`'s non-deterministic binding fingerprint (prior review N1(a)) now sits alone in `_legacy_binding`, visible and removable, instead of buried mid-file. This PR does not change it.
- **Second commit applies this PR's own review findings.** The split had made `_legacy_bridge` the owner of `named_json_value`, which `Model.meta()` needs too — so deleting the bridge would have broken the canonical plane, the one thing the split exists to make cheap. The primitive moves to `_shared` beside `copy_json_value`, and `model`'s inbound edge on the bridge drops from six names to five. `ModelMeta` is the one thread still to be cut by hand at deletion time (`Model.meta()` is public and `tasks/ruler_0shot_gen.py` calls it outside the legacy path); the module docstring says so now instead of implying the delete is free.

## Related Issues

Refs #25 — follow-up from the PR #45 review (N6). Does **not** close the `core/binding/` split, which is still gated on step 5 moving SGLang and on settling whether `capabilities.py` is vocabulary or binder logic.

### Automated

- [x] Lint/format clean (`ruff check && ruff format --check`) — 495 files formatted
- [x] Unit tests pass (`pdm run pytest`) — **5336 collected, all green, 5 deselected** (unit + integration); `tests/unit/core/models/` alone is 815; preflight 25/25, no FAIL/WARN. One repeat run flaked on the two load-sensitive performance tests (`test_benchmark_scenarios` at 55.5% concurrency efficiency, `test_pipeline_memory_scaling` at 6.6×) while the box was busy; both re-ran green in isolation and neither imports `core/models`.

### Manual

- [x] **No behaviour change, checked rather than asserted.** Every moved definition was compared against its post-#100 `origin/main` source (`d0d2027f`), not against the pre-#100 tree. Identity is measured **on the first commit**, the one that does the moving; the second commit's single deliberate change is listed separately in the last row of this section.

  | group | result |
  |---|---|
  | `ModelUsage`, `ModelMeta`, `ModelCallMeta`, `ModelOutput`, `_optional_float`, `_optional_int`, `_pop_compatible_alias`, `_LegacyOpenAIBinding`, `_legacy_runtime_plan`, `build_legacy_openai_binding` | byte-identical |
  | `_named_json_value` → `named_json_value`, `_coerce_structured_output` | identical modulo the documented rename |
  | `validate_n`, `kwargs_to_request`, `response_to_model_output` | **AST-identical** after the documented `self` → parameter substitutions (73 top-level body statements: 4 + 55 + 14) |
  | `_checked_builder_defaults` (new in #100, stays in `model.py`) | identical modulo the rename — only its call site is repointed |

- [x] **#100's work is carried, not silently reverted.** #100 narrowed `_named_json_value`'s sequence branch and `_kwargs_to_request`'s `stop` branch to `list`/`tuple`. Because the rebase relocated both definitions, git had no conflict to mark at their *new* locations — so this was checked by reverse-mutation: widening each branch back where it now lives makes #100's own guards fail (**3 cases** on the coercer, **2** on `stop`). #100's five test names are all present and the models suite went 73 → 74 tests.
- [x] **Zero runtime import cycles** across all 27 modules in `core/models` (TYPE_CHECKING excluded), same as before. `_legacy_bridge` out-edges: `_shared`, `deployment`, `ir`.
- [x] **The relocated bridge is live code, not an unreferenced copy** — three mutations, all caught by the suite: `validate_n` accepting `n < 1`, `kwargs_to_request` dropping the binding-resource guard, and `response_to_model_output` dropping the provenance attachment (that last one is caught by `tests/unit/core/runners/test_runner.py`, outside the model tests).
- [x] **The one deliberate behaviour change is guarded.** As a method, `_response_to_model_output` built the `model_meta` mapping itself, so mutating in place was safe by construction; as a free function it *receives* one, and writing into it leaked the first response's provenance into every later output built from the same mapping — into `ModelCallMeta`, which is persisted. The new test asserts the caller's mapping is untouched and that two outputs do not share it, and was confirmed to fail without the `.copy()`.

### Required (all PRs)

- [x] No internal paths, credentials, or personal info in committed files — also scanned the two commit messages
- [x] AI-generated code has `AI-Generated Code - <model> (<provider>)` in module docstring — both new modules carry it
- [x] Deleted code verified — no remaining call sites depend on it. Six test files imported the record types from `sieval.core.models.model` and now use the public `sieval.core.models` path; one test reached for the private `_response_to_model_output` and now calls the free function. `tests/unit/cli/test_resolution.py` resolved `ModelOutput` through `sieval.core.models.model`, which only re-exports it — six paths that would have broken at bridge-deletion time for reasons unrelated to the resolver; they use `Model` now, which that module actually defines.

### Follow-up (not in this PR)

- `copy_json_value` and `named_json_value` are still two functions. After #100 they no longer disagree about containers — both take `list`/`tuple` — so the only difference left is how an error names a rejected leaf: one indexes into the list, the other stops at the enclosing key. Folding them together changes the messages callers see, so it belongs in its own change rather than in a move.

### DIFF
--- a/sieval/core/models/__init__.py
+++ b/sieval/core/models/__init__.py
@@ -3,6 +3,7 @@
 AI-Generated Code - Claude Fable 5 (Anthropic)
 """
 
+from ._legacy_bridge import ModelCallMeta, ModelMeta, ModelOutput, ModelUsage
 from .capabilities import (
     Capability,
     CapabilityKey,
@@ -78,7 +79,7 @@ from .ir import (
     UsageStats,
     normalize_chat_input,
 )
-from .model import Model, ModelCallMeta, ModelMeta, ModelOutput, ModelUsage
+from .model import Model
 from .reconcile import (
     BindingReconcileInput,
     ConnectionScope,
--- a/sieval/core/models/_legacy_binding.py
+++ b/sieval/core/models/_legacy_binding.py
@@ -1,0 +1,170 @@
+"""Binding construction for the legacy ``ChatModel``/``GenModel`` constructors.
+
+These wrappers accept a bare ``model=``/``api_base=`` pair and must still
+present a truthful :class:`RuntimeBindingPlan`, so this module fabricates one
+from an externally-owned connection rather than from a reconciled deployment.
+Kept apart from ``model`` by lifetime: nothing here is reachable from the
+canonical ``Model.bind`` path, so it goes when the wrappers do.
+
+AI-Generated Code - Claude Opus 5 (1M context) (Anthropic)
+"""
+
+from dataclasses import dataclass
+from typing import Any
+from uuid import uuid4
+
+import anyio
+from openai import AsyncOpenAI
+
+from sieval.core.types import JSONValue
+
+from ._fingerprint import fingerprint_mapping
+from .capabilities import RequestDefaults, Supported
+from .deployment import (
+    ConnectionIdentity,
+    ConnectionPool,
+    Deployment,
+    Engine,
+    ServingFacts,
+    resolve_route,
+)
+from .dialect_registry import capability_decisions_for, get_dialect_spec
+from .reconcile import RuntimeBindingPlan
+
+
+@dataclass(frozen=True)
+class _LegacyOpenAIBinding:
+    deployment: Deployment
+    pool: ConnectionPool[Any]
+    runtime_plan: RuntimeBindingPlan
+    local_limiter: anyio.CapacityLimiter | None
+    parent_limiter: anyio.CapacityLimiter | None
+
+
+def _legacy_runtime_plan(
+    *,
+    dialect_id: str,
+    requested_model_id: str,
+    deployment: Deployment,
+    identity: ConnectionIdentity,
+) -> RuntimeBindingPlan:
+    """Build honest external-Python binding evidence for legacy wrappers."""
+
+    spec = get_dialect_spec(dialect_id)
+    route = resolve_route(deployment, dialect_id, spec.connection_family)
+    decisions = capability_decisions_for(dialect_id)
+    available = frozenset(
+        key for key, decision in decisions.items() if isinstance(decision, Supported)
+    )
+    effective: dict[str, JSONValue] = {key: {} for key in sorted(available)}
+    identity_fingerprint = fingerprint_mapping(
+        {
+            "endpoint": identity.endpoint,
+            "connection_family": identity.connection_family,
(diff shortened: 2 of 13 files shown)


ethan-scitix:
> ⚠️ **This PR has no CI yet, and its green "CLEAN" status is vacuous.** `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, and this PR targets `refactor/dedup-model-primitives`. GitHub will auto-retarget it to `main` when #98 merges, and CI runs then.
>
> It is not unverified — it was verified locally on this exact branch, which includes #98's changes: 5237 tests passed (unit + integration), `ruff check` / `ruff format --check` / `ty check` clean, preflight 25/25 with no FAIL or WARN. That is the same set CI runs. But please merge #98 first and let CI go green here before merging this one.

PR #118, opened 2026-08-20
fix(models): make deterministic request seeds dialect-aware

jack-scitix-ai:
## Type

## Summary

- Add typed per-request seed-support declarations to `DialectSpec`.
- Deterministic mode previously treated per-request `seed` as universal and baked `seed: 0` into every base model configuration.
- Work on #25 exposed the problem through OpenAI Responses, but the issue is protocol-general: request-seed support differs across current and future APIs.
- Model this capability explicitly in `DialectSpec` instead of special-casing OpenAI Responses; reserved dialects must declare a supported or unsupported outcome before their binders become active.
- Freeze seed value, provenance, scope, and inheritance policy during prelaunch reconciliation.
- Reuse the same frozen decisions for model bindings, external roles, task candidates, and the persisted deterministic contract.
- Inject the automatic seed only for supporting dialects. Unsupported dialects omit automatic seeds and strip inherited automatic request seeds, while preserving explicit user values for pre-I/O dialect audit.
- Keep legacy SGLang deterministic control engine-scoped rather than injecting a request seed.
- Resolve task execution from each task’s frozen prelaunch candidate binding so reconciliation, contract evidence, and wire execution cannot drift.
- Existing deterministic runs with a baked `models.<name>.args.seed: 0` cannot resume across this change; restart them from the source configuration.
- This is a prerequisite for #25; it does not activate the reserved `openai_responses` dialect.

## Related Issues

Refs #25

### Automated

- [x] Modified files pass Ruff lint and format checks
- [x] Modified production files pass `ty check`
- [x] Relevant regression suite passes (`528 passed`)
- [x] Full non-benchmark test coverage passes (`5764 passed`, `56 skipped`)
- [x] `git diff --check` passes

### Required (all PRs)

### If: Breaking Change

### DIFF
--- a/sieval/cli/leaderboard/session.py
+++ b/sieval/cli/leaderboard/session.py
@@ -161,11 +161,13 @@ class _RequestSeedDecision:
 
     dialect_id: str
     support: RequestSeedSupport
+    # Evidence-only: records where deterministic sampling is controlled
+    # (currently derivable from dialect_id). Execution consumes the frozen
+    # final seed state, not this field.
     scope: _RequestSeedScope
     seed_present: bool
     seed: int | None
     provenance: _RequestSeedProvenance
-    remove_inherited_seed: bool
 
     @property
     def explicit_seed_present(self) -> bool:
@@ -209,7 +211,6 @@ def _resolve_deterministic_request_seed(
             seed_present=True,
             seed=explicit_seed,
             provenance=explicit_provenance,
-            remove_inherited_seed=False,
         )
     if support is RequestSeedSupport.SUPPORTED:
         return _RequestSeedDecision(
@@ -219,7 +220,6 @@ def _resolve_deterministic_request_seed(
             seed_present=True,
             seed=DETERMINISTIC_DEFAULT_SEED,
             provenance=_RequestSeedProvenance.AUTOMATIC,
-            remove_inherited_seed=False,
         )
     return _RequestSeedDecision(
         dialect_id=dialect_id,
@@ -228,11 +228,15 @@ def _resolve_deterministic_request_seed(
         seed_present=False,
         seed=None,
         provenance=_RequestSeedProvenance.NONE,
-        # The legacy path is deterministic at engine scope and never inherits
-        # an orchestration-injected request default.  Ordinary unsupported
-        # dialects must remove one inherited from a supporting parent binding.
-        remove_inherited_seed=scope is _RequestSeedScope.PER_REQUEST,
     )
+
+
+def _validated_request_seed(value: object) -> int | None:
+    """Return one typed request seed or reject an invalid explicit value."""
+
+    if value is not None and (isinstance(value, bool) or not isinstance(value, int)):
+        raise TypeError("seed must be an integer")
+    return cast(int | None, value)
 
 
 def _with_task_infer_seed(
@@ -243,15 +247,12 @@ def _with_task_infer_seed(
 
     if "seed" not in infer_args:
         return decision
-    seed = infer_args["seed"]
-    if seed is not None and (isinstance(seed, bool) or not isinstance(seed, int)):
-        raise TypeError("seed must be an integer")
+    seed = _validated_request_seed(infer_args["seed"])
     return dataclasses.replace(
         decision,
         seed_present=True,
         seed=seed,
         provenance=_RequestSeedProvenance.TASK_INFER_ARGS,
-        remove_inherited_seed=False,
     )
 
 
@@ -276,8 +277,7 @@ def _apply_request_seed_decision_to_model(
             f"{decision.dialect_id!r}, but the bound model uses "
             f"{model.dialect_id!r}"
         )
-    if "seed" in model._kwargs:
-        model = model.without_args("seed")
+    model = model.without_args("seed")
     if decision.seed_present:
         model = model.with_args(seed=decision.seed)
     return model
@@ -2564,12 +2564,7 @@ class EvalSession:
     ) -> tuple[bool, int | None]:
         if "seed" not in values:
             return False, None
-        value = values["seed"]
-        if value is not None and (
-            isinstance(value, bool) or not isinstance(value, int)
-        ):
-            raise TypeError("seed must be an integer")
-        return True, value
+        return True, _validated_request_seed(values["seed"])
 
     def _freeze_deterministic_request_seed_decisions(self) -> None:
         """Freeze every model request-default seed decision after prelaunch."""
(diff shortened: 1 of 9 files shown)


ethan-scitix (changes requested):
**Request changes** — design is sound, no live defect, but the headline behavior has no test with teeth.

Verified: `tests/unit` 5722 passed, `ruff` + `ty` clean, `--deterministic` no longer bakes a seed. Commit messages and identities clean.

### P0 — must fix here

**`Bug:` Seed-strip enforcement has zero discriminating coverage** — `sieval/cli/leaderboard/session.py:279`

Replacing the strip in `_apply_request_seed_decision_to_model` with `pass` leaves the full `tests/unit` green (**5722 passed**). Same for `args.pop("seed", None)` at `sieval/cli/leaderboard/session.py:258`. The one test naming this behavior asserts `decision.remove_inherited_seed is True` — a field, not an effect.

No live bug today (`remove_inherited_seed=True` is unreachable — every non-OpenAI spec is `RESERVED`). It's P0 because this PR owns the machinery and #25 turns it on assuming it was validated here. The `future_seedless` harness already exists — bind a derived model over a seeded base, assert `meta()["default_params"]` has no `seed`.

### P1 — needs an answer, maybe no code

**`Question:` `runtime_plan_fingerprint` embeds the endpoint, now persisted** — `sieval/cli/leaderboard/session.py:2693`

`RuntimeBindingPlan._plan_payload` (`sieval/core/models/reconcile.py:90`) includes `resolved_route.endpoint`, while `_persist_effective_config`'s docstring says "Runtime endpoints are excluded". Consequence: same script, service restarted on a new port → fingerprint rotates → `--resume` aborts where it previously succeeded. Intended? If yes, only the error message needs work (it currently blames version/dialect drift); if not, drop the fingerprint from external-role entries.

### P2 — mechanical

- **`Bug:` `_resolve_task_model` is dead** — `sieval/cli/leaderboard/session.py:3883`. Only 4 tests still reference it.
- **`Nit:` New `Model._kwargs` read is removable** — `sieval/cli/leaderboard/session.py:279`. `without_args` already no-ops on absent names; unconditional call passes all 496 leaderboard tests.
- **`Question:` `remove_inherited_seed` is never consulted** — `sieval/cli/leaderboard/session.py:168`, read only for the contract JSON, and contradicts `sglang_legacy` behavior. Drop it or branch on it. Fix alongside P0.

### P3

Resume gate now aborts after prelaunch (`sieval/cli/leaderboard/session.py:4188`) — confirm intended. `_warn_best_effort_deterministic_roles` (`sieval/cli/leaderboard/session.py:2528`) has no self-managed exemption. Duplicated seed validation at `sieval/cli/leaderboard/session.py:248` / `2571`. Both commit bodies empty — fold in the breaking-change note while the branch is rewritable.

jack-scitix-ai:
Thanks for the detailed review. I addressed the actionable findings in the latest commit.

- Removed `remove_inherited_seed` from `_RequestSeedDecision` and from the persisted contract. The frozen decision now describes only the final request-seed state. Both model defaults and standalone argument mappings are enforced by always removing any prior `seed` and then reapplying the frozen value only when `seed_present=True`.
- Added discriminating tests for both enforcement points: inherited/existing seeds are removed for an absent decision, and an existing argument seed is replaced by the frozen value for a present decision.
- External-role contract entries no longer persist the volatile `binding_id` or `runtime_plan_fingerprint`. Their stable identity is the outer `task.role` key plus `requested_model_id` and `dialect_id`. The full runtime fingerprint remains in memory for same-run postlaunch drift checks.
- Added tests showing that reconstructing the same external model or changing only its endpoint/port leaves the contract unchanged, while changing its requested model, dialect, or seed changes the contract.
- Removed the dead `_resolve_task_model` helper and its tests, removed the private `_kwargs` inspection from seed enforcement, and consolidated explicit-seed validation.
- Retained `seed_scope` intentionally as evidence-only contract metadata and documented that execution consumes the frozen final seed state rather than this field.

Two behaviors were intentionally left unchanged:

1. Resume validation intentionally remains after prelaunch reconciliation. The seed contract cannot be generated until dialect reconciliation has completed, because the resolved dialect and its request-seed declaration are contract inputs. At that point, no model client, dataset, or serving I/O has occurred, so a resume mismatch still aborts before those costs or side effects.

2. The best-effort role warning intentionally has no self-managed exemption. Inline and external role bindings are outside the Session's named auto-launch topology, and a borrowed external model is not lifecycle-managed by the Session. An endpoint may happen to target a self-managed deployment, but the Session cannot prove that it participates in the managed deterministic launch contract, so the conservative non-fatal warning remains.

ethan-scitix:
Pushed a maintainer-edit commit (`4391d27e`) to this branch clearing the six review nits — **please pull before your next push** so it doesn't get overwritten.

- Deleted the four call-site `args.pop("seed", None)` guards. `Model.with_args(concurrency_limit=None, extra=None, **kw)` is a pure kwargs merge plus a shallow copy — no limiter or `extra` side effects — so with `_apply_request_seed_decision_to_model` stripping and reapplying right after, none of them could change the result. Also collapsed the `frozen_infer_args`/`infer_args` pair in `_setup_tasks` into one dict.
- Restored the assertions that went out with `TestResolveTaskModel`, re-pointed at `_task_model_config_name`. Three branches (non-string ref, unknown model, no models defined) had no discriminating coverage left anywhere in the repo — 3 of 5 mutants survived before, 0 of 5 now.
- Added the first test driving `_setup_tasks` for a deterministic `sglang_legacy` candidate — the one path where the seed helper receives a `Model` subclass whose `with_dialect` raises. Verified against two mutants (wrong `dialect_id`, decision skipped entirely).
- `_validated_request_seed` now takes a `subject`, so a bad value reports e.g. `Task 'eval' infer_args.seed must be an integer` instead of a bare `seed must be an integer`.
- `warned.update(labels_by_binding.keys())`.

On the derived contract fields: I **kept** both and documented the reasoning on `_request_seed_contract_entry` rather than removing them. `explicit_seed_present` is safely derivable from `seed_provenance`, but dropping `seed_scope` looked wrong — once `TODO(PR-5)` lands and `sglang_native` decouples scope from `dialect_id`, re-adding it would cost a second contract rotation. Happy to drop either if you disagree; the contract key is unreleased, so reshaping it is still free.

Verification on `4391d27e`: `tests/unit tests/integration tests/acceptance` → **5865 passed** (+7), `ruff check`, `ruff format --check`, `ty check` and `git diff --check` all clean. The only failure was `test_benchmark_scenarios`, a ratio gate that flakes under parallel load and passes in isolation.

One thing left for whoever merges: the resume-incompatibility note (existing deterministic runs with a baked `models.<name>.args.seed: 0` must restart from source) lives only in the PR description. Since this repo squash-merges, it needs pasting into the squash message or it won't reach `main`'s history.

ethan-scitix (approved):

PR #120, opened 2026-08-21
fix(models): harden dialect plane contracts

jack-scitix-ai:
## Type

## Summary

- Separate legacy model type, request dialect, deployment engine identity, and managed `infer.backend` so none silently selects another.
- Validate every named binding's inherited type/dialect contract before allocation or model I/O, including unused derived bindings.
- Preserve the frozen root model type when constructing compatibility wrappers and reject explicit null identity fields.
- Resolve external endpoints by connection family and require an explicit `api_base` for non-OpenAI SDK transports.
- Expose `ChatMessage.name` and `ImagePart.detail` as auditable request leaves, and reject Chat Completions message layouts that would silently reorder inline content after tool calls.
- Generalize test transports through the dialect registry so subsequent adapter PRs do not require CLI or test allowlists.
- **BREAKING:** `infer.backend: sglang` no longer activates the temporary `sglang_legacy` bypass. Configs relying on that behavior must declare model-level `engine: sglang`; managed deployments should use the dialect selected during reconciliation.

## Related Issues

- Refs #25
- Follow-up to #45

### Automated

### Required (all PRs)

### If: Breaking Change


ethan-scitix (changes requested):
Verified: 5826 unit tests pass on this branch; the findings below are probe-verified against `origin/main`. Commit messages and identities are clean.

### Blocking

**1. `Bug:` model-level `engine` still selects the dialect, and widening it to `str` made that silent**

`ModelConfigDict.engine` went `Literal["vllm","sglang"]` → `str` and `validation.py:203` now only checks non-empty, but `session.py:1578` still branches on `root_engine == "sglang"`:

```
engine='sglang'  type=gen -> dialect='sglang_legacy'
engine='sglagn'  type=gen -> dialect='openai_completions'   # silent
```

A managed root is saved by the assertion at `session.py:2326`; an unmanaged/external root pointed at a running sglang server is not — and pre-PR the enum caught it at preflight. The comment at `:1570` ("a managed DeploymentPlan's engine is a serving fact, not a dialect selector") is narrowly scoped and survives; the PR description's "so none silently selects another" does not.

Widening itself is correct — `engine` legitimately spans a closed domain (managed: `projection.engine_id` is `infer.backend`, closed by `_TRANSLATORS = {vllm, sglang}`) and an open one (external: someone else's TGI/vendor gateway, and it feeds `deployment_fingerprint`). The fix is to stop using the same key as a *selector*: require explicit `dialect: sglang_legacy` for the bypass. `resolution.py:206` only bars that for inline bindings, so named models should be free to declare it.

**2. `Bug:` `_as_legacy_wrapper` — the declared authority reaches only 1 of 3 call sites**

`session.py:3157` gained `model_type`, but only `:3319` passes it; `:3385` (inline) and `:3411` (external role) can't, because `_model_types_by_root` is built from `named_by_root` only (`:2416`). Those two fall through to `len(requirements.input) == 1`, and with the `dialect_id` fallback removed a zero-evidence binding now raises:

- `origin/main`: `openai_chat` + `AggregatedTaskRequirements()` → `ChatModel`
- this branch: `ValueError: ... has no legacy input kind`

Reachable: `TaskRequirements.input` defaults to `None`, so `_bind_role_requirement(ctx, "grader", TaskRequirements())` yields `input == set()` (verified). No shipped task trips it — all four grader-role tasks pass `input=InputKind.CHAT` — hence the green suite.

The asymmetry is the defect: `derive_model_type` tolerates zero evidence (`resolution.py:544`, defaults to `"chat"` with an INFO log), so a *named* model binds while an *inline/external role* with identical evidence hard-fails. The two docstrings also disagree — `resolution.py:561` calls `derive_model_type` "the sole compatibility-wrapper authority", `session.py:3163` says "normalized evidence **or** the deployment root's single type". Close the gap or align the prose; either way please add a `len(requirements.input) == 0` test (the new inline test covers `> 1`, not `0`).

### Non-blocking

- `Nit:` **drop the `[Unreleased]` CHANGELOG section** — first occurrence in the file's history; pending entries are staged elsewhere and consolidated at release, so this is a double-write to reconcile at version time.
- `Nit:` `resolution.py:577` adds a third `if dialect_id == "sglang_legacy"` pre-check shadowing a `DialectSpec` field (alongside `core/tasks/task.py:42` and `session.py:193`). Not worth extracting before PR-5 retires the bypass, but please carry the `TODO(PR-5)` marker so all three stay greppable.
- `Nit:` `resolution.py:211-215` and `session.py:2298-2303` raise the same `TypeError` with the same message. `normalize_inline_model_binding` runs first and `_safe_inline_model_config` preserves `engine`, so the session-side copy is unreachable.
- `Nit:` `session.py:2307` / `:2338` call `_external_endpoint` for its exception and discard the result — reads as a resolve, not a validation. The new non-`openai_sdk` branch (`:2828`) also can't fire today (both bindable dialects are `openai_sdk`; `sglang_legacy` bindings are `continue`d earlier), so it's untested by construction — fine if intentional.

### Right

`name`/`detail` close a real audit gap — `_message_to_wire` already emitted both on `main` with no leaf, so a future dialect would have dropped them silently. The reordering rejection is correctly scoped (`_message_to_wire` writes `tool_calls` before `content`; `input.chat` is active for every `ChatInput`) and `create.assert_not_awaited()` proves it lands before I/O. Removing the static capabilities-on-bypass check is safe — `session.py:2491` still covers it, keyed on the finalized `dialect_id` rather than on `engine`.

**Recommendation: request changes.** Both blocking items are this PR's own principles left half-applied; #1 is a cheap fix, #2 can be resolved either by closing the gap or by aligning the docstrings plus a test.

jack-scitix-ai:
Thanks — I addressed both blocking findings.

### Blocking findings

1. Model-level `engine` no longer selects the request dialect. The temporary native SGLang path now requires an explicit `dialect: sglang_legacy`; an omitted dialect can no longer silently change the wire protocol.

2. Compatibility-wrapper selection now uses the frozen model type for named, inline, and external bindings. Zero-evidence bindings consistently use the shared historical Chat fallback, with dedicated inline and external regression tests.

### Non-blocking notes

- Added the requested `TODO(PR-5)` marker to the temporary `sglang_legacy` pseudo-spec branch.
- Kept the session-side inline engine validation as a defensive boundary, but both validation sites now use one shared validator and diagnostic contract.
- Kept `_external_endpoint()` calls as intentional pre-I/O validation; their return value is not needed at those call sites.
- Removed the `[Unreleased]` section per repository release convention. The breaking migration remains documented in the PR body and task-facing documentation.

During a fresh re-review of the same authority changes, I found and fixed two additional PR-scoped edge cases:

- Explicit `engine: unknown` is now rejected because `unknown` is reserved for the internal “engine identity absent” state, including through the legacy bypass.
- An unavailable task class remains fail-soft during early model-type resolution, without weakening type/dialect validation when explicit type or resolved task evidence exists.

I also added the pre-I/O SGLang `input_scoring` guard identified while verifying the protocol-selection change.

ethan-scitix (changes requested):
Re-review of `27271b41`. 5919/5920 tests pass (the one failure, `test_benchmark_scenarios`, passes in isolation — load-sensitive gate, not this PR). Commit messages and identities clean: single author `jack-scitix-ai <jack02@scitix.ai>`, no trailers, no private paths.

Both prior blocking findings are fixed and probe-verified: `engine` no longer selects a dialect (`engine: sglagn` / `vendor-engine` → `openai_completions`, no bypass), and `_model_types_by_binding` now reaches all four `_as_legacy_wrapper` call sites with a zero-evidence `chat` fallback. Prior nits on the duplicate `TypeError` and on `_external_endpoint` are adequately answered — dropping both.

### Blocking

**`Bug:` `dialect: sglang_legacy` fails static validation, so `--dry-run` rejects a config the real run accepts**

```
# pre-PR config (type: gen, engine: sglang):
dry-run -> Model 'm' declares engine: sglang with type: gen but omits dialect.
           ... set dialect: sglang_legacy to preserve the native /generate path

# follow that verbatim:
dry-run -> Model 'm': unknown dialect 'sglang_legacy'; expected one of: openai_chat, ...
```

`sglang_legacy` isn't in `DIALECT_SPECS`, and `sieval/cli/validation.py:299-303` rejects every unregistered dialect. The pseudo-spec branch exists at `resolution.py:598`, `core/tasks/task.py:43`, `:84` and `session.py:195` — `cli/validation.py` never needed one, because before this PR nobody wrote the value by hand. Now it's the mandatory spelling in the CHANGELOG, three ppl task docstrings, `meta/index.json` and two error messages. `sieval eval` without `--dry-run` never calls `validate_eval_config`, so the real run accepts it: preflight ends up stricter than runtime, with no workaround but to skip preflight. `run_dry_run` also gates the later checks on `if result.ok:` (`validation.py:849`), so the user sees only the misleading line.

Fix: add the same `TODO(PR-5)` branch to `_validate_capabilities`, and restore the capabilities-on-bypass rejection this PR deleted there — keyed on the declared dialect now, mirroring `session.py:2608-2615`. Plus a `--dry-run` regression test; the suite covers the session path only, which is why 5919 green tests missed this.

### Non-blocking

- `Nit:` `[Unreleased]` is back (`CHANGELOG.md:8`) — removed in `8fefc919`, re-added in `de689809`, so the tree contradicts your comment. Fine either way, just say which.
- `Nit:` `TODO(PR-5)` marks 3 of 5 pseudo-spec sites; `core/tasks/task.py:43` and `:84` are bare. Free to bundle with the fix above, which adds a sixth.
- `Question:` the new `input_scoring` guard keys on `engine_id == "sglang"` exactly (`session.py:2380`), on a field this PR just made free-form — `sglang-0.4.6` and `SGLang` both slip past (verified). Managed roots stay safe via `projection.engine_id`; external ones lose the pre-I/O message and get a raw SGLang 400 instead. Intentional until PR-5 replaces the guard?

**Request changes** — the blocking item only, and it's a small fix. Worth noting the shape: all three blockers across both rounds are one authority applied at N−1 of N sites. Making the site list explicit when you fix this one would likely retire the pattern.

jack-scitix-ai:
Thanks — fixed in `775cfca`.

- Static validation now recognizes `sglang_legacy` as the temporary PR-5 pseudo-dialect, so the documented migration config reaches prelaunch instead of being rejected as unknown.
- Canonical capability declarations remain forbidden on that bypass.
- Added a full `run_dry_run()` regression test covering schema validation, imports, and capability reconciliation.
- Added the remaining `TODO(PR-5)` markers to the pseudo-spec sites.

I kept the `[Unreleased]` entry intentionally because this PR changes user-facing configuration semantics and the required migration should remain discoverable.

Regarding engine spelling: the temporary guard intentionally recognizes only the canonical engine ID `sglang`. Values such as `SGLang` or `sglang-0.4.6` are not treated as aliases. PR-5’s registered native binder will replace this string-based compatibility guard.

ethan-scitix (approved):

PR #129, opened 2026-08-26
fix(models): make request audits source-bound

### LINKED ISSUE #127: [Bug]: RequestAudit consumption cross-check is derived from the same bookkeeping path
### Your current environment

- Python version: 3.12.3
- SiEval version: `v0.8.0-16-gf2c86c3` (reproduced on `origin/main` @ `f2c86c3`)
- Relevant libraries and their versions: pytest 9.0.2
- Other information: reproduced without provider access or network I/O. The same coupled bookkeeping pattern exists in both `openai_chat` and `openai_completions`.

### 🐛 Describe the bug

`RequestAudit.finish()` is intended to cross-check the leaves declared as
`Consumed` against `PreparedRequest.consumed_paths`.

In both current OpenAI dialects, however, the two sets are populated by the
same local helper:

```python
def consume(path: str) -> None:
    if path in audit.active and path not in audit.decisions:
        audit.consumed(path)
        consumed.add(path)
```

The local set is then copied directly into the prepared request:

```python
PreparedRequest(
    ...,
    consumed_paths=frozenset(consumed),
)
```

Finally, `RequestAudit.finish()` reconstructs the other set from the
`audit.consumed()` calls and compares them:

```python
consumed = {
    path
    for path, decision in self._decisions.items()
    if isinstance(decision, Consumed)
}

if consumed != set(prepared.consumed_paths):
    ...
```

Relevant locations on `origin/main`:

- `sieval/core/models/dialect.py:210`
- `sieval/core/models/dialects/openai_chat.py:591-594`
- `sieval/core/models/dialects/openai_chat.py:688`
- `sieval/core/models/dialects/openai_completions.py:413-416`
- `sieval/core/models/dialects/openai_completions.py:485`

Because both observations are produced in the same helper call, they are not
independent. A lowering regression can omit a value from the wire body while
leaving `consume(path)` unchanged, and `finish()` will still accept it.

### Minimal reproduction on `main`

```python
from dataclasses import replace

from sieval.core.models.dialect import RequestAudit, active_request_leaves
from sieval.core.models.dialects.openai_chat import OpenAIChatDialect
from sieval.core.models.ir import (
    ChatInput,
    ChatMessage,
    Request,
    SamplingParams,
    TextPart,
)

request = Request(
    input=ChatInput(
        (
            ChatMessage(
                role="user",
                content=(TextPart("hello"),),
            ),
        )
    ),
    sampling=SamplingParams(temperature=0.25),
)

audit = RequestAudit(active_request_leaves(request))
dialect = OpenAIChatDialect(None, "probe-model")

# This dialect does not use the runtime plan during request validation.
dialect.validate_request(request, audit, None)
prepared = dialect.prepare(request, audit)

# Simulate a lowering regression: the leaf remains marked as consumed,
# but the corresponding wire field is absent.
body = dict(prepared.body)
emitted_temperature = body.pop("temperature")
tampered = replace(prepared, body=body)

audit.finish(tampered)

print(
    {
        "temperature_before_tamper": emitted_temperature,
        "temperature_after_tamper": tampered.body.get("temperature"),
        "audit_finish": "PASS",
    }
)
```

Observed result:

```console
{
    'temperature_before_tamper': 0.25,
    'temperature_after_tamper': None,
    'audit_finish': 'PASS'
}
```

The active-leaf completeness check still detects a leaf for which no audit
decision was recorded. The defect is narrower: the additional
`consumed_paths` comparison cannot independently prove that a leaf marked
`Consumed` was actually represented in the prepared wire request.

No currently shipped silent drop has been demonstrated from this issue. The
confirmed problem is that this defensive cross-check cannot catch the class of
lowering regressions it appears to guard against.

### Why this matters

Request auditing exists to make silent request-field loss fail loudly before
provider I/O. A self-derived cross-check can give future dialect changes false
confidence while accepting a request whose wire representation no longer
matches its audit decisions.

This becomes more important as additional dialects add mappings for the same
provider-neutral request leaves.

### Suggested fix

Make the prepared-request observation independent from
`audit.consumed(path)`. For example:

- record a path-to-wire-destination observation when the wire value is actually
  inserted;
- mechanically validate those destinations against the prepared body where
  possible; or
- introduce dialect-specific observation callbacks for values represented
  outside the flat request body.

Add a mutation-style regression test in which a leaf remains marked
`Consumed`, but its wire value is removed. `RequestAudit.finish()` should reject
that prepared request before SDK or network I/O.

If independent verification is intentionally out of scope, the alternative is
to remove or rename the cross-check and document that `consumed_paths` is
dialect-supplied bookkeeping rather than independent evidence.

Related to #25, which introduced the request-audit mechanism, but this specific
self-comparison weakness is not currently tracked there.

- [x] Make sure you already searched for relevant issues and documentation.

jack-scitix-ai:
## Type

## Summary

- Verify consumed request fields against immutable snapshots of the original request instead of dialect-provided expected values.
- Add source-aware verification for Chat messages, including ordering, roles, mixed content, images, tool calls, and tool results.
- Verify structured-output `type`, `schema`, `name`, and `strict` fields independently.
- Recursively freeze prepared wire bodies and reject audit/request mismatches before provider I/O.
- Derive persisted Chat and Completions request parameters from the exact body supplied to the provider SDK.
- Reject empty structured-output names and empty dialect-option keys with `RequestAuditError`.
- Share the frozen-JSON implementation between prepared requests and capability reconciliation.

## Related Issues

Fixes #127

### Automated

### Manual

- [x] Verified message lowering that drops `content` is rejected before provider I/O.
- [x] Verified response-format lowering that drops or changes `strict=False` is rejected before provider I/O.
- [x] Verified message count, order, role, text, image, tool-call, and tool-result changes are rejected.
- [x] Verified audits created for request A cannot validate or prepare request B in either OpenAI dialect.
- [x] Verified prepared request bodies reject nested mutation after audit completion.
- [x] Verified streaming and non-streaming provenance remains equal to the actual SDK payload after caller or SDK mutation.
- [x] Verified empty structured-output names and empty dialect-option keys are rejected before provider I/O.
- [x] Verified all provider interactions use fake SDK clients; no real provider was contacted.

### Required (all PRs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,28 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- Request audits now verify consumed Chat aggregates independently of their
+  lowering helpers, recursively freeze prepared wire bodies, and derive stored
+  request parameters from the exact body handed to the provider SDK.
+- Empty structured-output names and empty dialect-option keys are rejected with
+  `RequestAuditError`. Use `name=None` to request the dialect's default schema
+  name; empty option keys must be removed or renamed.
+
+### Changed
+
+- **BREAKING — custom dialect preparation evidence.** `PreparedRequest` now
+  accepts only `operation`, the complete provider `body`, and optional response
+  `context`; call `prepared.thaw_body()` once when invoking an SDK.
+  `RequestAudit.consumed()` now requires `WireObservation` or `WireVerifier`
+  evidence, and `PassthroughObservation` is removed. Identity-preserving fields
+  should use `WireObservation(destination)` so the expected value comes from
+  the audited request leaf; aggregate or transformed fields should use an
+  independent `WireVerifier`.
+
 ## [0.8.0] - 2026-08-14
 
 A minor bump, and a wide one: the on-disk record shape, several `report.json`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,28 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
-## [Unreleased]
-
-### Fixed
-
-- Request audits now verify consumed Chat aggregates independently of their
-  lowering helpers, recursively freeze prepared wire bodies, and derive stored
-  request parameters from the exact body handed to the provider SDK.
-- Empty structured-output names and empty dialect-option keys are rejected with
-  `RequestAuditError`. Use `name=None` to request the dialect's default schema
-  name; empty option keys must be removed or renamed.
-
-### Changed
-
-- **BREAKING — custom dialect preparation evidence.** `PreparedRequest` now
-  accepts only `operation`, the complete provider `body`, and optional response
-  `context`; call `prepared.thaw_body()` once when invoking an SDK.
-  `RequestAudit.consumed()` now requires `WireObservation` or `WireVerifier`
-  evidence, and `PassthroughObservation` is removed. Identity-preserving fields
-  should use `WireObservation(destination)` so the expected value comes from
-  the audited request leaf; aggregate or transformed fields should use an
-  independent `WireVerifier`.
-
 ## [0.8.0] - 2026-08-14
 
 A minor bump, and a wide one: the on-disk record shape, several `report.json`
--- a/sieval/core/models/_shared.py
+++ b/sieval/core/models/_shared.py
@@ -12,8 +12,43 @@ AI-Generated Code - Claude Opus 5 (1M context) (Anthropic)
 
 import math
 from collections.abc import Mapping
+from types import MappingProxyType
+from typing import Never, Self, cast
 
 from sieval.core.types import JSONValue
+
+
+class _FrozenJSONList(list[JSONValue]):
+    """List-shaped JSON value that rejects mutation while preserving equality."""
+
+    def _immutable(self, *args: object, **kwargs: object) -> Never:
+        del self, args, kwargs
+        raise TypeError("frozen JSON sequences do not support mutation")
+
+    def __copy__(self) -> Self:
+        return self
+
+    def __deepcopy__(self, memo: dict[int, object]) -> Self:
+        del memo
+        return self
+
+
+for _method_name in (
+    "append",
+    "clear",
+    "extend",
+    "insert",
+    "pop",
+    "remove",
+    "reverse",
+    "sort",
+    "__delitem__",
+    "__iadd__",
+    "__imul__",
+    "__setitem__",
+):
+    setattr(_FrozenJSONList, _method_name, _FrozenJSONList._immutable)
+del _method_name
 
 
 def copy_json_value(value: object, path: str) -> JSONValue:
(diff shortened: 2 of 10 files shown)


ethan-scitix:
Pushed `bdb88ec3` as a maintainer edit, closing every item from the review above. The fix itself was sound — I could not break it — so this is test-pinning plus small polish, no change to the audit design.

**The one that mattered.** Three `_ChatMessagesVerifier` guards could be deleted with the whole suite still green: the message `role` check, the `body.messages` count check, and the tool-call count check. The implementation was right — I confirmed separately that it rejects all of those — but nothing pinned them, which reintroduces this PR's own failure mode one level up: an observation nothing independently verifies. `test_messages_verifier_checks_nested_semantics` now carries one labelled mutation per field the verifier proves (role, inline text *and* its exact concatenation, image url, content-part and message counts, message order, tool-call identity/name/arguments/count, tool-result id) and reports which mutation slipped rather than just "DID NOT RAISE". Note the all-text message added to the fixture: concatenated content takes a different verifier branch from the mixed text+image one, and that branch was unpinned until it existed.

Two cases are pinned as *accepted*, since over-strictness here would break working requests before provider I/O: the honest body, and tool arguments reformatted as `{ "value" : 1 }` — structured arguments compare as JSON, so whitespace is correctly not a mismatch.

**The test that could not fail.** `test_empty_dialect_option_key_is_an_explicit_audit_rejection` matched `"non-empty"`, but `finish()` emits that same wording as a backstop — with each dialect's own check deleted the test still passed, the error simply came from `dialect.py` instead. Both dialects now additionally assert the decision recorded during `validate_request` is `Rejected`, which pins the layer. Same for the empty structured-output name.

The `_CompletionsLogprobsVerifier` bool guard is now tested directly rather than through a request, and the docstring says why: no reachable request reaches it, because whenever the expected count is positive a sibling `WireObservation` on `body.logprobs` rejects the bool first.

**One source fix.** `body.logprobs` is now emitted whenever `top_logprobs` is claimed against it. Unreachable through `Model`/`arun` — `validate_request_invariants` already requires `sampled_logprobs` alongside it — but without it a caller driving the split contract directly claims a field the body never emitted and gets a wire-level audit error instead of that invariant's clear `ValueError`. Reachable bodies are unchanged.

**Polish.** Restored the rationale comment lost when `_FrozenJSONList` moved into `_shared.py` (that bare `setattr` loop is the least self-evident code in the file); dropped the `thaw_json_value` and `_freeze_json` pass-throughs for the canonical helpers; and recorded on `PreparedRequest` that `model` and `stream_options` are the two body fields carrying no audit evidence, why that is coherent, and when to prefer `active_paths` over `active`.

**CHANGELOG removed.** This project keeps no `Unreleased` section — `CHANGELOG.md` has only ever been touched by the four `chore: release …` commits, and the release skill writes it at cut time from out-of-tree staging, where I've recorded the breaking dialect-contract change. That fully reverts `aa43f070`, so that commit is now a no-op; harmless under squash-merge, just don't be surprised it contributes nothing.

Verified after the change: 6023 unit tests pass, `ruff` and `ty` clean, all preflight checks pass, `sieval/core` coverage 98% against the 95% gate, and all 16 hand-rolled guard deletions I could think of now fail the suite (was 8 of 16). Not quoting a mutmut score — `sieval/core/CLAUDE.md` records that it cannot currently run here.

Two corrections to my review, for the record: I initially read the `schema, name, and strict require format='json_schema'` rejection as new and the `top_logprobs` audit error as a production regression. Neither is a behavior change — the former is pre-existing on `main` and the latter is blocked upstream by `validate_request_invariants`. My base-side probe only appeared to accept those structured-output combinations because base's `prepare()` did not call `raise_rejections()`; adding that self-defence to chat's `prepare` is itself a real improvement in this PR.

PR #130, opened 2026-08-31
feat(tasks): add MATH-Perturb — MATH level-5 problems under simple and hard perturbations

ethan-scitix:
## Type

## Summary

Ports [MATH-Perturb](https://github.com/Kaffaljidhmah2/MATH-Perturb) (ICML 2025, [arXiv:2502.06453](https://arxiv.org/abs/2502.06453)) @ `df4840f6` as two datasets and two tasks. 279 Level-5 MATH problems edited so the original solution method still applies (**MATH-P-Simple**) or no longer does (**MATH-P-Hard**), row-aligned by `problem_id` — the 10–25pt gap between them is the benchmark's point.

- **The grader is vendored as its own module, not reused from `deepseek_math`.** Upstream's `evaluation/` is a *fork* of DeepSeek-Math's (its own README says so): tolerance tightened 1e-3 → 1e-7, unicode normalization for o1/o3-mini, `{,}` thousands separators, `\boxed(`/`\boxed[`, spacing macros stripped, space and `\cdot` deletion disabled, plus a new gold-side extractor. Vendoring the fork is what keeps it diffable.
- **Prompt is a sieval choice, stated as such.** Upstream ships an evaluation package and *no* inference code; the paper specifies only "zero-shot CoT". Used: DeepSeek-Math's zero-shot CoT turn — the same lineage the grader was forked from, so it is the instruction its extractor was written against.
- **`report.json`**: `pass@1` headline, plus `score_seed_{train,test}` with their own population counts and intervals (the paper's second axis — which MATH split each problem was *seeded* from), plus a per-subject breakdown. No `maj@k` / `self_consistency`: an answer here is a *set* of atoms and upstream defines equivalence only pairwise, so there is no canonical vote key.
- **Not ported: the paper's third column, `Original`.** Upstream ships no such file and no `problem_id` → MATH mapping, so building it is inference rather than a port. `answer_check`'s `dataset_type="original"` arm is carried anyway, so the grader is ready if that mapping is established.
- **`status="stable"`** — the port is verified against upstream's *code* **and**, since the alignment run below, against a published row on a live run. Promoted in `b0e3b47`.

### ⚠️ Upstream trap worth knowing: `backend="lark"` cannot parse `\pi`

Upstream calls `parse_latex(s, backend="lark")` **in its own source**, which is not sympy's default. On sympy 1.14 that backend cannot parse `\pi` at all, nor comma-grouped integers (`270,000`), `\emptyset`, or interval/set unions. Over the **571 gold atoms** these two files extract to, `parse_latex` fails on **40 (7.0%) under lark vs 32 (5.6%) under ANTLR**, and the sets are *not* nested — 13 lark-only, 5 ANTLR-only (all `\begin{pmatrix}` matrices, which `math_equal` handles in a branch of its own before parsing). What fails to parse falls through to `parse_expr`, fails there too, and is decided by **string equality**.

Kept, because this is upstream's *code* rather than upstream's *environment* — the opposite of the GSM-Plus case, where a missing ANTLR runtime was an environment defect sieval repairs. Documented in `reference_impl.notes` so nobody reads the symbolic path as active on those rows.

## Related Issues

None.

### Automated

- [x] Lint/format clean (`ruff check && ruff format --check` over `sieval tests scripts`)
- [x] Type check clean (`ty check`: **0** diagnostics under `sieval/`, `tests/`, `scripts/`; `mypy --strict` at the same baseline as `math_500_0shot_gen` / `gsm_plus_0shot_gen`)
- [x] Unit tests pass — **101 new** (46 community / 15 dataset / 39 task / 1 preflight regression)
- [x] `check_preflight.py` — all checks pass **and no WARN** (was 1: see the checker fix below)

### Manual

- [x] **Differential against upstream's own functions**, executing them on shared inputs — all 558 rows plus 9043 `(row, synthetic response)` cases spanning every branch of its extractor:

  | | agreement |
  |---|---|
  | extracted gold | **558 / 558** |
  | extracted prediction | **9043 / 9043** |
  | verdict | **9043 / 9043** |

- [x] **Execution-safety divergence costs nothing.** Rerunning the same 9043 cases with `parse_latex` disabled in *each module's own globals* — which forces every symbolic comparison down the `parse_expr` path the guards sit on — upstream and this port agree **9043 / 9043**.
- [x] `sieval dataset download math_perturb_simple` and `… math_perturb_hard` both succeed; staged files match the declared sha256.
- [x] **Full 279-row run of both tasks** against a canned model on the real staged data: `fails=0`, `n_unextracted=0`, `interval_declaration_problems()` empty, seed-split counts `164 train / 115 test` and the seven subject counts reproducing upstream's own metadata exactly.
- [x] **1116 real grading cases through the offload process pool**: 0 timeouts, 0 worker deaths.
- [x] **Alignment run — done**, with results, repeats and the ruled-out alternatives in the section below. Anchors read off the paper, not the project page; the column-order trap there is real and is now recorded in `reference_impl.notes` together with the refinement that only the *closed*-model rows of that CSV were re-run.

### Required (all PRs)

- [x] **Score comparison table** — below.
- [x] Dataset loading tested (`sieval dataset download <name>` succeeds for both)
- [x] Tasks registered and resolvable (`meta/index.json` in sync; both leaves resolve to their own dataset through the shared generic base — asserted in tests, since a mistake there would silently score one set's answers against the other's questions)

### If: community/ Changes

- [x] Upstream diff documented — the fork-vs-DeepSeek-Math diff and every sieval deviation are enumerated in `sieval/community/math_perturb.py`'s module docstring, with the measured cost of each
- [x] License attribution preserved — upstream is Apache-2.0; the header names the repo, the pinned commit and the four source files. Note the **data** is more restricted than the code: upstream's README limits it to academic research and asks in bold that it never be used as training data, which is why the loaders publish a `test` split only and no `train`.

### If: New Dependency

- [x] Added to the correct PDM dependency group — `lark` in `[math]`, locked with `pdm lock --update-reuse` (diff confined to the one package; `check_deps` reports no version drift)
- [x] Justified: upstream *names* this backend in its source, so it is not interchangeable with the ANTLR runtime already pinned there. Both LaTeX backends are now required because two vendored graders name different ones. No lighter alternative — it is sympy's own optional parser backend.

## Alignment run — done, and `status` is now `stable`

The anchor is **deepseek-math-7b-rl**, not one of the stronger rows, and the choice is the
whole argument: this task's prompt *is* that model's own zero-shot CoT turn (the port lifts
`markup_question` from DeepSeek-Math's `run_subset_parallel.py`), so it is the one published
row where "the prompt is a sieval choice" does not apply. Its chat template is also
essentially its raw format, so a harness that skips templating scores it the same either
way — checked: 31.18 / 14.70 through `/v1/completions` with no template at all, against
33.33 / 14.70 with one.

sglang on one H100, greedy `temperature=0`, `max_tokens=3072`, `n=1`, client concurrency 32.

### deepseek-math-7b-rl — reproduces

| | run 1 | run 3 | run 4 | mean | Table 1 |
|---|---|---|---|---|---|
| **MATH-P-Simple** All | 33.33 | 32.97 | 33.69 | **33.33** | **33.33** |
| … train | 35.37 | 36.59 | 36.59 | 36.18 | 35.37 |
| … test | 30.43 | 27.83 | 29.57 | 29.28 | 30.43 |
| **MATH-P-Hard** All | 14.70 | 13.98 | 13.98 | **14.22** | **13.62** |
| … train | 12.20 | 14.02 | 14.02 | 13.41 | 15.85 |
| … test | 18.26 | 13.91 | 13.91 | 15.36 | 10.43 |

The claim is that the published pair is **inside the measurement**, not that it is a digit
match — and a single run must not be quoted as one. Run 2, differing *only* in client
concurrency (8 rather than 32), lands 30.82 / 12.54: the batching regime is worth ~2.5 points
on this model. Same-config range is 0.72 on both `All` arms.

The seed-split cells scatter more than the headlines do — Hard `test` runs 13.91–18.26 against
a published 10.43 — which is what 115 problems at a ~14% rate buys you, and the reason the
`All` column is the one to read. Run 1 alone puts Simple at 33.33 / 35.37 / 30.43, digit-exact
on all three; runs 3 and 4 say plainly that this was a draw, not a reproduction.

### max_tokens is the dominant free parameter, and upstream publishes none

Same model, same prompt, only the budget moving (Qwen2.5-Math-7B-Instruct):

| `max_tokens` | Simple | Hard | responses cut off (Simple / Hard) |
|---|---|---|---|
| 512 | 31.90 | 14.34 | 190 / 230 |
| 1024 | 62.01 | 31.18 | 60 / 101 |
| 2048 | 66.67 | 40.14 | 15 / 22 |
| 3072 | 66.67 | 39.78 | 10 / 18 |

~34 points of range on Simple, against ~3 for the entire prompt sweep. Fix this knob before
comparing anything here to Table 1.

### Qwen2.5-Math-7B-Instruct — does not, and it is not the port

Four same-config runs: Simple **68.10 / 66.67 / 67.03 / 66.67**, Hard **39.07 / 39.78 / 39.78
/ 39.78**, against published 51.61 / 27.24. A ~15-point residual, with these ruled out:

- **Grading** — upstream's unmodified `answer_check` re-derives every verdict (below).
- **Measurement** — same-config range 1.43 Simple, 0.71 Hard.
- **Prompt** — four arms (this one; a generic system turn, which suppresses Qwen-Math's
  built-in `\boxed{}` system prompt; no boxed-answer instruction at all; "Let's think step by
  step") span 66.67–70.25 Simple and 38.35–39.78 Hard.
- **Chat templating** — the bare text through `/v1/completions`, no template, gives
  70.25 / 40.50. *Higher*, so the "the paper ran open models untemplated" hypothesis is
  tested and refuted, not assumed.

What *does* move it is the generation budget: the published 51.61 / 27.24 falls between the
512 and 1024 arms of the sweep above. A smaller `max_tokens` than this port's default is the
most economical account of that row — but upstream publishes no budget, so it stays a
hypothesis consistent with the data rather than a reproduction. Recorded in
`reference_impl.notes` so the two rows are read differently: reproducible for the model whose
native prompt this is, indicative for the others.

## Two findings worth their own lines

**A missing `lark` degrades scoring silently — it does not fail.** `parse_latex(backend="lark")`
raises `ImportError` when the package is absent, `symbolic_equal`'s bare `except` swallows it
(upstream's control flow, kept), and every symbolic comparison falls through to string
equality. Measured on the same 1116 real responses:

| | with `lark` | without | delta | verdicts flipped |
|---|---|---|---|---|
| Qwen2.5-Math Simple | 68.10 | 62.01 | −6.09 | 19 |
| Qwen2.5-Math Hard | 39.07 | 34.05 | −5.02 | 14 |
| deepseek-math Simple | 33.33 | 28.32 | −5.02 | 14 |
| deepseek-math Hard | 14.70 | 13.62 | −1.08 | 3 |

Always understating, never loud. `lark` is in the `math` extra for exactly this reason, but an
environment that installed that extra *before* `lark` was added to it satisfies the group and
still lacks the package.

**This PR was the first task in the tree to fall out of `check_reference_kind`, and it fixed
the checker rather than the symptom.** `check_preflight.py` resolves an inherited
`build_judgement_record` through `cls.bases`, matching only `ast.Name` — and these two leaves
name their base as `MathPerturbZeroShotGenTask[TSample]`, an `ast.Subscript`. On the parent
commit the check reads all 65 tasks; with this PR it read 65 and reported 2 `unverified`, a
WARN that says "could not read this one" where a mis-declaration would be a FAIL. The
subscript is now unwrapped, with a regression test that fails on the parent. Preflight is
back to **all 67 tasks verified, zero WARN**.

### Automated

- [x] `ruff check` / `ruff format --check` clean over `sieval tests scripts`
- [x] `ty check` — **0 diagnostics**
- [x] Unit tests — 101 new (46 community / 15 dataset / 39 task / 1 preflight regression)
- [x] `check_preflight.py` — all checks PASS, **and no WARN** (was 1)

### Manual

- [x] Differential against upstream's own functions on **22,320** `(row, synthetic response)`
      cases spanning every extractor branch — gold, prediction and verdict **22320 / 22320**
- [x] Differential on **live text**: 1116 real responses from the two anchor models, graded by
      upstream's unmodified `answer_check` — **1116 / 1116** on gold, prediction and verdict
- [x] Every published report cell (`pass@1`, both seed splits, all seven subject cells, both
      `n_problems`) **re-derived from the per-sample judgements on disk** — 40 cells, 0 mismatches
- [x] Prompts checked byte-for-byte against the dataset row, and each task's prompts carry
      **its own** file's problem statements — 0 cross-variant mismatches over 4 task-runs
- [x] `sieval dataset download` for both — staged bytes match the declared sha256, which also
      match upstream's files at the pinned commit
- [x] **sympy version**: upstream pins 1.13.2, sieval runs 1.14. Over the same 1116 responses
      that moves 3–4 verdicts per model — every one `1.13.2`-False / `1.14`-True, and every one
      a complex number written with `i` (plus one interval union). At most +0.7 points, and only
      ever accepting an answer that is correct. An environment difference, like `gsm_plus`'s
      missing ANTLR runtime, not a code one.
- [x] **Alignment run** — above.

### Checklist

- [x] Score comparison table — above.
- [x] Anchor sourcing: read off the paper, not the project page. Worth adding that only the
      **closed**-model rows on that CSV were re-run; all four open-weight rows
      (Llama-3.1-8B-Instruct, Gemma-2-9b-it, Deepseek-math-7b-rl, Qwen2.5-Math-7B-Instruct) are
      digit-identical to Table 1, so the anchors usable without an API key are the same in both
      places. The column order really is `Hard/Simple/Original`, the reverse of Table 1's.

---


ethan-scitix:
Follow-up pushed as `90dcab4b`: dropped the `except Exception` arm in `feedback`, so this task now catches `TimeoutError` only and matches the convention #131 establishes across the rest of the math family. This was the only grading site in the tree carrying both handlers.

**Score-neutral, measured on the real `report()`** — and checked over this task's rich key set rather than just the headline, because a fail reaches the seed and subject denominators only through `raw_sample`:

| | before | after |
|---|---|---|
| `score` / `pass@1` | 50.0 | **50.0** |
| both `score_seed_*`, both `n_problems_seed_*` | | **identical** |
| all seven `score_type_*` | | **identical** |
| `n_problems`, `n_unextracted` | | **identical** |
| `fails` | 0 | **1** ← the signal |

What moves is intervals only, and in the fixture one *vanishes*: the `test` seed cell is left with a single judged problem and `wilson_interval` needs `0 < p < 1`, so `score_seed_test_ci95` drops — **taking its `ci95_units` entry with it**, so the pair stays a pair and `interval_declaration_problems()` is empty either way. The count stays published unconditionally. Pre-existing `_clustered_interval` semantics, reached more often now; a real 115-problem cell does not go uniform.

Why this and not an `n_grade_errors` metric: raising from `feedback` goes straight to `to_failed(reason="exception::<class>")` — no re-inference, no budget burned — so `fails` plus that reason already **is** the count.

6 new tests (a timeout still scores wrong; `ValueError`/`AttributeError`/`ImportError`/`OSError` each propagate; neutrality over ≥12 published keys). Stub counts its calls so a renamed call site fails rather than passes vacuously; restoring the swallowing arm is caught by exactly the four propagation cases. Full unit suite **6214 passed**, preflight all PASS / no WARN / 67 tasks, ruff + ty clean, `meta/index.json` and stubs untouched.

The reasoning is stated in full in the code rather than citing #131's rule section, so the two PRs can merge in either order.

ethan-scitix:
Two review follow-ups pushed: `dea606e1` and `1c80be11`. **No behaviour, no scores** — `report()` output is untouched and `meta/index.json` does not move.

## `dea606e1` — the vendored functions are now pinned to upstream's bytes

The review found the port had normalized **20 lines** of upstream's trailing whitespace and dropped two already-commented-out prints in `answer_check`. Cosmetic on its own — but the same class of edit that produced the EM SPACE regression, where one key of `_fix_unicode`'s table shipped folded to its ASCII lookalike. Nothing in the tree could see that one: `ruff` skips `sieval/community`, a unified diff renders a folded key identically to the original, and the 9043-case fidelity corpus is ASCII so it exercised no entry of that table. Pinning the table codepoint-by-codepoint closed the *instance*; this closes the *class*.

Those five sites are restored, so the module now partitions cleanly:

| | count | held by |
|---|---|---|
| byte-identical to upstream @ `df4840f6` | **18** of 23 | `UPSTREAM_FUNCTION_SHA256` |
| documented divergence | **5** | `DECLARED_DIVERGENT` |
| port-only (`_guarded_parse_expr`) | 1 | — |

`_fix_unicode`'s only remaining difference is now the documented print drop, rather than that *plus* two whitespace sites a reader has to triage first.

Two properties worth stating explicitly:

- **The digests come from upstream's own four files, not from this repo's copy.** Regenerating them from our own file would make the test tautological — it would assert "unchanged since last commit" while reading as "faithful to upstream".
- **The two maps are a total partition.** A divergence introduced into a verbatim function, *and* a helper copied in without a decision, both land in neither map and fail.

**Reverse-mutated**, since a hash test that never fires is worth nothing:

| mutation | result |
|---|---|
| re-apply the whitespace normalization (i.e. the pre-`dea606e1` state) | **4 tests fail** |
| change one character inside a string literal in `strip_string` | **its pin fails** |

The failure message says to move the function to `DECLARED_DIVERGENT` and update the docstring — explicitly *not* to refresh the hash, which would record the drift as the new reference.

Also unwraps a source line break in `_math_perturb_base`'s alignment notes left over from `87f50ca2`'s sanitize reflow. The joined string is byte-identical, which is why `meta/index.json` does not move.

## `1c80be11` — same guarantees, less prose

Pin section **138 → 98 lines**, module-docstring paragraph **10 → 6**. Dropped restatements rather than reasons: the EM SPACE motivation is given once, beside the table it explains, instead of twice. Digests truncated to their first 128 bits, which fits one entry per line and removes the parenthesized continuations `ruff format` was wrapping into three lines each — still far past what an accidental edit collides with. Both mutations above still fail identically after the trim.

## One review item deliberately **not** done here

These two tasks share the `k > n` guard (byte-identical to `math_500_0shot_gen`'s, error string included) and the grading-exception policy with the pass@k math family, but are asserted locally in `test_math_perturb_0shot_gen.py` rather than by `test_math_pass_at_k_family.py` — which is the drift vector that family exists to prevent, and #131 is about to add a `FAMILY`-wide assertion of exactly the same four exception classes.

It is **not implementable in either PR today**: #130 and #131 are both based on `3e92f6d8` and neither contains the other's code, so adding `math_perturb` to that family list on #131's branch would import modules that do not exist there.

Worth recording that the fix is **not** full `FAMILY` membership. That family's `test_postprocess_survives_a_missing_raw_sample` requires `postprocess` to degrade when `raw_sample` is absent; MATH-Perturb raises `TypeError` instead — and so does `gsm_plus_0shot_gen:255`, with the identical unguarded `ctx.raw_sample[...]`, also outside that family. The reason is structural: both extractors are **question-dependent** (`extract_math_answer` tests `'separated by commas' in question`), while the 13 family members' `postprocess` is not, so they degrade trivially. That is a consistent house pattern, not a defect in either.

So the follow-up is a **narrow** list covering only the two contracts that genuinely generalize — the `k > n` guard and grader-exception propagation. Cheapest moment is when #131 rebases onto a `main` that already contains this PR, since it is editing `test_math_pass_at_k_family.py` and `.claude/rules/tasks.md` anyway; that rule's "Asserted over the whole pass@k math family" line wants updating at the same time.

## Gates

`ruff` / `ruff format` / `ty` clean · preflight **all PASS, no WARN, 67 tasks** · `sanitize.sh` passes · full unit suite **6252 passed** (+19) · `meta/index.json` and the stubs untouched.

PR #132, opened 2026-09-01
fix(models): stabilize legacy model provenance

### LINKED ISSUE #128: [Bug]: Legacy ChatModel and GenModel persist UUID-randomized plan provenance
### Your current environment

- Python version: 3.12.3
- SiEval version: `v0.8.0-16-gf2c86c3` (reproduced on `origin/main` @ `f2c86c3`)
- Relevant libraries and their versions: openai 2.9.0, pytest 9.0.2
- Other information: reproduced without sending a provider request. The example uses a localhost placeholder endpoint, and both `ChatModel` and `GenModel` reproduce the issue.

### 🐛 Describe the bug

The legacy `ChatModel` and `GenModel` compatibility constructors create a
random private scope for every model instance:

```python
private_scope = uuid4().hex
```

That random value is included in the connection identity:

```python
credential_scope=(
    f"legacy-private:{private_scope}:explicit-credential"
    if api_key is not None
    else f"legacy-private:{private_scope}:environment-credential"
),
quota_scope=f"legacy-private:{private_scope}",
```

Relevant locations on `origin/main`:

- `sieval/core/models/_legacy_binding.py:45-104`
- `sieval/core/models/_legacy_binding.py:136-146`

`_legacy_runtime_plan()` fingerprints this identity and uses it in
`binding_id`, `root_deployment_key`, and the resulting
`RuntimeBindingPlan.fingerprint`:

- `sieval/core/models/_legacy_binding.py:61-72`

The runtime-plan fingerprint is subsequently included in model provenance:

- `sieval/core/models/model.py:563-582`
- `sieval/core/models/_legacy_bridge.py:463-464`

Consequently, two compatibility wrappers created from identical user-visible
configuration receive different persisted plan fingerprints.

### Minimal reproduction on `main`

```python
import asyncio

from sieval.core.models.chat_model import ChatModel
from sieval.core.models.gen_model import GenModel


async def main() -> None:
    models = [
        ChatModel(
            model="probe-model",
            api_base="http://127.0.0.1:8000/v1",
            api_key="test",
        ),
        ChatModel(
            model="probe-model",
            api_base="http://127.0.0.1:8000/v1",
            api_key="test",
        ),
        GenModel(
            model="probe-model",
            api_base="http://127.0.0.1:8000/v1",
            api_key="test",
        ),
        GenModel(
            model="probe-model",
            api_base="http://127.0.0.1:8000/v1",
            api_key="test",
        ),
    ]

    try:
        fingerprints = [
            model.runtime_plan.fingerprint
            for model in models
            if model.runtime_plan is not None
        ]

        print("ChatModel equal:", fingerprints[0] == fingerprints[1])
        print("GenModel equal:", fingerprints[2] == fingerprints[3])
    finally:
        await asyncio.gather(*(model.aclose() for model in models))


asyncio.run(main())
```

Observed result:

```console
ChatModel equal: False
GenModel equal: False
```

No network request is made. The difference is introduced entirely during local
model construction.

### Why this matters

After a successful call, the randomized plan fingerprint is written through:

```text
RuntimeBindingPlan.fingerprint
→ Response.provenance.capabilities.plan_fingerprint
→ ModelOutput.model["provenance"]
→ model_calls[*].model.provenance
```

Two otherwise equivalent runs can therefore produce different persisted
artifacts solely because each process generated a different private UUID.

This introduces noise into:

- byte-level artifact comparisons;
- provenance diffs;
- reproducibility checks performed by downstream tooling;
- any cache or analysis keyed by the persisted plan fingerprint.

The private UUID is appropriate for runtime connection ownership, limiter
isolation, and quota scoping. The bug is allowing that volatile ownership token
to influence semantic provenance presented as reproducibility evidence.

No impact on the current strict resume gate has been demonstrated. Resume uses
separate persisted configuration and inference-plan evidence.

This report is specifically about the legacy OpenAI-compatible `ChatModel` and
`GenModel` path. `SglangGenModel` also creates a private UUID, but currently has
no `RuntimeBindingPlan` and does not publish this UUID-derived plan fingerprint
through the same provenance path.

### Suggested fix

Separate volatile runtime ownership identity from stable semantic binding
identity.

Possible implementations include:

- retaining the UUID for private pool and quota ownership while excluding it
  from persisted fingerprint payloads; or
- representing stable connection properties and volatile ownership scope as
  separate fields.

The stable fingerprint should still include replay-relevant properties such as:

- requested model;
- dialect;
- endpoint;
- connection family;
- retry policy;
- whether credentials are explicit or environment-derived.

It should not include a freshly generated UUID that has no semantic effect on
the request or deployment.

Add regression tests proving that:

1. two equivalent `ChatModel` constructions produce the same persisted
   plan/provenance fingerprint;
2. two equivalent `GenModel` constructions do the same;
3. meaningful configuration changes still change the fingerprint;
4. private connection pools and limiter ownership remain isolated.

- [x] Make sure you already searched for relevant issues and documentation.

jack-scitix-ai:
## Type

## Summary

- Legacy `ChatModel`/`GenModel` keep a UUID-scoped runtime identity for pool/quota
  isolation, but persist a stable semantic provenance plan — equivalent configs now
  produce identical run artifacts instead of UUID-randomized ones.
- `Model` carries provenance explicitly: a read-only `provenance_plan` property, an
  atomic `with_reconciled_plan(runtime_plan, provenance_plan)` for the composition
  layer, and a completeness gate in `arun` that fails before provider I/O.
- External-model reconciliation projects persisted evidence structurally: each record
  type has a field-classification policy (pinned by tests), identity references are
  replaced only in declared `sources` fields, and a runtime token found anywhere in
  semantic data fails loud — it is never rewritten.
- Malformed legacy identities, runtime/provenance semantic drift, opaque reconciled
  evidence, and unresolved runtime tokens are all rejected before anything persists.

## Related Issues

Fixes #128

### Automated

### Manual

- [x] Equivalent models retain separate runtime identities and produce identical
      persisted provenance (also verified under `python -O`)
- [x] Semantic configuration changes still change persisted provenance
- [x] Incomplete provenance fails before provider I/O

### Required (all PRs)

### DIFF
--- a/sieval/cli/leaderboard/session.py
+++ b/sieval/cli/leaderboard/session.py
@@ -401,7 +401,11 @@ def _replace_provenance_tokens(
         re.compile(
             "|".join(
                 re.escape(runtime_value)
-                for runtime_value in sorted(replacements, key=len, reverse=True)
+                for runtime_value in sorted(
+                    replacements,
+                    key=lambda runtime_value: len(runtime_value),
+                    reverse=True,
+                )
             )
         )
         if replacements
@@ -706,16 +710,17 @@ def _project_external_deployment_plan(
 
         projected = _replace_provenance_tokens(evidence, replacements)
         assert isinstance(projected, Mapping)
+        projected_evidence = cast(Mapping[str, JSONValue], projected)
         if evidence.get("source") == "external_runtime_plans":
-            projected_fingerprints = projected.get("plan_fingerprints")
+            projected_fingerprints = projected_evidence.get("plan_fingerprints")
             assert isinstance(projected_fingerprints, list)
-            projected = {
-                **projected,
+            projected_evidence = {
+                **projected_evidence,
                 "plan_fingerprints": sorted(
                     {cast(str, item) for item in projected_fingerprints}
                 ),
             }
-        outcome_evidence[capability] = cast(Mapping[str, JSONValue], projected)
+        outcome_evidence[capability] = projected_evidence
 
     recipe_parameters = _replace_provenance_tokens(
         cast(JSONValue, plan.recipe_parameters), replacements
--- a/sieval/cli/leaderboard/session.py
+++ b/sieval/cli/leaderboard/session.py
@@ -326,6 +326,127 @@ def _model_value_paths(value: object) -> tuple[str, ...]:
     return tuple(sorted(found))
 
 
+def _is_uuid_scoped_legacy_plan(plan: RuntimeBindingPlan) -> bool:
+    """Return whether a plan still carries wrapper-private pool identity."""
+
+    identity = plan.connection_identity
+    quota_scope = identity.quota_scope
+    return quota_scope.startswith(
+        "legacy-private:"
+    ) and identity.credential_scope.startswith(f"{quota_scope}:")
+
+
+def _record_provenance_replacement(
+    replacements: dict[str, str],
+    ambiguous: set[str],
+    runtime_value: str,
+    stable_value: str,
+) -> None:
+    """Record a replacement, retaining collisions for fail-loud evidence checks."""
+
+    if runtime_value == stable_value:
+        return
+    if runtime_value in ambiguous:
+        return
+    previous = replacements.get(runtime_value)
+    if previous is not None and previous != stable_value:
+        del replacements[runtime_value]
+        ambiguous.add(runtime_value)
+        return
+    replacements[runtime_value] = stable_value
+
+
+def _record_plan_provenance_replacements(
+    replacements: dict[str, str],
+    ambiguous: set[str],
+    runtime_plan: RuntimeBindingPlan,
+    provenance_plan: RuntimeBindingPlan,
+) -> None:
+    """Map every runtime-only plan identity that can appear in evidence."""
+
+    pairs = (
+        (runtime_plan.binding_id, provenance_plan.binding_id),
+        (runtime_plan.root_deployment_key, provenance_plan.root_deployment_key),
+        (runtime_plan.fingerprint, provenance_plan.fingerprint),
+        (
+            runtime_plan.verification_fingerprint,
+            provenance_plan.verification_fingerprint,
+        ),
+        (
+            runtime_plan.binding_plan_fingerprint,
+            provenance_plan.binding_plan_fingerprint,
+        ),
+        (
+            runtime_plan.deployment_plan_fingerprint,
+            provenance_plan.deployment_plan_fingerprint,
+        ),
+        (
+            runtime_plan.connection_identity.credential_scope,
+            provenance_plan.connection_identity.credential_scope,
+        ),
+        (
+            runtime_plan.connection_identity.quota_scope,
+            provenance_plan.connection_identity.quota_scope,
+        ),
+    )
+    for runtime_value, stable_value in pairs:
+        _record_provenance_replacement(
+            replacements,
+            ambiguous,
+            runtime_value,
+            stable_value,
+        )
+
+
+def _replace_provenance_tokens(
+    value: JSONValue,
+    replacements: Mapping[str, str],
+) -> JSONValue:
+    """Recursively replace runtime identities, including embedded strings."""
+
+    if isinstance(value, str):
+        projected = value
+        for runtime_value in sorted(replacements, key=len, reverse=True):
+            projected = projected.replace(
+                runtime_value,
+                replacements[runtime_value],
+            )
+        return projected
+    if isinstance(value, list):
+        return [_replace_provenance_tokens(item, replacements) for item in value]
+    if isinstance(value, Mapping):
+        projected_mapping: dict[str, JSONValue] = {}
+        typed_mapping = cast(Mapping[str, JSONValue], value)
+        for key, item in typed_mapping.items():
+            projected_key = cast(str, _replace_provenance_tokens(key, replacements))
+            if projected_key in projected_mapping:
+                raise ValueError(
+                    "external provenance projection collapses evidence key "
+                    f"{projected_key!r}"
+                )
+            projected_mapping[projected_key] = _replace_provenance_tokens(
+                item, replacements
+            )
+        return projected_mapping
+    return value
+
+
+def _contains_provenance_token(value: JSONValue, token: str) -> bool:
+    """Return whether a runtime token survives anywhere in JSON evidence."""
+
+    if isinstance(value, str):
+        return token in value
+    if isinstance(value, list):
+        return any(_contains_provenance_token(item, token) for item in value)
+    if isinstance(value, Mapping):
+        typed_mapping = cast(Mapping[str, JSONValue], value)
+        return any(
+            token in key or _contains_provenance_token(item, token)
+            for key, item in typed_mapping.items()
+        )
+    return False
+
+
 class _PR1CompatibilityServingReconciler:
     """Keep existing scoring paths behind their named response guards.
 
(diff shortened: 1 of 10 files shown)


ethan-scitix (commented):
## Review — `fix(models): stabilize legacy model provenance`

**Verdict: the fix is correct and well-tested; the findings are about how much machinery it takes to get there. Recommend Comment — nothing here blocks correctness.**

**Leak surface: clean.** Single author `jack-scitix-ai <jack02@scitix.ai>` — the same identity already on 26 commits in `origin/main`, so it's established, not a new bad mailbox. No `Co-authored-by:`/`Signed-off-by:` trailers, no internal paths in any of the three commit messages.

**Verified empirically** (base `origin/main` @ `3e92f6d8`, three commits `c68e119d..fc6dd2e2`):

- Ran the issue #128 repro against the branch. All four requirements met: two equivalent `ChatModel`s and two equivalent `GenModel`s each produce identical persisted plan fingerprints; their **runtime** plans remain distinct, so private pool and limiter ownership is still isolated; a model-id change and an explicit-vs-environment credential change both still move the fingerprint; chat and gen differ from each other.
- 2068 tests pass (`tests/unit/core/models` + `tests/unit/cli`). `ruff check`, `ruff format --check`, and `ty check` all clean.
- Coverage on the two `core/` modules: `_legacy_binding.py` 98% (missing 172, 276), `model.py` 96% (missing 174-176, 187-189, 199-201, 553). Both clear the 95% bar. I did not run mutation testing.

---

### 1. Question: the fix reconstructs information the call site already has

`sieval/core/models/_legacy_binding.py:429`

At `build_legacy_openai_binding`, `endpoint`, `api_key is not None`, `max_retries` and `private_scope` are all in scope at `_legacy_binding.py:394-404`. The stable `ConnectionIdentity` could be built right there, next to the volatile one — which is what #128 suggested ("representing stable connection properties and volatile ownership scope as separate fields").

Instead the PR builds the UUID-scoped runtime plan, then calls `_legacy_provenance_projector_for_plan()`, which re-parses the UUID back out of `quota_scope`, validates it is 32 hex characters, re-parses the credential kind out of `credential_scope`, re-derives `binding_id` and `root_deployment_key`, recomputes the initial fingerprint, and raises `ValueError` on any mismatch — all to recover facts that were ten lines above it in the same function.

The `assert provenance_projector is not None` on the next line is the tell: the caller knows it cannot fail. The recovery helper exists only so `Model.bind` can classify a plan handed in from outside.

Constructing both identities at the call site and keeping the recovery path only for `Model.bind` would remove the parsing round-trip and the invented failure paths at `:172` and `:276`. It would not remove the helper itself — `Model.bind` is public — so #8 stays either way, and session.py's projection (#2) is reconciliation's problem regardless.

### 2. Question: provenance is rewritten by regex substring replacement

`sieval/cli/leaderboard/session.py:352-450` (`_replace_provenance_tokens`)

~450 lines of `session.py` exist to substring-replace runtime identity tokens across arbitrary persisted JSON — dict keys, free-text `reason` fields, `engine_id`, `recipe_parameters`, `launch_patch`. This is rewriting the artifact that is supposed to *be* the trustworthy evidence.

I checked the sharp edges and they hold today: the longest-first sort correctly handles `quota_scope` ⊂ `credential_scope` and `binding_id` ⊂ sibling-`binding_id`; `re.sub` does not rescan its own output so replacements can't cascade; the `runtime_tokens` leak checks fail loud rather than silently; and `test_external_provenance_projection_does_not_cascade_replacements` covers the cascade case directly. Good.

The concern is that it's textual rather than structural: any future field whose value happens to contain a `binding_id` or a sha256 as a substring gets silently rewritten, and nothing in the type system says so.

### 3. Question: `arun`'s new hard failure rests on two adjacent lines staying adjacent

`sieval/core/models/model.py:695`, `sieval/cli/leaderboard/session.py:4257-4260`

I confirmed that `ChatModel(...).with_dialect(dialect_id, <reconciled plan>)` leaves `_provenance_plan = None`, and `_require_provenance_plan()` then raises:

```
RuntimeError: model provenance is incomplete: the reconciled binding and
deployment evidence must be attached by the composition layer
```

Failing before provider I/O is the right call. The only production path that produces this state is `session.py:4257`, rescued by `_with_provenance_plan` at `:4260`. That's correct today, but the safety of a hot-path `RuntimeError` now depends on those two statements never drifting apart. Consider folding the re-attachment into `with_dialect` itself, or adding a test that pins the pairing.

### 4. Nit: two schema validators that cannot fail at runtime, on the per-construction path

`sieval/core/models/model.py:163` and `sieval/cli/leaderboard/session.py:525`

`_validate_provenance_schema()` compares `dataclasses.fields(RuntimeBindingPlan)` against a hardcoded set. Dataclass fields are fixed at import, so its three `raise RuntimeError` branches are unreachable in a running process — and coverage confirms it: `model.py:174-176, 187-189, 199-201` are the only uncovered blocks in the file.

Meanwhile it runs on every model construction. I measured `_validate_provenance_plan(rt, rt)` at **72 µs** even when both arguments are the *same object* — two full `to_json_value()` serializations and 20 `json.dumps` calls to prove an object equals itself. That's the canonical (non-legacy) path, where `provenance_plan is runtime_plan` always. Per-bind rather than per-sample, so the absolute cost is small, but it's pure waste with an easy fix:

- add `if provenance_plan is runtime_plan: return` at the top of `_validate_provenance_plan`;
- move both schema checks to module import time, or to a unit test where their failure branches would actually be exercised.

Same applies to `_validate_external_provenance_schema()`, called from three sites (`session.py:649, 691, 2380`).

### 5. Nit: the stable identity round-trips into a misleading error

`sieval/core/models/_legacy_binding.py:164` and `:172-175`

Verified: passing a projected provenance plan back through `_legacy_provenance_projector_for_plan` raises

```
ValueError: legacy-private connection identity has an invalid quota scope
```

but `quota_scope="legacy-private"` is exactly the value this PR mints at `:303`. The `quota_scope == "legacy-private"` arm at `:164` deliberately catches the stable form, then `:172` rejects it as malformed. Not reachable today, and line 172 is one of only two uncovered lines in the module — but the message reads like data corruption when the real meaning is "this is a projected provenance plan, not a runtime plan." Either say that, or drop the `== "legacy-private"` arm so the plan falls through as non-legacy.

### 6. Nit: 10 `assert`s added to production code

`sieval/core/models/_legacy_binding.py:429` — `assert provenance_projector is not None`. Under `python -O` this is stripped and `None` lands in a field declared non-optional. I ran the repro under `-O`: it self-heals, because `Model._initialize` re-derives the projector when it receives `None`. So this is a type-lie, not a bug — but it should be an `if ... raise` per CLAUDE.md, or the field should be `| None`.

The nine `assert isinstance(...)` in `session.py` (`:285, 308, 412, 416, 434-436, 644-645`) narrow `JSONValue` for `ty`. They narrow values returned from `_replace_provenance_tokens`, which is a genuine runtime shape rather than a pure typing artifact, so `if ... raise TypeError` would be more honest.

### 7. Nit: CLI reaches into three more private `Model` members

`sieval/cli/leaderboard/session.py:2310, 2313, 2441, 2443` read `Model._provenance_projector` / `Model._provenance_plan`; `:4260` calls `Model._with_provenance_plan`.

There is precedent in this file — `origin/main` already reaches `_limiter` (x4), `_parent_limiter` (x2), `_kwargs` (x2) — so this is consistent with the surrounding style rather than novel. But the PR roughly doubles that surface, and CLAUDE.md's "imports imply public API… promote the name or redesign the call site" applies to attributes too. Worth promoting at least `provenance_plan` to a read-only property.

### 8. Question: `_legacy_binding` is now permanent

The module docstring used to read *"nothing here is reachable from the canonical `Model.bind` path, so it goes when the wrappers do."* The PR rewrites that line because `model.py:_initialize` now imports and calls `_legacy_provenance_projector_for_plan` for **every** `Model`, canonical ones included. `model.py` already imported `_legacy_bridge`, so this isn't new coupling in kind — but the deletability guarantee is gone. Is the `Model.bind` recovery path load-bearing enough to give that up?

### 9. Nit: PR body

The manual test item "External borrowed model resources are not closed by the session" doesn't correspond to anything in this diff — looks like a carry-over from a template or an earlier branch.

---

**Recommended action: Comment.** The fix is correct and the tests are genuinely good — 21 new tests, including fail-closed cases and the cascade guard. #1 is local and cheap — it drops the `assert` and two invented failure paths — but it doesn't subsume #2 or #8. If there's a reason the projector must reconstruct rather than receive, that reason belongs in the module docstring. Priorities in a follow-up comment.

ethan-scitix:
Priorities on the findings above, since they're not all worth the same.

**Before merge (both cheap):**

- **#4** — the schema validators are unreachable at runtime (dataclass fields are fixed at import), so they're 0% covered and the protection isn't real, while costing 72 µs per construction. Move them to import time or a test, plus an `is`-identity early-out.
- **#3** — one test pinning `session.py:4257` and `:4260` together. Nothing currently stops a refactor from separating them.

**Worth doing, not blocking:** #1 in its narrow form only — build the stable identity at the call site and drop the `assert` (~15 lines). It stops the code validating data it generated two lines earlier. Plus #5 and #9.

**Follow-up:** #2 is the one I'd most like an answer on — it's the only finding that fails *silently*, since the leak checks prove no runtime token survives, not that nothing got rewritten which shouldn't have been. A comment on the assumption plus one invariant test closes it. Then #6, #7, #8.

**If only three land: #4, #3, #2.** #1 is what I wrote the most about, but it's "should be cleaner", not "will break".

ethan-scitix:
## Review fixes pushed to this branch

I pushed two commits onto `fix/stable-legacy-model-provenance` (fast-forward, `f4007e3b..d29a22d0`) rather than asking you to round-trip them. Everything below is verified; revert either commit freely if you disagree with a call.

Your response to the first review closed all of #1–#8 — #2 in particular came back re-architected rather than defended, which is the right shape. These commits close the test-quality gap that was left and act on the `Any`/`cast` preference.

### `test(models)`: eight tests, each verified by reverse mutation

Four fail-closed branches could be **deleted with the whole suite still green**. Each row is a real mutation run against `tests/unit/core/models` + `tests/unit/cli`:

| mutation | before | after |
|---|---|---|
| `_initialize` canonical-verbatim guard deleted | 2082 passed | **caught** |
| `with_provenance_plan` same guard deleted | caught | caught |
| `with_provenance_plan` identity early-out disabled | uncovered | **caught** |
| projected-scope rejection → `return None` | 2082 passed | **caught** |
| inconsistent-binding-identity rejection removed | 2082 passed | **caught** |
| invalid-quota-scope rejection removed | 2082 passed | **caught** |
| credential-kind derivation pinned to explicit | caught | caught |
| env-credential recovery branch returns wrong value | uncovered | **caught** |
| **`binding_id` relabelled PROJECTED → SEMANTIC** | **2082 passed** | **caught** |
| per-check leak rejection disabled | 2082 passed | **caught** |

Coverage: `_legacy_binding.py` 97% → **100%**, `model.py` 99% (remainder is a pre-existing branch in `capabilities`).

**The one worth your attention** is the mislabel row. The 19 `_EXTERNAL_PROVENANCE_*_FIELDS` sets were inert: each is referenced exactly twice — its own definition, and once inside `_EXTERNAL_PROVENANCE_POLICIES`, which unions them all back together — while the projection functions route fields by hand and never read them. So I moved `binding_id` from PROJECTED to SEMANTIC, which the projector demonstrably rewrites; the union stayed identical, the completeness test could not notice, and all 2082 tests passed. `test_projection_changes_only_the_fields_labelled_projected` now runs each pure projection and asserts the changed fields land inside PROJECTED/SPECIAL/COMPUTED and never touch SEMANTIC, so the labels describe behaviour instead of intent.

Two judgement calls I want to flag:

- **I kept the `model.py` `_initialize` guard rather than deleting it as unreachable.** It cannot fire from any of its five call sites today. But it differs from the schema validators you removed for #4: those were *logically impossible* to trip (dataclass fields are fixed at import), whereas this is a private-constructor invariant any future `_initialize` caller can trip — and the callers grow with each binder. So the test drives the constructor directly and says why. Its twin in `with_provenance_plan` was already covered, so the pair was inconsistently protected.
- **The per-check leak rejection is redundant with the plan-wide backstop.** Asserting only that *something* raises passes with the per-check pass deleted, because the whole-plan check catches the same token. The test asserts the context string that distinguishes the two layers.

### `refactor(models)`: `Any`/`cast` cleanup and one honest rename

- **`_project_provenance_check` → `_reject_provenance_tokens_in_checks`.** It stopped projecting anything when check reasons became fail-loud, so it returned its input unchanged while three call sites rebuilt identical tuples through `dataclasses.replace`. Call sites now pass their original tuples through, which makes "no `DeferredCheck` field is rewritten" visible rather than implied.
- **Six of the branch's seven `cast()` calls were unnecessary.** `JSONValue` already admits `Mapping[str, JSONValue]`, so the five `cast(JSONValue, ...)` at the leak-check boundary were no-ops; the `list[str]` one is now a comprehension that is lossless after the guard above it. The survivor recovers type arguments `isinstance` erases — widening the walker to `object` would need a non-str key branch that can never run, which is the same dead-branch smell the review was about.
- **`SglangGenModel`'s two refusal overrides took `Any`; they now take `RuntimeBindingPlan`.** That immediately surfaced a test handing them `object()` — so "cannot rebind" may have been passing only because the argument was garbage. It now builds a real plan.
- Legacy credential category is derived **once** instead of twice, so the runtime and provenance identities cannot disagree about it. I confirmed a test already caught drift here, so this was hygiene, not a hole.
- Said in the source why `_EXTERNAL_PROVENANCE_PROJECTED_CHECK_FIELDS` is empty, and gave it an element type.

I left `sglang_gen_model.py`'s pre-existing `__new__ -> Any` and `cast(Any, legacy_transport)` alone — unrelated to this PR's provenance work.

### Non-regression

The issue #128 repro still satisfies all four requirements, and the persisted provenance `binding_id` is **byte-identical** before and after (`legacy:openai_chat:probe-model:9aee960a34e7210d`) — the credential-kind dedupe changed no persisted value. Full `tests/unit`: **6776 passed**. `ruff check`, `ruff format --check`, `ty check` clean.

### One thing left for you

The Test Plan still checks "External borrowed model resources are not closed by the session", which doesn't correspond to anything in this diff — carry-over from a template or an earlier branch. I didn't edit the PR body since it's yours.

PR #134, opened 2026-09-01
fix(infer): a deploy that never becomes ready must say why

ethan-scitix:
## The gap

A readiness timeout is the only deploy failure that throws its evidence away.

A process that *crashes* raises `DeployError` carrying the engine's log tail. A process that stays alive and never answers `/health` raises this, and nothing else:

```
ERROR | sieval.cli.output - Not all processes ready within 300.0s (full=running)
```

Then the process is killed. The engine's log lives at `Path.home()/".sieval"/"logs"` — outside the run directory — so on an ephemeral worker the only record of *why* the server never came up is destroyed moments after the one message that failed to quote it. A wrapper that harvests results from the run dir keeps nothing.

`full=running` also understates what is known. `status()` computes a **reason** and the summary discards it:

```python
ready_str = "ready" if ready and ready.status else phase.value
```

`connection_refused` (still loading) and `health_check_failed` (up, but unhealthy) are the difference between "raise the timeout" and "the engine is broken", and neither reaches the operator.

This matters most for an engine that spawns subprocesses. With `--dp-size N`, sglang forks a scheduler per replica; when a child dies on a bad install the parent commonly goes on waiting, so the parent PID stays alive and the *only* trace of the real error is the child's traceback in that discarded log.

## Before / after

Before:

```
Not all processes ready within 300.0s (full=running)
```

After, same failure:

```
Not all processes ready within 300.0s (full=running(connection_refused))
--- full (pid 375) cmd: sglang serve --model-path /models/Qwen3-30B-A3B --port 30000 --dp-size 4
    log: /root/.sieval/logs/full-2026-09-01_08-45-45.log ---
[2026-09-01 16:45:50] server args: dp_size=4, mem_fraction_static=None
[2026-09-01 16:46:02] Scheduler hit an exception: Traceback (most recent call last):
  File "sglang/srt/managers/scheduler.py", line 1, in run_scheduler_process
ImportError: /usr/local/lib/python3.12/dist-packages/sgl_kernel.so: undefined symbol
```

The cause, the command that produced it, and where to look — none of which was reachable before.

## What the message now carries

1. **The timeout quotes each not-ready role's log tail, path, and launch command.** The path prints even when the tail is empty, because an engine that logged nothing is itself a finding and the path is not otherwise discoverable. The tail reuses the poll pass that just ran rather than re-probing `/health`.

2. **The not-ready reason rides along in the status summary**, so per-poll progress lines and the final error both distinguish still-loading from unhealthy.

3. **A crash quotes the same block as a timeout.** `FAILED`/`STOPPED` went through its own ad-hoc format that carried the tail but neither the log path nor the command. The log sits outside the run directory either way, so both paths now share `_quote_logs`.

## What was keeping that message from being true

4. **A redrawn progress bar evicted the traceback from the tail.** `str.splitlines()` splits on `\r` as well as `\n`, so a weight-loading bar redrawn with carriage returns cost one "line" per frame — engines do this even when stdout is a redirected file, and with `--dp-size N` the children that did not die keep redrawing after the one that did has printed its traceback. On a 9 KB log the entire 30-line window went to redraws of a single bar, and the `ImportError` this branch exists to surface was not in the message at all. Fixing the split alone was not enough: the read window was sized as lines × 256 bytes, so collapsing the frames made it come back nearly empty — the window is now a byte budget. `sieval infer logs` had the identical defect and is the hand-driven route to the same log, so both go through one `_tail_lines` helper.

5. **A missing engine executable is reported as such** — *a pre-flight gate, not only a message.* Previously `Popen` raised a bare `FileNotFoundError` naming neither the binary nor the role. Resolution mirrors `Popen` exactly, because a gate that resolves differently rejects launches that would have worked: a bare name is looked up in the **child's** PATH (`os.get_exec_path(env)`, so a plan's `env:` wins over ours), and a name carrying a directory is resolved against `working_dir`. A file that exists but is **not executable**, and a **directory** at the engine path, are each reported as themselves — `Popen` raises `PermissionError` in both cases, and sending an operator to install a binary already in front of them is the exact misdiagnosis this gate exists to prevent.

6. **The gate runs before the log directory exists.** It used to run after the directory was created and after `Logging to <path>` was printed, so a rejected launch left a directory behind for a process that never started and named a path that will never exist, one line before saying no log exists.

## Verification

- Full `tests/unit` — **6366 passed**, on top of the current `main`. `ruff` and `ty` clean repo-wide.
- Every behaviour above is pinned by a test confirmed to fail when the corresponding code is reverted — reverse-mutation, all mutants killed.
- The executable-gate resolution rules are checked against the real `_launch_one`, not only mocks: an engine reachable only through a plan's `env:` PATH launches, a genuinely absent binary is still caught, and a non-executable one is reported as non-executable.
- The final commit is prose only, and provably so: after stripping docstrings, the AST of both changed files is byte-identical to the commit before it.

### DIFF
--- a/sieval/infer/deployer.py
+++ b/sieval/infer/deployer.py
@@ -43,10 +43,9 @@ _HEALTH_CHECK_TIMEOUT = 2.0
 _GRACEFUL_SHUTDOWN_TIMEOUT = 10.0
 _LOG_DIR = Path.home() / ".sieval" / "logs"
 _FAILURE_LOG_TAIL = 30
-# Bytes read back from the end of a log before taking its trailing lines.
-# Sized in bytes rather than lines × bytes-per-line: collapsing carriage-return
-# redraws (see _tail_lines) turns tens of kilobytes of progress bar into one
-# line, so a window derived from the line count comes back nearly empty.
+# Bytes read back from the end of a log before taking its trailing lines. In
+# bytes, not lines × bytes-per-line: collapsing carriage-return redraws (see
+# _tail_lines) can turn tens of kilobytes into one line, emptying such a window.
 _LOG_TAIL_WINDOW_BYTES = 256 * 1024
 
 ProgressCallback = Callable[[float, str], None]
@@ -56,15 +55,12 @@ def _tail_lines(text: str) -> list[str]:
     """Split *text* into log lines, one per line the engine actually wrote.
 
     ``str.splitlines`` splits on ``\\r`` as well as ``\\n``, so a progress bar
-    redrawn with carriage returns costs one "line" per frame. Engines redraw
-    weight-loading bars this way even when stdout is a redirected file, which
-    is enough to spend a whole tail window on redraws of a single bar and bury
-    the traceback the log was read for. Split on ``\\n`` only and keep each
-    line's last frame — what a terminal would have shown.
+    redrawn with carriage returns costs one "line" per frame — enough to spend
+    a whole tail window on one bar and bury the traceback it was read for.
+    Split on ``\\n`` only and keep each line's last frame, as a terminal would.
 
-    The trailing ``\\r`` of a CRLF line is stripped first; taking the last
-    frame of ``"line1\\r"`` would otherwise yield ``""`` and drop every line
-    of a CRLF log.
+    The trailing ``\\r`` of a CRLF line is stripped first, or the last frame of
+    ``"line1\\r"`` is ``""`` and every line of a CRLF log drops.
     """
     return [line.rstrip("\r").rsplit("\r", 1)[-1] for line in text.split("\n")]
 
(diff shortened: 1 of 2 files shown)


ethan-scitix:
Pushed `7c1afd38` — review follow-up. Two corrections and two gap-fills.

**The pre-flight executable check was rejecting launches that would have worked.** `shutil.which()` searches *this* process's PATH; `Popen` searches the child's, via `os.get_exec_path(env)`. A plan whose `env:` points PATH at the prefix holding the engine was rejected before launch, with a message asserting the opposite of the truth — the binary is installed and would have run. Reproduced against the real `_launch_one`: the same `BackendCommand` launches on the parent commit and raised `DeployError` at `3e443e53`. Reachable from user config: `InferConfigDict.env` is merged unrestricted by `inject_user_env`, and the only locked key anywhere is `VLLM_BATCH_INVARIANT`.

Resolution now mirrors `Popen`: bare name → the child's PATH, name carrying a directory → resolved against `working_dir`, since `Popen(cwd=...)` looks there rather than at the deployer's own cwd.

**A non-executable engine is no longer reported as absent.** `which()` returns `None` for a failed `X_OK` too, but `Popen` raises `PermissionError` there — telling the operator it is "not found on PATH" sends them to install a binary already in front of them, which is the exact misdiagnosis this check exists to prevent.

**The crash path now quotes the same block as the timeout path**, via `_quote_logs`. `FAILED`/`STOPPED` was still printing `--- last N lines ---` with the tail but no log path and no launch command. Both arguments for including them — the log is written outside the run directory, and the command makes a wrong engine visible — apply to a crash exactly as they do to a hang.

**The `cmd:` header had no test coverage.** Replacing the whole `isinstance(launched, list)` block with a no-op left all 431 tests passing, even though quoting the launch command is one of the stated deliverables and sits behind a type-narrowing branch. Now asserted on both the timeout and crash paths.

Verification: `tests/unit/infer` 434 passed, full `tests/unit` 6343 passed. Each of the five behaviours is pinned by a test confirmed to fail when its fix is reverted — 5/5 mutants killed.

One note for the description: "All three are reporting only; no control flow moves" was not true of change 3, and that framing is arguably why the bug got in — a new pre-flight gate was weighed as a cosmetic change. Worth restating as a gate.

PR #135, opened 2026-09-01
feat(infer): expose the readiness budget on run, and stop cutting it fine

ethan-scitix:
## The gap

`sieval run` enforces a readiness timeout that no flag can reach. The value is written three separate times:

| where | value |
|---|---|
| `LocalDeployer.deploy(timeout=...)` | `300.0` |
| `launch_model(timeout=...)` (`cli/infer/lifecycle.py`) | `300.0` |
| `sieval infer start --timeout` default | `300.0` |

`sieval run` never passes one, so it silently takes the default three frames down. `sieval infer start` has a flag; the all-in-one path — the one most likely to be handed a cold, large checkpoint — has none.

## 300s was not a budget

**The failure modes are asymmetric.** A process that dies is reported within one poll interval regardless of this value, because that is phase `STOPPED`/`FAILED`, not a timeout. This budget therefore only ever bounds a process that is **alive and silent** — still loading, or hung. Setting it too low converts a slow-but-healthy cold start into a hard failure that discards the whole GPU allocation; setting it too high only delays discovery of a genuine hang. The costs are not comparable.

**And it had no margin.** Measured on one cluster, the same 30B MoE checkpoint, identical launch line (`--tp 4 --dp-size 2 --dtype bfloat16 --mem-fraction-static 0.9 --context-length 40960`):

| run | ready at |
|---|---|
| warm page cache | **77 s** |
| warm page cache | **82 s** |
| cold | **264 s** |

A 3.4× spread from page cache and shared-storage contention alone, under a 300 s ceiling — 14% of headroom over an observed *success*. A run that legitimately loads four replicas from cold storage has no chance, and the failure it produces looks like a broken engine rather than an impatient client.

## Changes

**`--ready-timeout` on `sieval run`**, threaded through `_run_all` → `launch_model`.

Named for what it bounds rather than mirroring `infer start --timeout`: `run` also evaluates and grades, so a bare `--timeout` there would read as a budget for the whole run. The help text states it applies **per model**, since each deployment gets its own budget.

**The default moves 300s → 900s**, and all four call sites now read one named `DEFAULT_READY_TIMEOUT`, so they cannot drift apart again. The rationale lives with the constant, since the next person to touch it will ask why it is large.

900s is a judgement call, not a measurement: it clears the observed cold start by ~3.4× while still bounding a hang. **A reviewer who wants the flag without the new default can revert the constant alone** — the wiring does not depend on its value.

## Verification

- `tests/unit` failure set is **byte-identical** to the `origin/main` baseline (46, all pre-existing missing-optional-dep failures in this environment); 6235 passed.
- All four new tests fail on the parent commit. One is a drift guard asserting the deployer, `launch_model`, `_run_all` and `infer start` defaults are the *same* constant — three copies of a number is how this bug happened, and that test is what stops it recurring.
- The wiring test asserts a caller's `ready_timeout=1234.0` actually arrives as `launch_model(timeout=1234.0)`, rather than being accepted and dropped.
- `ruff check` / `ruff format` / `ty` clean on the changed files.
- Not verified on live hardware.

## Relation to the other two

Found while investigating one failed run, with [#133](https://github.com/scitix/sieval/pull/133) (a bare `NVIDIA H200` matched no recipe hardware key, silently dropping every profile param) and [#134](https://github.com/scitix/sieval/pull/134) (a readiness timeout discarded the engine log that explained it). Independent of both: this one is about the budget, #134 about what you learn when it runs out. They compose — a longer budget is only tolerable because #134 makes the eventual failure legible.

### DIFF
--- a/sieval/cli/infer/commands.py
+++ b/sieval/cli/infer/commands.py
@@ -187,12 +187,11 @@ def infer_start(
     timeout: Annotated[
         float,
         typer.Option(
-            # `--timeout` stays primary for compatibility; `--ready-timeout`
-            # is the name `sieval run` uses, accepted here so one spelling
-            # works across both commands. See the min= note in cli/run.py.
+            # `--ready-timeout` is what `sieval run` calls this, accepted here
+            # so one spelling works on both; `--timeout` stays primary.
             "--timeout",
             "--ready-timeout",
-            min=1.0,
+            min=1.0,  # see cli/run.py
             help="Seconds to wait for ready",
         ),
     ] = DEFAULT_READY_TIMEOUT,
--- a/sieval/cli/infer/commands.py
+++ b/sieval/cli/infer/commands.py
@@ -33,7 +33,11 @@ from sieval.infer.config import (
     InferPhase,
     ParamValue,
 )
-from sieval.infer.deployer import DeployError, DeployTimeoutError
+from sieval.infer.deployer import (
+    DEFAULT_READY_TIMEOUT,
+    DeployError,
+    DeployTimeoutError,
+)
 from sieval.infer.params import merge_params
 from sieval.infer.recipes import capability_model_type
 from sieval.infer.topology.models import (
@@ -183,7 +187,7 @@ def infer_start(
     timeout: Annotated[
         float,
         typer.Option("--timeout", help="Seconds to wait for ready"),
-    ] = 300.0,
+    ] = DEFAULT_READY_TIMEOUT,
     name: Annotated[
         str | None,
         typer.Option("--name", "-n", help="Override handle name"),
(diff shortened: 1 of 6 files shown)


ethan-scitix:
One reviewer question worth answering up front, since #133 in this series *does* have this problem: **does raising the default perturb `infer_plans.yaml`, and so break `--resume`'s byte comparison?**

No. Verified rather than assumed:

- `timeout` is not a field of `DeploymentPlan` or `BackendCommand`, so it cannot reach a persisted plan.
- `ready_timeout` appears in exactly four places in `sieval/`, all in `cli/run.py`: the `_run_all` parameter, the typer option, the option→`_run_all` hand-off, and the `launch_model(timeout=...)` call.

It is a client-side wait, not a launch argument. Nothing about the engine's command line or the resolved plan changes, so a run started before this commit resumes against an identical `infer_plans.yaml`. That is the opposite of #133, where the resolved launch args genuinely do change and I flagged the resume interaction there.

PR #141, opened 2026-09-09
feat(models): add OpenAI Responses API dialect

jack-scitix-ai:
## Type

## Summary

- Add an active [OpenAI Responses API](https://developers.openai.com/api/reference/resources/responses/methods/create) dialect with audited request lowering, strict response lifting, and a shared parser for streaming and non-streaming responses.
- Keep plain requests minimal: request encrypted reasoning only for active or replayed reasoning, output logprobs only when requested, and web-search sources only for hosted web search. This avoids unnecessary request failures on non-reasoning models and partially compatible endpoints.
- Implement dialect mechanics for text, reasoning, function and web-search tools, structured output, image input, logprobs, stored sessions, and opaque continuations. Capabilities lacking admission fail before I/O, while missing required response channels fail during lifting rather than being silently ignored.
- Capability-gated features on canonical configuration paths remain unavailable until the model/serving profile reports support; vendor capability profiles and broader vendor live probes remain follow-up work.
- Declare per-request seeds unsupported: deterministic mode does not inject a seed, and an explicit seed is rejected before I/O.
- Persist request evidence from the final `PreparedRequest` sent to the SDK and recognize the provider-verbatim `max_output_tokens` truncation reason.

## Related Issues

Refs #25  
Refs #113

### Automated

### Manual

- [x] Live-probed a third-party Responses-compatible MaaS endpoint with the reasoning-capable `gpt-5.6-sol` model: verified non-streaming and streaming completion, `max_output_tokens` truncation, and usage parsing. The probe did not request or assert reasoning output items.

Coverage note: reasoning-item lifting and encrypted continuation, function and web-search tools, image input, `store` / `previous_response_id`, and non-reasoning third-party models were not live-probed. They are covered by unit-level contract and wire tests, including real SDK calls over mocked HTTP for the SDK integration paths, and remain follow-up live-probe coverage.

### Required (all PRs)

### DIFF
--- a/sieval/core/models/dialect_registry.py
+++ b/sieval/core/models/dialect_registry.py
@@ -1,4 +1,4 @@
-"""Stable dialect descriptors and executable PR-1 binders.
+"""Stable dialect descriptors and executable adapter binders.
 
 Descriptors are serializable symbol metadata.  Binders are ordinary functions
 and exist only for dialect packages that are executable in the current delivery
@@ -38,6 +38,10 @@ from .dialects.openai_completions import (
     CAPABILITY_DECISIONS as OPENAI_COMPLETIONS_CAPABILITY_DECISIONS,
 )
 from .dialects.openai_completions import OpenAICompletionsDialect
+from .dialects.openai_responses import (
+    CAPABILITY_DECISIONS as OPENAI_RESPONSES_CAPABILITY_DECISIONS,
+)
+from .dialects.openai_responses import OpenAIResponsesDialect
 from .ir import (
     ChatInput,
     ChatMessage,
@@ -313,6 +317,10 @@ def _bind_openai_completions(connection: Any, requested_model_id: str) -> Dialec
     return OpenAICompletionsDialect(connection, requested_model_id)
 
 
+def _bind_openai_responses(connection: Any, requested_model_id: str) -> Dialect:
+    return OpenAIResponsesDialect(connection, requested_model_id)
+
+
 def _binder(
     function: Callable[[Any, str], Dialect],
     decisions: Mapping[str, DialectCapabilityDecision],
@@ -338,6 +346,13 @@ DIALECT_BINDERS: Mapping[str, DialectBinder] = MappingProxyType(
             cast(
                 Mapping[str, DialectCapabilityDecision],
                 OPENAI_COMPLETIONS_CAPABILITY_DECISIONS,
+            ),
+        ),
+        "openai_responses": _binder(
+            _bind_openai_responses,
+            cast(
+                Mapping[str, DialectCapabilityDecision],
+                OPENAI_RESPONSES_CAPABILITY_DECISIONS,
             ),
         ),
     }
@@ -372,7 +387,8 @@ DIALECT_SPECS: Mapping[str, DialectSpec] = MappingProxyType(
         "openai_responses": _spec(
             "openai_responses",
             "openai_sdk",
-            request_seed_support=RequestSeedSupport.RESERVED,
+            decisions=_registered_decisions("openai_responses"),
+            request_seed_support=RequestSeedSupport.UNSUPPORTED,
             input_kinds=("chat",),
             input_modalities=("text", "image", "tool_call", "tool_result"),
         ),
--- a/sieval/core/models/dialects/openai_responses.py
+++ b/sieval/core/models/dialects/openai_responses.py
@@ -1,0 +1,2076 @@
+"""OpenAI Responses API dialect for the provider-neutral model IR.
+
+This adapter deliberately implements only semantics the shared IR can preserve.
+Unsupported request fields and output item variants fail before they can be
+silently reinterpreted.  Streaming is lifted from the single terminal event's
+complete ``Response`` object, so the streaming and non-streaming paths share one
+response parser.
+
+AI-Generated Code - GPT-5.6 (OpenAI)
+"""
+
+import json
+import math
+from collections.abc import AsyncIterable, Callable, Mapping, Sequence
+from copy import deepcopy
+from dataclasses import dataclass, field
+from types import MappingProxyType
+from typing import Any, NoReturn, cast
+
+from sieval.core.models._shared import copy_json_value, thaw_json_mapping
+from sieval.core.models.capabilities import (
+    CAPABILITY_KEYS,
+    Capability,
+    CapabilityKey,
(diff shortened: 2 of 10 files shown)


ethan-scitix:
Verified empirically: 1888 tests pass at head (233 for the dialect); coverage 95% / 98% / 100% on the three changed core modules; 9/9 targeted reverse-mutations killed; `ruff` + `ty` clean. Wire format checked against the pinned SDK 2.36.0 — the flat Responses `text.format` (correctly *not* Chat's nested shape), `top_logprobs` 0–20, the three `include` values, and the genuine absence of `seed`/`n`/`stop`/penalties all confirmed. `"max"` effort is real, just newer than the SDK enum. Widening `TRUNCATION_FINISH_REASONS` does not rotate `rules_hash`, and no pre-existing dialect can emit `max_output_tokens`, so no historical `anomalies.json` changes verdict.

---

**1. Record two rationales — the only thing I'd change in this PR (a few comment lines).** Both are cases where the code is right, the reason isn't on site, and the repo's own *"extract on coupling"* rule actively pushes the next reader the wrong way.

**(a) `sieval/core/models/dialects/openai_responses.py:84-111`** — why 19 hand-written frozensets instead of `openai.types.responses.*`. The reason is *not* that the SDK under-models the schema; it's that the SDK **chooses not to validate**: `_base_client.py:669-677` falls through to `construct_type` ("if the given value does not match the expected type then it is returned as-is") unless `strict_response_validation` is set, which we don't set (`sieval/core/models/connection_factory.py:214`). End-to-end through a mocked transport, a non-standard `output_audio` content part is coerced into `ResponseOutputText` with `.text = None` and the real answer stranded in `.transcript` — an empty prediction, silently scored. Pinning the SDK doesn't help: it pins the client, not the server, and this dialect explicitly targets third-party endpoints. The replay path (`_decode_opaque_continuation`) never touches the SDK at all, so these allowlists are its only schema check.

*(FYI: `strict_response_validation=True` at client construction would catch the first two drift classes for free, but it's connection-global and converts lenient degradation into fleet-wide hard failure — likely not worth it, and it doesn't cover replay.)*

*(Minor: `_REASONING_ITEM_STATUSES`/`_MESSAGE_ITEM_STATUSES` and `_REASONING_TEXT_PART_KEYS`/`_INPUT_TEXT_PART_KEYS` are value-identical. Keep them separate — they mirror independent upstream enums — but a word saying so would stop them reading as copy-paste.)*

**(b) `sieval/core/models/dialects/openai_responses.py:791-841`** — `_verification_message_items` deliberately re-implements `_message_to_input_items`. Deduplicating them would turn the audit into a tautology, and every test would stay green.

**2. `Question:` A single terse reasoning item aborts the whole response — `openai_responses.py:1944`.** With `reasoning.summary` requested, every reasoning item must carry non-empty `summary_text`. Probed: two items where only the *second* has `summary: []` → the whole response is rejected, discarding the first item's good summary; and an item carrying `reasoning_text` but no `summary_text` → rejected even though `_visible_reasoning_text` (`:1152-1162`) already extracted usable text, making that fallback unreachable whenever a summary was requested. Multiple reasoning items are expected across consecutive function calls, and OpenAI documents no per-item guarantee. Suggest requiring *at least one* summarized item per response, or accepting the `content` fallback. Since this path wasn't live-probed, the cheapest resolution is to add a summary-non-empty assertion to the planned follow-up probe and decide from the result.

**3. `Nit:` Truncation is swallowed for structured output — `openai_responses.py:1494-1506`.** A `json_schema` request that hits the cap raises `structured output is not valid JSON` before `max_output_tokens` is ever recorded — which the second commit just taught the detector to see. `openai_chat` has the same shape, so this belongs in a cross-dialect follow-up, not here.

**4. `Nit:` The continuation bundle carries logprobs — `openai_responses.py:111`, `:1113-1137`.** `_OUTPUT_TEXT_PART_KEYS` includes `logprobs`, so every token's top-k is stored in `opaque_roundtrip` and replayed as request input on the next hop. Narrow combination, cheap to exclude.

**5. `Nit:` `openai_responses.py:1245`** — `cancelled` is terminal but lands in the `"non-terminal status"` branch. Rejecting it is right; the message misdescribes it.

The audit/verifier design is the strongest part of this PR — the `PreparedRequest` evidence chain genuinely catches lowering drift, which the mutation probes confirmed rather than assumed.

jack-scitix-ai:
Thanks — addressed the actionable findings in the latest commit:

- Documented why the dialect maintains local wire allowlists instead of
  deriving them from the SDK, including lenient SDK response construction,
  opaque replay bypassing SDK parsing, and intentionally separate
  value-identical upstream enums.
- Documented why `_verification_message_items()` must remain independent from
  `_message_to_input_items()` so the audit does not become self-verifying.
- Changed requested reasoning-summary validation from a per-item requirement
  to a response-level contract:
  - at least one non-empty `summary_text` is required;
  - items without summaries no longer invalidate an otherwise valid response;
  - raw `reasoning_text` is not mixed into a requested summary.
  
  This also closes an escape path in the earlier implementation: with an
  opaque continuation, a response containing no new reasoning item could
  satisfy the required `reasoning` channel using only the replay bundle,
  without delivering the requested visible summary. It now fails loudly.
  Unit tests cover zero-summary, mixed-item, and multi-summary responses. The
  same non-empty-summary assertion will be included in the planned live-probe
  follow-up.
- Clarified that `cancelled` is an unsupported terminal status.
- Kept output-message logprobs in stateless continuation history intentionally.
  This is supported by three independent pieces of evidence:
  1. OpenAI's reasoning guide says the relevant history items must be passed
     into the next response “untouched”.
  2. Its stateless workflow says to preserve “every output item” and replay the
     complete history.
  3. The repository-pinned OpenAI SDK 2.9.0 includes `logprobs` on the
     input-side `ResponseOutputTextParam`.

  References:
  - https://developers.openai.com/api/docs/guides/reasoning#keeping-reasoning-items-in-context
  - https://developers.openai.com/api/docs/guides/reasoning#preserve-reasoning-without-stored-responses

  A real-SDK/mock-HTTP regression test now verifies that non-empty logprobs
  survive into the second request's actual JSON body.

The structured-output/truncation ordering issue is shared with `openai_chat`,
so I am leaving it for a cross-dialect follow-up rather than expanding this PR.

Validation:

- 234 Responses dialect tests passed
- 836 related tests passed
- `ruff check` passed
- `ruff format --check` passed
- `ty check` passed

